PR #219, opened 2026-04-08
chore: kg_nodes deduplication and uniqueness constraint (closes #157)

### LINKED ISSUE #157: chore: deduplicate kg_nodes and enforce uniqueness constraint on (label, type)
## Background

Migration 014 (feat/relationship-management) adds a bidirectional unique constraint to `kg_edges`. The same problem exists for `kg_nodes`: there is no unique constraint on `(label, type)`, so concurrent relationship extractions can race and produce duplicate nodes for the same entity (e.g., two `person` nodes both labeled "Joseph Fung").

The `query-relationships` skill degrades gracefully when duplicates exist (returns `ambiguous: true`), but the underlying data is still messy and will confuse future queries.

## What needs to happen

1. **Dedup pass** — for each `(lower(label), type)` group with more than one node, merge them:
   - Pick a canonical node (highest confidence, ties broken by oldest `created_at`)
   - Re-point all `kg_edges` referencing the duplicate node IDs to the canonical node ID
   - Re-point any other foreign keys that reference `kg_nodes.id`
   - Delete the duplicate nodes
2. **Unique constraint** — `CREATE UNIQUE INDEX idx_kg_nodes_unique ON kg_nodes (lower(label), type) WHERE type != 'fact'`
   - Exclude `fact` nodes — facts are intentionally many per entity and have no meaningful label uniqueness
   - Legitimate label collisions across types (e.g., "Apple" org vs "Apple" concept) are allowed by scoping the constraint to `(label, type)` pairs
3. **`createEntity` update** — use `INSERT ... ON CONFLICT (lower(label), type) WHERE type != 'fact' DO UPDATE SET ...` to make node creation atomic, mirroring the edge upsert change in #014

## Notes

- Node merging is more involved than edge deduplication because edges must be re-pointed, not just deleted
- Consider whether `EntityMemory.mergeEntities()` (added in the contact-merge PR) can be reused or extended for this
- The `contact-merge` flow already handles KG node merging for contacts — check for overlap before implementing from scratch

josephfung:
## Summary

- **Migration 016** — one-time dedup pass that identifies duplicate `kg_nodes` rows by `(lower(label), type)`, re-points `kg_edges` and `contacts.kg_node_id` to canonical nodes, then deletes duplicates. Adds `idx_kg_nodes_unique` on `(lower(label), type) WHERE type != 'fact'` to prevent future duplicates.
- **`KnowledgeGraphStore.upsertNode()`** — new idempotent method mirroring `upsertEdge`. Uses `INSERT ... ON CONFLICT DO UPDATE` on Postgres; case-insensitive scan in the in-memory backend. Raises confidence on conflict (never lowers), leaves properties/label/embedding untouched.
- **`EntityMemory.createEntity()`** — now delegates to `upsertNode` and returns `{ entity, created }`. Breaking change; all internal call sites updated.
- **`EntityMemory.updateNode()`** — new public method. Label changes that collide with an existing node of the same type trigger `mergeEntities` automatically instead of throwing a constraint violation. Returns `{ node, merged }` with a strong JSDoc warning when `merged: true` (node.id in the result differs from the input id).
- **`mergeEntities` Phase 2** — completes the deferred TODO: re-points secondary entity edges to primary via `upsertEdge`, deletes secondary fact nodes, deletes the secondary entity node.

## Test plan

- [ ] All 1009 unit tests pass (`npm test`)
- [ ] `upsertNode` returns `created: false` on second call with same label (case-insensitive)
- [ ] `updateNode` with `label: 'Joseph Fung'` when that node already exists merges and returns `merged: true`
- [ ] Migration 016 runs cleanly against a fresh Postgres instance (integration test harness)
- [ ] After migration, verify `idx_kg_nodes_unique` exists and prevents duplicate inserts

## Summary by CodeRabbit

* **New Features**
  * Idempotent entity upsert/creation that returns the created entity plus a creation flag; public node-update API that can merge on label/type collisions. Contact creation now applies roles when reusing existing entities.

* **Bug Fixes**
  * One-time deduplication migration and enforcement of case-insensitive label+type uniqueness for non-fact nodes.
  * Completed merge flow to re-point relationships and remove merged nodes.

* **Documentation**
  * Added design and changelog entries describing uniqueness and migration behavior.

* **Tests**
  * Added and updated tests for upsert, merge, update and related call-site changes.

* **Chores**
  * Version bumped to 0.12.2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,20 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **KG node deduplication** — one-time migration deduplicates existing `kg_nodes` rows with matching `(lower(label), type)`, re-pointing edges and contacts to canonical nodes before removing duplicates.
+
+### Added
+- **`kg_nodes` uniqueness constraint** — `idx_kg_nodes_unique` on `(lower(label), type) WHERE type != 'fact'` prevents future duplicate entity nodes.
+- **`KnowledgeGraphStore.upsertNode()`** — idempotent node creation; raises confidence on conflict, never creates duplicates. Returns `{ node, created }`.
+- **`EntityMemory.createEntity()`** — now returns `{ entity, created }` instead of `KgNode`. Delegates to `upsertNode` for race-safe creation. **Breaking change** for callers (all internal call sites updated).
+- **`EntityMemory.updateNode()`** — new public method. Label changes that collide with an existing node of the same type automatically merge the nodes. Returns `{ node, merged }` — callers must use the returned `node.id`, which may differ from the input id when `merged: true`.
+- **`mergeEntities` Phase 2** — re-points secondary entity edges to primary and deletes the secondary node (was previously deferred with a TODO).
+
+---
+
 ## [0.12.1] — 2026-04-07
 
 ### Added
--- a/docs/wip/2026-04-08-kg-node-uniqueness-design.md
+++ b/docs/wip/2026-04-08-kg-node-uniqueness-design.md
@@ -99,18 +99,41 @@ The `KnowledgeGraphBackend` interface gains a corresponding `upsertNode` method.
 
 ### 3. `EntityMemory` changes
 
-#### 3a. Expose `updateNode()` publicly
+#### 3a. Expose `updateNode()` publicly — with merge-on-collision
 
-`EntityMemory` gains a public `updateNode()` pass-through to `store.updateNode()`. It is
-currently called only internally (in `mergeEntities` and `storeFact`), but `contact-service`
-needs to call it after a `createEntity` conflict.
+`EntityMemory` gains a public `updateNode()` method. It is not a simple pass-through —
+when `updates.label` is present it must guard against the case where the new label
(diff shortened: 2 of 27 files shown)


josephfung:
Pushed. @coderabbitai Here's the summary of what was fixed vs dismissed:

**Fixed (6 real issues):**

- upsertNode/upsertEdge Postgres: (xmax = 0) AS is_new — more reliable than timestamp comparison
- mergeEntities: added otherId === primaryId guard to prevent self-loops on primary after merge
- Migration Step 4: replaced contacts UPDATE with ROW_NUMBER() CTE to handle multiple duplicates mapping to the same canonical without violating the unique index
- Migration Step 1: added id ASC as stable tie-breaker in canonical selection
- updateNode: added try/catch around store.updateNode to handle the TOCTOU window with a merge on Postgres 23505 constraint violation
- Test stub in template-meeting-request.test.ts: duplicate check now includes type to match real upsertNode semantics

**Dismissed (1 false positive):**

The edges EXISTS check with both-endpoints-duplicated — the OR-based join catches both paths safely
Design doc: Updated versioning rationale to clarify why createEntity() stays at patch (EntityMemory is not a listed public API surface per CLAUDE.md; all callers updated atomically).

PR #338, opened 2026-04-24
fix: prevent research-analyst context flooding and surface specialist errors

### LINKED ISSUE #329: research-analyst floods context with raw HTML, stalls before synthesizing
## Summary

When the coordinator delegates a multi-source research task, the `research-analyst` defaults to `web-fetch` calls that return full HTML pages. After ~10-15 fetches the context window is saturated with markup and the agent is unable to synthesize a response.

## Observed behavior (2026-04-23)

- **First attempt (pre-restart):** silently stalled — no `agent.response`, no error logged, coordinator hung indefinitely waiting on `delegate`. Audit log shows 12 sequential `skill.invoke`/`skill.result` pairs ending at `18:36:16 UTC`, then nothing.
- **Second attempt (post-restart):** delegate timed out at 60s with `"Specialist 'research-analyst' did not respond within 60000ms"`; coordinator fell back to `web-search` (Tavily) and completed successfully.

Both runs were triggered by a CEO reply to a writing scout results email asking for data on AI training costs.

## Root cause

`web-fetch` returns raw HTML; a single page can be 50–100KB of markup. The research-analyst's configuration doesn't constrain it to use `web-search` first or limit `max_length` on fetches. After 10–15 large fetches the context window is exhausted mid-run.

The coordinator's own fallback (`web-search` / Tavily returning compact snippets) completed the same research task successfully in a single pass.

## Fix direction

- Research-analyst system prompt should prefer `web-search` (Tavily) for broad discovery; `web-fetch` only for targeted follow-up on a specific URL once a source is identified
- Add `max_length` cap on all `web-fetch` calls in the research-analyst context (e.g. 5,000 chars)
- Confirm the pre-restart silent stall (no error logged on context exhaustion) is fixed by the 60s delegate timeout; if not, add explicit error handling so the delegate always gets a result or a logged failure

josephfung:
Closes #329

## Summary

- **research-analyst now has `web-search` (Tavily) pinned** and the system prompt instructs it to search first and limit `web-fetch` to at most 3–4 targeted follow-up fetches. Previously the agent had only `web-fetch`, causing it to accumulate 50KB/page of raw HTML until the context window was exhausted, stalling the coordinator.

- **`isError?: boolean` added to `AgentResponsePayload`** (optional, backward compatible). Set on every fallback/error response path in the runtime — both `sendErrorResponse`, the outer `handleTask` catch, and three additional fallback branches inside `processTask` (`tool_use` without execution layer, empty-text recovery failure, unreachable else guard).

- **Delegate skill surfaces specialist failures correctly** — checks `isError: true` and returns `{ success: false }` instead of forwarding the fallback message as if it were a real research result.

- **Scheduler correctly tracks failed runs** — the `agent.response` subscriber now skips `isError: true` responses (deferring to `agent.error`), preventing a double `handleCompletion` call and ensuring `consecutive_failures` increments on error runs.

## Public API change

`AgentResponsePayload` in `src/bus/events.ts` gains `isError?: boolean`. Additive and backward compatible — existing consumers that don't read the field are unaffected.

## Test plan

- [x] All 1432 unit tests pass (`npm test`)
- [x] New test: delegate returns `{ success: false }` when specialist responds with `isError: true`
- [x] Updated test: runtime error response now asserts `payload.isError === true`
- [x] Verify research-analyst YAML loads without errors at startup (agent registry validates `pinned_skills` against registered skills)

## Summary by CodeRabbit

* **New Features**
  * Research analyst workflow enhanced with improved web search prioritization for greater efficiency.

* **Bug Fixes**
  * Improved error detection in specialist agent responses for better reliability.
  * Fixed job completion tracking for scheduled tasks to prevent dropped notifications.
  * Corrected draft handling in triage workflows.

* **Breaking Changes**
  * Calendar registration now requires explicit contact ID instead of using a default value.


josephfung:
@coderabbitai review

PR #341, opened 2026-04-24
feat: add contact-rename skill

josephfung:
## Summary

- Adds `skills/contact-rename/` — a new skill that updates a contact's display name via `ContactService.updateDisplayName()`
- Closes the capability gap where Curia could create contacts and set their role but had no tool to correct or expand a display name (e.g. "Jodi" → "Jodi Arnott")
- Follows the same pattern as `contact-set-role` and `contact-set-trust`

## What was broken

`ContactService.updateDisplayName()` has existed since PR #63 but was never exposed as a skill. When asked to rename a contact, Curia correctly reported it had no tool available.

## Implementation notes

- Pre-flight `getContact()` check for structured not-found detection (no error-message string matching — same convention as `contact-set-trust`)
- Whitespace-only display names are rejected before hitting the service
- Leading/trailing whitespace is trimmed before storing
- `log.warn` on not-found so stale-UUID attempts are auditable
- 9 unit tests covering all validation paths and success cases

## Known gap (pre-existing, not in scope)

The `updateContact()` Postgres implementation does not check `rowCount` after the UPDATE, so a phantom write (row deleted between the pre-flight read and the write) would silently return success. This affects all contact-update skills equally — tracked for a separate fix.

## Test plan

- [x] All 1448 unit tests pass (`npm test`)
- [x] Typecheck clean (`npm run typecheck`)
- [ ] In Curia chat: create a contact with a first name only, then ask Curia to rename it — should succeed
- [ ] Ask Curia to update name + title for Jodi, Barb, and Thusenth in a single message — Curia should call `contact-rename` and `contact-set-role` for each

## Release Notes

* **New Features**
  * Added the ability to update a contact's display name with built-in validation for contact identification and name formatting.

* **Chores**
  * Updated package version to 0.19.5.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Added
 
+- **`contact-rename` skill** — new skill that updates a contact's display name via `ContactService.updateDisplayName()`. Closes the gap where Curia could create contacts and set their role but had no tool to correct or expand a display name (e.g. from "Jodi" to "Jodi Arnott").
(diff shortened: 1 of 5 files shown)


josephfung:
@Coderabbitai review

josephfung:
@coderabbitai review

PR #344, opened 2026-04-24
fix: inject coordinator's own email and phone into system prompt

josephfung:
## Summary

- Adds `channelAccounts?: { email?, phone? }` to `AgentConfig`, sourced from the existing `NYLAS_SELF_EMAIL` and `SIGNAL_PHONE_NUMBER` env vars — no new configuration required
- The coordinator runtime appends a **"Your Contact Details"** block to the effective system prompt each turn, giving the LLM concrete values for its own accounts
- Updates `agents/coordinator.yaml` to reference the injected block explicitly instead of the vague "your own account" language
- Follows the exact same per-task injection pattern as `timezone` and `autonomyService`

## Why

Curia was guessing the CEO's email when MCP tools (e.g. `workspace-mcp`'s `user_google_email`) asked for an account identifier. The system prompt already said "use your own account" but never said what that was concretely. The LLM fell back to the sender's email, which has no tokens cached in the MCP server.

This is the generic fix for the pattern: any future tool that takes an "acting as" parameter will see the correct values in the system prompt.

## Test plan

- [ ] Boot Curia locally; confirm "Your Contact Details" block appears in coordinator's effective system prompt
- [ ] Ask Curia "Can you create a Google Doc?" — should now succeed using `nathancuria1@gmail.com` rather than erroring with an auth prompt for the CEO's email
- [ ] Confirm `channelAccounts` is only passed to the coordinator (specialist agents omitted), matching the `timezone` pattern
- [ ] Confirm that if `NYLAS_SELF_EMAIL` is unset, the email line is omitted rather than emitting a blank bullet

## Release Notes – Version 0.19.7

* **Changed**
  * Enhanced agent identity handling for tool calls—the coordinator now uses provided contact details (email/phone) instead of defaulting to CEO credentials, improving account routing accuracy for third-party integrations like messaging and calendar.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,6 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ## [Unreleased]
 
-### Changed
-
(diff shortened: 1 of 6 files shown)


josephfung:
@coderabbit review

PR #348, opened 2026-04-25
feat: defense-in-depth skill-layer block for observation mode (closes #305)

### LINKED ISSUE #305: Defense-in-depth: block email-reply in observation mode at the skill layer
## Background

PR #304 added explicit \"Do NOT call email-reply\" language to the observation-mode triage preamble for NOISE-classified emails, fixing the immediate dangling-draft bug. But prompt-only enforcement is fragile — the coordinator can still ignore the instruction, and if it does, the draft lands silently in the CEO's inbox with no error and no audit trail linking it to the NOISE classification.

The silent-failure-hunter review of PR #304 flagged this as a predating silent-failure surface that the prompt fix doesn't fully close.

## Proposed fix

Short-circuit \`email-reply\` and \`email-draft-save\` at the skill handler when \`metadata.observationMode === true\` and the triage classification was NOISE or LEAVE FOR CEO.

Two possible designs:

### Option A: Hard block in observation mode

If the handler sees \`observationMode: true\` on the inbound task's metadata, reject any \`email-reply\` call with a structured \`AgentError\`. Simple, but prevents legitimate NEEDS_DRAFT replies.

### Option B: Context-aware block (Recommended)

The dispatcher injects the triage classification (URGENT / ACTIONABLE / NEEDS DRAFT / LEAVE FOR CEO / NOISE) into the task metadata. \`email-reply\` handler checks: if \`observationMode === true\` AND classification is NOISE or LEAVE FOR CEO, reject with \`AgentError\`.

Option B requires the coordinator to emit its classification as a structured field, not just prose. This could be a tool call (\`classify-triage\`) or a required JSON block in the response. Worth brainstorming.

### Option C: Autonomy engine rule

Add an autonomy rule that downgrades \`email-reply\` to require score ≥ 100 (impossible) when \`observationMode === true\` AND classification is NOISE or LEAVE FOR CEO. Same effect as B, routed through the existing autonomy plumbing.

## Why this matters

- **Auditability.** A blocked call produces an \`AgentError\` that's logged and audited. The current silent draft produces nothing.
- **Defense in depth.** The prompt fix assumes perfect coordinator compliance. This gives us a hard wall when the model slips.
- **Platform pattern.** Observation mode is likely to expand to other channels and other \"do not act\" categories. Codifying the enforcement as a platform rule (not a prompt line) scales better.

## Scope

- Surface the triage classification as structured metadata (may require coordinator prompt adjustment — separate work)
- Add the block in \`skills/email-reply/handler.ts\` and \`skills/email-draft-save/handler.ts\`
- Tests covering the rejection path

## Related

- PR #304 — the prompt-only fix that motivated this
- ~~Issue #301~~ — \`email-draft-save\` landed; the block in this issue covers it
- Spec 14 (autonomy engine) — if Option C is chosen

josephfung:
## Summary

- **Threads `taskMetadata`** from the `agent.task` event through `ExecutionLayer.invoke()` into `SkillContext`, making task-level signals (e.g. `observationMode`) available to any skill handler
- **Hard-blocks `email-reply`** when `observationMode === true` — the skill returns a structured error before any gateway call, since `email-reply` always sends immediately and is never appropriate in observation mode
- **Classification-aware block on `email-draft-save`** in observation mode — only `triage_classification: "NEEDS DRAFT"` is allowed through; NOISE, LEAVE FOR CEO, or absent classification returns an auditable error
- **Updates coordinator prompt** to always pass `triage_classification: "NEEDS DRAFT"` when saving a draft in observation mode
- Bumps version to `0.19.9`, adds CHANGELOG entry

## Why

Prompt-only enforcement is fragile — a model slip can cause the coordinator to call `email-reply` or `email-draft-save` for emails it should leave alone. This adds a second enforcement layer at the skill layer so that mistakes produce auditable `{ success: false, error }` results rather than unintended sends or drafts from the CEO's inbox.

## Test plan

- [ ] Full test suite passes (`npm test`) — 1489 tests, 0 failures
- [ ] `email-reply` blocks with a structured error in obs mode; gateway never called
- [ ] `email-draft-save` allows `NEEDS DRAFT` through in obs mode; blocks NOISE, LEAVE FOR CEO, absent
- [ ] Both skills behave normally when `observationMode` is absent or `false`
- [ ] `taskMetadata` pass-through verified via `execution.test.ts`

## Release Notes

* **New Features**
  * Introduced observation-mode support with task metadata propagation, enabling smarter email handling; email reply is now blocked in observation mode, and email draft save validates triage classification to prevent incorrect operations.

* **Tests**
  * Added comprehensive test coverage for observation-mode guards and email draft/reply validation logic.

* **Documentation**
  * Updated changelog documenting new observation-mode defense-in-depth behaviours.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ bus event types) are noted explicitly even in the `0.x` range.
 ### Changed
 
 - **Dispatcher observation-mode branch** — now emits a `warn`-level log when a non-`LEAVE FOR CEO` classification completes with zero skill calls (defensive check for coordinator stalls)
+- **Defense-in-depth: skill-layer block for observation-mode triage** (closes #305)
+  - `email-reply` is hard-blocked when `observationMode` is set in task metadata — the skill returns a structured error rather than silently sending a reply from the CEO's inbox.
+  - `email-draft-save` requires `triage_classification: "NEEDS DRAFT"` in observation mode. Calls with NOISE, LEAVE FOR CEO, or no classification are rejected with an auditable error.
+  - Task metadata (`observationMode`, future signals) is now threaded from the `agent.task` event through `ExecutionLayer.invoke()` into `SkillContext`, making task-level context available to any handler without bus access.
 
 ---
 
@@ -406,8 +410,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 - **Bootstrap orchestrator** — `src/index.ts` wires all layers in dependency order
 - Architecture specs 00–08, contributor docs (CONTRIBUTING.md, CODE_OF_CONDUCT.md, SECURITY.md)
 
-[Unreleased]: https://github.com/josephfung/curia/compare/v0.19.9...HEAD
-[0.19.9]: https://github.com/josephfung/curia/compare/v0.19.8...v0.19.9
+[Unreleased]: https://github.com/josephfung/curia/compare/v0.19.7...HEAD
 [0.19.7]: https://github.com/josephfung/curia/compare/v0.18.1...v0.19.7
 [0.18.1]: https://github.com/josephfung/curia/compare/v0.18.0...v0.18.1
 [0.18.0]: https://github.com/josephfung/curia/compare/v0.17.0...v0.18.0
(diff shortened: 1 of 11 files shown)


josephfung:
@coderabbitai review

PR #353, opened 2026-04-25
feat: memory-store skill — explicit agent-driven KG fact writes

### LINKED ISSUE #297: feat: memory-store skill — explicit agent-driven KG fact writes
## Summary

Agents currently have no way to explicitly write a named fact to the knowledge graph. `extract-facts` handles the implicit extraction-from-text pipeline, and the domain-specific `knowledge-*` skills cover narrow categories. What's missing is a general-purpose write path where an agent can say: "store this specific fact about this entity, with this confidence, decay class, and sensitivity."

Implement a `memory-store` skill (`skills/memory-store/`) that writes a fact node to the KG, linked to a named entity.

## Inputs

- `entity` — the entity label or KG node ID to attach the fact to
- `field` — the attribute name (e.g. `preferred_airline`, `current_role`)
- `value` — the value to store
- `confidence` — float 0–1 (default 0.8)
- `decay_class` — `permanent | slow_decay | fast_decay` (default `slow_decay`)
- `source` — provenance string
- `sensitivity` — optional: `public | internal | confidential | restricted`. When omitted, the memory engine auto-classifies from content and defaults to `internal`. Callers that know the sensitivity of their data (e.g. a skill writing financial info) should set this explicitly rather than relying on keyword detection.
- `sensitivity_category` — optional category hint (e.g. `"financial"`). Passed through to the classifier; allows category-based classification without keyword scanning. Only used when `sensitivity` is omitted.

## Behavior

- Resolve entity by label if an ID isn't given; if ambiguous, return an `ambiguous: true` response with candidates (same pattern as `query-relationships`)
- Pass the resolved entity + fact through `EntityMemory.storeFact()`, which runs the full validation pipeline. The skill's response must distinguish all four pipeline outcomes:
  - **`stored: true, action: "created"`** — new fact node and edge were created
  - **`stored: true, action: "updated"`** — existing fact node was updated in place; `last_confirmed_at` bumped
  - **`stored: false, action: "conflict"`** — a contradicting fact exists; includes `reason` and `existing_node_id` so the agent can surface the conflict to the CEO
  - **`stored: false, action: "rejected"`** — the validator blocked the write; includes `reason`
- On success, return `node_id` and the `sensitivity` level that was actually assigned (auto-classified or explicit). This lets the agent emit its own note if a fact ended up tagged `restricted` unexpectedly.
- `action_risk: low` (internal state write)

## Files

- `skills/memory-store/skill.json`
- `skills/memory-store/handler.ts`
- `skills/memory-store/handler.test.ts`

## Checklist

- [ ] `skill.json` manifest with `action_risk: low`
- [ ] Handler resolves entity by label or ID; returns `ambiguous` response when multiple nodes match
- [ ] `sensitivity` and `sensitivity_category` inputs passed through to `EntityMemory.storeFact()` / `StoreFactOptions`
- [ ] Response distinguishes all four `ValidationResult` outcomes: `created`, `updated`, `conflict`, `rejected`
- [ ] Success response includes `node_id` and the assigned `sensitivity`
- [ ] `capabilities.ts` entry: `entityMemory` (once #119 lands; add `infrastructure: true` in the interim)
- [ ] `pnpm typecheck` and `pnpm test` pass

Closes part of spec §03 (Built-in skill: `memory-store`).

josephfung:
## Summary

Closes #297.

- **`skills/memory-store/`** — new skill with `skill.json`, `handler.ts`, and 17 tests in `handler.test.ts`
- **`EntityMemory.storeFact()` refinements** — attribute-based facts now automatically route through `validateContradiction()` (previously only `validate()` was called, so same-field contradictions were silently missed); `StoreFactResult` gains `action: 'created' | 'updated' | 'conflict' | 'rejected'` and `existingNodeId` (for the conflict case)

### Skill behaviour

Agents write a named attribute fact about a known entity. The skill:

1. Resolves entity by label (case-insensitive) or direct node ID
2. Returns `{ ambiguous: true, candidates }` when multiple nodes share the label
3. Returns `{ stored: false, action: 'rejected' }` when no entity is found
4. Calls `storeFact()` and maps all four outcomes to distinct response shapes:
   - `created` — new fact node, returns `node_id` + `sensitivity`
   - `updated` — near-duplicate merged in place, same fields
   - `conflict` — contradicting attribute fact, returns `reason` + `existing_node_id` for CEO escalation
   - `rejected` — rate limit or entity no longer exists, returns `reason`

### Pre-existing issues flagged (not touched, scope discipline)

- `mergeEntities` discards `storeFact` return values in its loop — facts can be silently lost during an entity merge if the rate limiter fires
- `validateContradiction` delegates to `validate()` after finding no contradiction, causing a double edge-walk
- `recordWrite` is called before `getNode` in the update branch — a throw inflates the rate-limit counter

## Test plan

- [ ] `pnpm typecheck` passes
- [ ] `pnpm test` passes (1520 tests, 17 new)
- [ ] Skill loads correctly at startup (skill registry smoke test covers this)
- [ ] Manual: call `memory-store` with a known entity name, verify `created` response
- [ ] Manual: call again with a different value for the same field, verify `conflict` response with `existing_node_id`

## Release Notes

* **New Features**
  * Added `memory-store` skill for storing named attribute facts about entities with configurable confidence, decay class, and sensitivity. Returns four action outcomes (created, updated, conflict, rejected) with enhanced conflict reporting for contradicting entities.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Added
 
+- **`memory-store` skill** — general-purpose KG fact write path: agents can now explicitly store a named attribute fact about a known entity with full control over confidence, decay class, and sensitivity. Resolves entity by label or node ID; returns one of four outcomes (`created`, `updated`, `conflict`, `rejected`) so the agent can surface contradictions to the CEO before proceeding (closes spec §03, #297)
+- **`StoreFactResult.action`** — `storeFact()` now returns the pipeline outcome (`created | updated | conflict | rejected`) alongside `stored`; attribute-based facts now route through contradiction detection automatically
+- **`StoreFactResult.existingNodeId`** — populated when a fact write produces a `conflict`, letting callers surface the contradicting node to the CEO
+
 - **KG explorer: sensitivity visualization** — node detail drawer (tap any node to inspect all attributes), color-by toggle (Type / Sensitivity / Decay class), degree-based node sizing, and confidence-based opacity (#350)
 - **KG API: `sensitivity` and `properties` fields** — `/api/kg/nodes` and `/api/kg/graph` now return both fields on every node response (#350)
 - **`SensitivityClassifier` unit tests** — 5 tests covering keyword matching, property-value matching, category hint bypass, and most-restrictive-wins behaviour (#350)
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "curia",
-  "version": "0.20.1",
+  "version": "0.21.0",
   "description": "The AI executive staff your C-Suite will use and your Board will trust",
   "type": "module",
   "engines": {
--- a/skills/memory-store/handler.test.ts
+++ b/skills/memory-store/handler.test.ts
@@ -1,0 +1,296 @@
+// handler.test.ts — memory-store skill unit tests.
+//
+// Uses real EntityMemory (backed by in-memory KG store) for entity resolution
+// and the full created/conflict pipeline. Uses a mock entityMemory stub for
+// the updated and rejected outcomes, which require state that is hard to
+// construct reliably with the in-memory backend.
+
+import { describe, it, expect, vi } from 'vitest';
+import pino from 'pino';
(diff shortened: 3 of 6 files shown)


josephfung:
@coderabbitai review please

Issue #388, opened 2026-04-28
chore: Add nested property definitions to skill tool schemas for bare "object" inputs


josephfung:
## Context

When a skill manifest declares an input as a bare `"object"` (e.g. `"writing_voice": "object"`), the generated tool schema has no nested property definitions. The LLM must guess the key names for the inner object, and may use either camelCase or snake_case.

This caused a bug where `executive-profile-update` silently dropped the `sign_off` field when the LLM emitted `signOff` (camelCase) instead of `sign_off` (snake_case). Fixed by adding a `normalizeKeysToSnakeCase` utility, but the root cause remains: the LLM lacks schema guidance.

## Proposal

Extend the skill manifest `inputs` notation to support nested property definitions for object-typed parameters. For example:

```json
"inputs": {
  "writing_voice": {
    "type": "object",
    "properties": {
      "tone": "string[]?",
      "formality": "number?",
      "patterns": "string[]?",
      "vocabulary": "object?",
      "sign_off": "string?"
    }
  }
}
```

The registry's `toToolDefinitions()` would then emit a proper JSON Schema with nested `properties`, giving the LLM explicit key names to use.

## Scope

- Extend manifest `inputs` parsing in `src/skills/registry.ts` to handle object-with-properties
- Update `executive-profile-update/skill.json` as the first consumer
- Audit other skills for bare `"object"` inputs that would benefit
- The `normalizeKeysToSnakeCase` utility remains as defense-in-depth

PR #389, opened 2026-04-28
fix: executive-profile-update persists sign_off regardless of key casing

josephfung:
## Summary

- **Bug:** `sign_off` is the only `writing_voice` field where snake_case and camelCase differ. The handler only checked `sign_off`, so when the LLM emitted `signOff` (camelCase) the value silently fell through to the existing empty default.
- **Fix:** Adds a shared `normalizeKeysToSnakeCase` utility (`src/skills/normalize.ts`) that any handler can import for bare-object inputs, and wires it into the update handler before the merge logic.
- **Future:** josephfung/curia#388 tracks adding nested property definitions to tool schemas so the LLM gets explicit key guidance (defense-in-depth, P5).

## Test plan

- [x] New utility tests (9 cases): basic conversion, identity, precedence when both forms present, empty input, type preservation, no recursion into nested objects
- [x] New handler tests (2 cases): camelCase `signOff` persists correctly, snake_case wins when both present
- [x] Existing handler tests still pass (no regressions)
- [x] Full test suite passes (1652 tests, 0 failures)

## Summary by CodeRabbit

**Bug Fixes**
- Fixed a persistence issue in the executive profile update skill where the sign_off field would not save correctly when using camelCase input naming conventions instead of the expected snake_case format. This fix ensures all profile updates persist reliably and consistently, regardless of the naming convention or format used in the input data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ## [Unreleased]
 
+### Fixed
+
+- **executive-profile-update** sign_off field not persisting when LLM emits camelCase key (`signOff`) instead of snake_case (`sign_off`). Added shared `normalizeKeysToSnakeCase` utility in `src/skills/normalize.ts` that any handler can use for bare-object inputs.
+
 ---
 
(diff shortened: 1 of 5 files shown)


josephfung:
@coderabbitai review please?

PR #450, opened 2026-05-05
fix: approve-action routes outbound-send re-execution to send-draft

josephfung:
## Summary

- **Root cause:** `approve-action` always called `ctx.executionLayer.invoke(row.skillName, ...)`. For gateway-gated sends, `skillName` is `'outbound-send'` — a synthetic label, not a registered skill. The execution layer couldn't find it, producing a registry error and `reExecutionSuccess: false`.
- **Fix:** When `skillName === 'outbound-send'`, route to `send-draft` instead, using `draftId` and `accountId` from the payload (written there by `linkGatedAction`). If the payload has no `draftId`, return a clear error rather than a confusing registry message.
- **Bonus:** Log the actual re-execution error when `reResult.success === false` — previously only `reExecutionSuccess: false` appeared, making failures invisible in the log trail.

## Test plan

- [ ] `npm test skills/approve-action/handler.test.ts` — 2 new tests pass: routes to `send-draft` with correct args, returns error when `draftId` missing
- [ ] All 9 tests in the file pass
- [ ] Full suite: no regressions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Fixed
 
+- **`approve-action` outbound-send re-execution** — approving a gateway-gated email now correctly invokes `send-draft` with the linked draft ID, instead of attempting to invoke `outbound-send` (a synthetic skill name not in the registry) which produced a registry error. Also added re-execution error logging so failures are visible in the log trail.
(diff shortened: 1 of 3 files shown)


josephfung:
Closing in favor of a cleaner architectural approach: instead of special-casing `outbound-send` in `approve-action`, channel adapters will pass a `reExecRecipe` to `gateway.send()` so the gateway stores the correct re-execution recipe at gate time. `approve-action` stays fully generic. Replacement PR coming.

PR #456, opened 2026-05-05
fix: send-draft not account-aware (#455)

### LINKED ISSUE #455: send-draft not account-aware: searches wrong account, sends from wrong account
## Summary

The `send-draft` skill has three related failures, all stemming from not being account-aware. When operating on a draft that belongs to a secondary account (e.g. the `joseph` `draft_gate` account), the skill searches the wrong account, falls back to creating a new draft on failure, and sends via the wrong account's credentials.

## Steps to reproduce

1. Ensure a draft exists in the `joseph` Nylas account (observation-mode triage draft, or manually created)
2. From CLI (CEO trust): `Send the draft reply to <subject/recipient description>`
3. Observe the three failure modes below

### 1. Draft discovery searches wrong account
The skill looks up drafts in the primary (curia) Nylas account rather than the `joseph` draft_gate account. The draft is not found.

### 2. Falls back to creating a new draft instead of failing
When draft discovery fails, the skill creates a new draft rather than surfacing an error or asking for clarification. This masks the discovery failure.

### 3. Sends from the wrong account
Even when the correct draft is explicitly identified, the send is executed using the curia account's Nylas credentials rather than the joseph account's. The email arrives from the wrong sender address.

## Expected behaviour

- Draft discovery queries the account the draft belongs to (derivable from the draft's Nylas `account_id` or the channel account config)
- If discovery fails, the skill surfaces an error rather than silently creating a new draft
- The send is executed using the credentials of the account that owns the draft

## Root cause hypothesis

`send-draft` does not thread the draft's originating account through to the send call. Both lookup and send appear to default to the primary channel account (curia). The skill needs to either: (a) derive the account from the draft's metadata, or (b) accept an explicit `accountId` parameter and propagate it to the Nylas send call.

## Acceptance criteria

- [ ] `send-draft` correctly locates drafts in the `joseph` account when invoked from CLI
- [ ] `send-draft` sends from the account that owns the draft (joseph's address, not curia's)
- [ ] If the draft cannot be found, the skill returns a clear error — it does not create a new draft as a fallback
- [ ] Existing `send-draft` test for non-CEO context block (test 4.2) continues to pass

josephfung:
## Summary

- **send-draft auto-discovers draft account**: The `account` input is now optional. When omitted, the skill iterates all configured email accounts via a new `gateway.listAccountIds()` method and stops at the first match. The send always uses the discovered account's credentials.
- **No silent fallback to draft creation**: If the draft is not found in any account, the skill returns a clear error — it never falls back to creating a new draft.
- **Structured error discrimination**: Draft discovery failures use a `reason` field (`'not_found'` | `'fetch_error'`) instead of string-prefix matching, per project conventions. When all accounts have fetch errors, the error message says "could not search" rather than the misleading "Draft not found."

Closes #455

## Test plan

- [x] 29 send-draft handler tests pass (8 new cross-account discovery tests + 1 all-accounts-fail test)
- [x] Gateway and draft-fallback-flow regression tests pass (63 tests)
- [x] Full suite passes (164 files, 1962 tests — 2 pre-existing integration failures unrelated to this change)
- [ ] Manual: create a draft in the secondary (joseph) account, invoke `send-draft` from CLI without specifying account, verify it discovers and sends from the correct account
- [ ] Manual: invoke `send-draft` with a nonexistent draft ID and no account, verify clear "not found" error listing all searched accounts

## Release Notes

* **New Features**
  * The send-draft skill now auto-discovers which email account owns a draft when the account input is omitted, automatically searching across all configured accounts to identify the correct account for sending.

* **Bug Fixes**
  * Missing drafts now return clear, descriptive error messages instead of silently falling back to draft creation, improving error visibility.


josephfung:
@coderabbitai review

PR #475, opened 2026-05-06
fix: coordinator no longer calls email-reply for direct inbound emails

josephfung:
## Summary

- When the CEO replies to a proactive email from Curia (e.g. a scheduled writing-scout report), the reply arrives as a direct inbound email with no observation-mode or CC preamble
- The coordinator prompt told it to call `email-reply` with a Nylas message ID — but no message ID is injected into the task for direct emails
- The coordinator would search `email-list` on the CEO's account, retrieve message IDs from that Nylas grant, and try to use them via Curia's account → 404 Not Found
- Fallback: `email-send` (new thread) + the `agent.response` "Replied. Here's the summary…" landing in the original thread via `sendOutboundReply()`
- Fix: for direct inbound emails, just compose the reply and return it; the dispatcher's `sendOutboundReply()` already handles threading correctly via `email:{threadId}`

## Root cause (from audit log)

```
20:29:50 skill.invoke  email-reply  {"reply_to_message_id": "message_id_placeholder"}  → 404
20:30:40 skill.invoke  email-reply  {"reply_to_message_id": "19dfef85aa12d07f"}        → 404 (CEO's Nylas grant, not Curia's)
20:31:19 skill.invoke  email-send   {"to": "joseph@josephfung.ca", ...}                 → new thread
20:31:28 agent.response "Replied. Here's the summary..." → sendOutboundReply → original thread
```

## Test plan

- [ ] Reply to a proactive Curia email (e.g. next writing-scout run) — verify reply lands in the original thread, no meta-commentary message
- [ ] CC a third-party email to Curia (`[OWNER CC —]` path) — verify coordinator still calls `email-reply` correctly
- [ ] Observation-mode email — unchanged path, still delegates to email-triage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ## [Unreleased]
 
+### Fixed
+- **Email reply threading** — coordinator no longer calls `email-reply` when responding to a direct inbound email. Previously, it would attempt to use Nylas message IDs from the CEO's inbox (a different Nylas grant) via Curia's account, receive a 404, fall back to `email-send` (creating a new thread), and then also send a meta-commentary "Replied." message in the original thread via its `agent.response`. Now it returns the reply content directly; the dispatcher threads it correctly via `sendOutboundReply()`.
+
 ---
 
 ## [0.25.0] — 2026-05-05 — "By Your Leave"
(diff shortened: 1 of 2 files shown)


josephfung:
Closing in favor of architectural redesign: CEO email moves from observation-mode channel to dedicated ceo-inbox agent + skills in curia-deploy. The prompt fix here is superseded by removing observation mode entirely.

PR #495, opened 2026-05-09
refactor(runtime): refresh Bullpen context on every tool-use loop iteration

### LINKED ISSUE #213: refactor(runtime): refresh Bullpen pending-thread context on every model round in tool-use loop
## Problem

In `src/agents/runtime.ts`, `getPendingThreadsForAgent` is called **once** (line 452) before the first `chatWithRetry` invocation (line 485). The resulting Bullpen system message is spliced into `messages` at that point and never updated, so any replies or thread closures that occur during subsequent rounds of the tool-use loop remain invisible to the model for the duration of that task.

The tool-use loop calls `chatWithRetry` at two points (lines 485 and 734), but the pending-thread fetch at lines 450–467 sits outside the loop entirely.

## Proposed Fix

Move (or repeat) the pending-threads fetch + `formatBullpenContext` + `messages.splice` (lines 450–467) inside the tool-use loop so that Bullpen state is refreshed before each `chatWithRetry` invocation. The existing `try/catch` behaviour (log error, continue without thread context) must be preserved to avoid aborting the task on a lookup failure.

On refresh, any previously injected Bullpen system message should be replaced (not duplicated) in the `messages` array.

## References

- PR: https://github.com/josephfung/curia/pull/208
- Review comment: https://github.com/josephfung/curia/pull/208#discussion_r3046076772
- Requested by: @josephfung

josephfung:
## Summary
- Extract `refreshBullpenContext()` private method on `AgentRuntime` that finds/removes the existing Bullpen system message by its `[Bullpen —` sentinel prefix, fetches fresh pending threads, and inserts the updated message
- Call it before both `chatWithRetry` invocations (initial + inside the tool-use loop) so the model sees current thread state, not a stale snapshot from task start
- On refresh failure, preserve the existing stale message (format before mutating the array)
- On refresh with no remaining threads, remove the stale message entirely

## Test plan
- [x] New test: re-fetches pending threads before each chatWithRetry in the tool-use loop (verifies `getPendingThreadsForAgent` called twice, system message content changes)
- [x] New test: removes stale Bullpen message when no pending threads remain
- [x] New test: preserves existing Bullpen message when refresh fails
- [x] All 33 runtime tests pass
- [x] Full test suite: 169 passed, 9 skipped (2 pre-existing DB-dependent failures unrelated to this change)

Closes #213

## Release Notes

* **New Features**
  * Agents now detect and respond to inter-agent messages and thread updates continuously throughout task execution, providing real-time visibility into agent interactions rather than only initial state.

* **Tests**
  * Added comprehensive test suite verifying inter-agent message detection, removal when no pending threads exist, and graceful handling of refresh failures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/josephfung/curia/pull/495)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 - **MCP `fixed_inputs`** — MCP server entries in `skills.yaml` now support a `fixed_inputs` field that binds constant parameter values at the tool layer. Values are resolved from env vars or literals at startup, stripped from tool schemas (invisible to agents), and merged into every `callTool` invocation (#432)
 
 ### Changed
+- **Bullpen context refresh in tool-use loop** — `AgentRuntime` now re-fetches pending Bullpen threads before every `chatWithRetry` call, not just once at task start. Replies and closures that occur mid-task are visible to the model on the next loop iteration (#213)
 - **`contact.resolved` bus event** — `sourceLayer` widened from `'dispatch'` to `'dispatch' | 'execution'`; `createContactResolved()` factory accepts an optional `sourceLayer` parameter (defaults to `'dispatch'` for backward compatibility). This is a public API surface change.
 - **`IdentitySource`** — new value `'agent_called'` for contacts registered by agents outside the normal dispatcher pipeline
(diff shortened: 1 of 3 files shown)


josephfung:
@coderabbitai review

PR #522, opened 2026-05-11
fix: resolve email addresses in entity-context via contact_channel_identities (#461)

### LINKED ISSUE #461: entity-context receives raw email address instead of contact UUID for CC'd recipients
## Summary

During CC role detection flows, `entity-context` emits a warning — `non-UUID id passed to resolveKgNodeId — treating as unresolved` — for the primary recipient's email address. This occurs even when that address belongs to a registered KG contact. The warning fired twice (coordinator turns 1 and 2) for `josephbrnnn@gmail.com` (Joseph Brennan), who is a known contact.

## Log evidence

```
entity-context: non-UUID id passed to resolveKgNodeId — treating as unresolved
id: josephbrnnn@gmail.com
```

Fired twice during conversation `email:19dfb5e006cf1cc3` (CC role detection test, 2026-05-05T22:59:23Z).

## Expected behaviour

When the coordinator assembles entity context for a CC'd conversation's primary recipient, it should resolve the email address to the contact's UUID via the identity index before calling `resolveKgNodeId`. The contact is registered; entity context should be populated.

## Actual behaviour

The raw email address string is passed directly to `resolveKgNodeId`. Since it's not a UUID, the node lookup is skipped and the entity is treated as unresolved — meaning no KG context (facts, relationships, history) is assembled for that contact.

## Root cause hypothesis

The CC preamble injection path extracts `primaryRecipients` from the email headers as a raw address string. When the coordinator later calls `entity-context` with that value, the skill skips the identity-to-UUID resolution step that the standard inbound sender resolution path performs. The fix is to run the same identity lookup (email address → contact UUID) before handing off to `resolveKgNodeId`.

## Impact

In CC flows, the coordinator operates without KG context for the primary recipient — missing relationship data, stored facts, and conversation history for that contact. Non-fatal, but degrades response quality for CC'd emails involving known contacts.

## Acceptance criteria

- [ ] When `entity-context` receives an email address, it resolves it to a contact UUID via the identity index before calling `resolveKgNodeId`
- [ ] No `non-UUID id passed to resolveKgNodeId` warning for registered contacts
- [ ] KG context (facts, relationships, history) is correctly assembled for the primary recipient in CC flows
- [ ] Unregistered addresses (genuinely unknown contacts) still degrade gracefully with the unresolved path

josephfung:
## Summary

- **Root cause:** `resolveKgNodeId()` passed all inputs directly to UUID-typed columns (`contacts.id`, `kg_nodes.id`). Email addresses from CC preambles and ceo-inbox triage triggered PostgreSQL `22P02` errors, causing the entity to be treated as unresolved with no KG context.
- **Fix:** Added an email detection branch (`/^[^\s@]+@[^\s@]+\.[^\s@]+$/`) at the top of `resolveKgNodeId()` — inside the existing `try/catch` — that resolves via `contact_channel_identities JOIN contacts` with case-insensitive matching before touching UUID columns.
- **Tests:** 4 new unit tests covering registered email→KG node, email→contact with no KG node, unregistered email, mixed email+UUID batch with cache validation, and 2 DB error propagation tests.

Fixes #461. Follow-up: prompt clarification in curia-deploy (`ceo-inbox.yaml` line 145) to say `contactIds` instead of "sender's email address" — the code fix handles either, but the prompt should be accurate.

## Test Plan

- [x] `npm test -- tests/unit/entity-context/assembler.test.ts` — 17/17 pass
- [x] `npm test` — 2056/2056 pass (2 pre-existing integration tests skip without `DATABASE_URL`)
- [x] Email path sits inside `try/catch` — DB errors propagate consistently with UUID path
- [x] `inputKind: 'email'` used in debug logs instead of `email: id` (which is on the pino redact list)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed knowledge graph context resolution to support email address inputs, enabling proper contact identification for CC flows and CEO inbox triage workflows for known contacts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/josephfung/curia/pull/522)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@ bus event types) are noted explicitly even in the `0.x` range.
 - **Principal identity consolidation** — replaced fragmented CEO identity (env var config, `role` column, `OutboundGatewayConfig` flat fields) with a database-driven `system_role` column on the `contacts` table. One contact holds `system_role='principal'`, one holds `system_role='agent'`. Enforced by partial unique indexes.
 - **TaskOriginator on every task** — replaced `ceoInitiated: boolean` with a `TaskOriginator` object stamped by the dispatcher on every task, carrying `contactId`, `systemRole`, `channel`, and `initiatedAt`. CEO-authorization checks now use `isPrincipalOriginated(taskMetadata)`. Originator context survives task delegation, resolving the autonomy gate bug for scheduled CEO-authorized actions.
 
+### Fixed
+
+- **Entity-context email resolution** — `resolveKgNodeId` now resolves email addresses via `contact_channel_identities` before attempting UUID-based queries, fixing missing KG context in CC flows and ceo-inbox triage for known contacts ([#461](https://github.com/josephfung/curia/issues/461))
+
 ---
 
 ## [0.26.0] — 2026-05-10 — "In the Record"
--- a/docs/wip/2026-05-11-entity-context-email-resolve-design.md
+++ b/docs/wip/2026-05-11-entity-context-email-resolve-design.md
@@ -1,0 +1,101 @@
+# Entity-Context Email Address Resolution — Design
+
+**Issue:** [#461](https://github.com/josephfung/curia/issues/461)
+**Date:** 2026-05-11
+
+## Problem
+
+`resolveKgNodeId()` in `src/entity-context/assembler.ts` only handles UUID inputs
+(contact IDs and KG node IDs). When an email address is passed — which happens in
+CC flows (coordinator) and CEO inbox triage (ceo-inbox agent) — PostgreSQL throws
+error `22P02` (invalid text representation for UUID). The catch block logs a warning
+and returns `undefined`, so no KG context is assembled for that entity.
+
+Two code paths are affected:
+
+1. **CC flow**: The dispatcher injects a `[OWNER CC —]` preamble containing raw
+   primary recipient email addresses. The coordinator extracts these and passes them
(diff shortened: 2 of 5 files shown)


josephfung:
@coderabbitai review

josephfung:
## CodeRabbit review response

Addressed the one valid nitpick; skipped the two outside-diff suggestions after verifying against the current code.

### Mixed-case email test ✅ Fixed

Added a test for `'Jenna@Example.com'` that directly exercises the `LOWER()` in the `contact_channel_identities` query. Commit: de1bde1.

### AgentError normalization ⏭️ Skipped

`EntityContextAssembler` has no `AgentError` imports anywhere — the module uses plain TypeScript errors throughout, consistent with the rest of the DB layer. `AgentError` belongs to the agent/runtime layer (`src/agents/`, `src/dispatch/`). Converting just the `22P02` catch in isolation would be inconsistent with every other error re-throw in the file. A full assembler conversion is out of scope for this fix.

### `id` leaking in warn log ⏭️ Skipped

The email detection branch (`/^[^\s@]+@[^\s@]+\.[^\s@]+$/`) at the top of `resolveKgNodeId` intercepts all email-shaped strings before they reach the `22P02` catch. Whatever lands in the catch block is guaranteed to be a hallucinated non-UUID synthetic string like `'primary-user'` or `'joseph-fung-contact-id'` — not PII. Logging it is the intended behaviour: operators need that string to trace what the LLM is hallucinating upstream.

PR #603, opened 2026-05-15
feat: notify CEO when watchdog auto-recovers a stuck job

### LINKED ISSUE #207: feat: notify coordinator when schedule.recovered fires
## Summary

When the scheduler auto-recovers a stuck job (resets it from `running` → `pending` after timeout), the coordinator should be notified so Joseph is alerted that a job silently failed and was reset.

## Blocked by

#206 — requires `outbound.notification` event type before adding any new direct-send notification paths.

## Acceptance criteria

- [ ] `schedule.recovered` event triggers a coordinator notification via `outbound.notification`
- [ ] Notification includes: job ID, agent ID, how long it was stuck, computed timeout threshold
- [ ] Covered by tests

josephfung:
## Summary

Closes #207.

- **`RecoveryNotifier`** — new system-layer subscriber to `schedule.recovered` that emails the CEO when the watchdog resets a stuck job. Mirrors the `SuspensionNotifier` pattern (direct `outbound.notification` path, bypassing LLM pipeline).
- **`'schedule_recovered'` notificationType** — added to the `OutboundNotificationPayload` discriminated union in `events.ts`.
- **Registration in `index.ts`** — registered immediately after `SuspensionNotifier`, with the same `outboundGateway && ceoPrimaryEmail` guard.

### Notification content

The email includes:
- Agent ID and job ID
- Stuck duration (computed from `runStartedAt` to event timestamp; falls back gracefully for pre-migration null rows)
- Computed timeout threshold that was exceeded
- Outcome: reset to pending or SUSPENDED (if `consecutiveFailures >= 3`)

### Important nuance

When `recoverStuckJobs()` suspends a job (consecutive failures ≥ 3), it emits **only** `schedule.recovered` — not `schedule.suspended`. `RecoveryNotifier` is therefore the sole CEO notification path for watchdog-triggered suspensions. `SuspensionNotifier` covers only the normal completion-failure path.

## Test plan

- [ ] `RecoveryNotifier` registers on `schedule.recovered` at the system layer
- [ ] Correct notification content for reset-to-pending case (stuck duration, timeout, `reset to pending` outcome)
- [ ] Correct notification content for suspended case (`SUSPENDED`, web app resume instructions)
- [ ] Null `runStartedAt` (pre-migration rows) falls back gracefully
- [ ] Hour-formatted timeouts render correctly (`1 h`, `1 h 30 min`)
- [ ] `sendNotification → false` logs an error and resolves without throwing
- [ ] CI passes (7 new unit tests, 2370 existing pass)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 - **Email attachment support** — surface metadata in email skills, append human-readable summaries to message content, and download bytes as base64 for `file-parse`.
 - **Email folder management skills** — `email-label`, `email-list-folders`, `email-create-folder`, and `email-mark-read` expose Nylas folder and read-state operations as general-purpose skills.
 - **`SuspensionNotifier`** — emails CEO when a scheduled job suspends after consecutive failures, bypassing the LLM pipeline for reliability during Anthropic outages. (#538)
+- **`RecoveryNotifier`** — emails CEO when the watchdog auto-recovers a stuck job (timeout exceeded), including stuck duration, timeout threshold, and whether the job was reset or suspended. (#207)
 
 ### Changed
 
--- a/src/bus/events.ts
+++ b/src/bus/events.ts
@@ -136,6 +136,7 @@ interface OutboundBlockedPayload {
 //   - 'approval_expired':     CEO alert that pending approvals expired without response (approval-expiry-sweep)
 //   - 'pending_actions_digest': Daily summary of open approvals awaiting CEO decision (pending-actions-digest)
 //   - 'schedule_suspended': CEO alert that a scheduled job was auto-suspended after consecutive failures (#538)
+//   - 'schedule_recovered': CEO alert that a stuck job was auto-recovered (reset to pending or suspended) (#207)
 export interface OutboundNotificationPayload {
   notificationType:
     | 'blocked_content'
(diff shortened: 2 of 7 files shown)


josephfung:
## ✅ Approve

**Codex review:** Great work on this PR. The `RecoveryNotifier` addition is well-structured and consistent with the existing scheduler notification architecture.

### What looks good
- Adds `schedule_recovered` cleanly to the outbound notification contract.
- Registers notifier in `index.ts` with the same safe guard pattern used elsewhere (`outboundGateway` + `ceoPrimaryEmail` checks).
- Keeps the recovery alert path out of the LLM pipeline (important for reliability when infra issues cause stuck jobs).
- Handles key cases clearly:
  - recovered + rescheduled,
  - recovered + suspended,
  - missing `runStartedAt` (pre-migration rows),
  - failed outbound publish (`sendNotification() === false`) with explicit error logging.
- Unit tests are focused and cover message semantics + duration formatting edge cases.

### Optional nits (non-blocking)
1. **Rounding semantics in `formatMinutes()`**  
   `Math.round(seconds / 60)` may round up early (example: ~31s becomes `1 min`). If desired, `Math.floor` could provide more conservative duration wording.

2. **Private method testing style**  
   Tests currently call `handle()` via casting. This is fine, but if you want less coupling to internals, you could capture/invoke the subscribed callback from `bus.subscribe()` and assert behavior through the public registration path.

Approved ✅

josephfung (commented):
Thanks for the review! Leaving both nits as-is:

1. **Rounding** — the `~` prefix already signals approximation, and `Math.round` is the conventional choice for rough duration labels. The edge cases only arise on values that are genuinely ambiguous either way. For the timeout path specifically, the recovery formula produces whole-minute multiples so it never triggers there.

2. **Testing style** — this follows the same `as never as { handle(...) }` pattern used in `suspension-notifier.test.ts`. Changing it here would create inconsistency; if we want to standardize on callback-capture that's a broader convention change.

Merging.

PR #604, opened 2026-05-15
fix(runtime): separate uninitialized executive service guard from voice-block compile catch

### LINKED ISSUE #508: fix(runtime): separate uninitialized-service guard from voice-block compilation catch
## Background

In `src/agents/runtime.ts` lines 202–211, compiling the `${executive_voice_block}` prompt
placeholder is wrapped in a single broad `try/catch`:

```ts
if (executiveProfileService) {
  try {
    effectiveSystemPrompt = effectiveSystemPrompt.replace(
      '${executive_voice_block}',
      executiveProfileService.compileWritingVoiceBlock(executiveDisplayName ?? 'the executive'),
    );
  } catch (err) {
    logger.error({ err, agentId }, 'Failed to compile executive voice block — ...');
  }
}
```

`compileWritingVoiceBlock` does two things in one expression:
1. `executiveProfileService.get()` — which throws `'ExecutiveProfileService not initialized'`
   if `initialize()` was never awaited (a startup/wiring programming error)
2. The actual prompt compilation — which could theoretically throw due to a bug

Both errors are caught and swallowed identically. The task continues with the literal
`${executive_voice_block}` string left in the coordinator prompt for the rest of the process
lifetime, silently degrading every task — with no operator signal beyond one error log line.

Spotted in review of PR #507.

## Problem

- An uninitialized service is a programming error / startup race condition that should
  **never** be silently tolerated in production — it should surface loudly.
- A bug in the compilation logic (unexpected profile shape, etc.) is a legitimate reason
  to degrade gracefully and leave the placeholder.
- The current code conflates both, masking the more serious case.

## Proposed Fix

Separate the `get()` call from the compilation:

```ts
if (executiveProfileService) {
  // get() throws only if initialize() was never awaited — that is a fatal startup
  // error and should surface loudly, not be swallowed.
  const profile = executiveProfileService.get();
  try {
    effectiveSystemPrompt = effectiveSystemPrompt.replace(
      '${executive_voice_block}',
      compileWritingVoiceBlock(profile, executiveDisplayName ?? 'the executive'),
    );
  } catch (err) {
    logger.error({ err, agentId }, 'Failed to compile executive voice block — placeholder left in prompt');
  }
}
```

This also takes advantage of the now-exported `compileWritingVoiceBlock` pure function
(added in PR #507).

## Acceptance Criteria

- [ ] `executiveProfileService.get()` is called **outside** the try/catch so an
  uninitialized service surfaces as an unhandled exception (fails the task, not silently)
- [ ] The try/catch wraps only the `compileWritingVoiceBlock` compilation call
- [ ] The pure `compileWritingVoiceBlock(profile, executiveName)` import is used directly
  rather than the instance method (removes one layer of indirection)
- [ ] A unit test covers the "uninitialized service → throws" path at the runtime level,
  or the existing integration tests confirm the behavior
- [ ] All existing tests continue to pass

josephfung:
### Motivation
- Prevent a startup wiring error from being silently swallowed when compiling the `${executive_voice_block}` placeholder so uninitialized service usage surfaces as a task failure. 
- Allow prompt compilation errors to be handled gracefully while still signaling operator-visible errors in the logs.

### Description
- Call `executiveProfileService.get()` outside the `try/catch` so an uninitialized `ExecutiveProfileService` throws and fails the task instead of being swallowed. 
- Use the exported pure `compileWritingVoiceBlock(profile, executiveName)` function directly instead of the instance method indirection. 
- Narrow the `try/catch` to wrap only the compilation call so only compilation errors are logged and the task continues. 
- Add unit tests in `tests/unit/agents/runtime.test.ts` covering the uninitialized-service → task failure path and the compilation-throws → log-and-continue path.

### Testing
- Attempted to run `pnpm vitest run tests/unit/agents/runtime.test.ts` but the environment-level `pnpm` install step failed due to the local build approval policy (`ERR_PNPM_IGNORED_BUILDS`).
- Ran `./node_modules/.bin/vitest run tests/unit/agents/runtime.test.ts` and the file passed: `1 passed (1)` test file, `39 passed (39)` individual tests.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a077188f10083239234bbbd4bf7f37a)

### DIFF
--- a/src/agents/runtime.ts
+++ b/src/agents/runtime.ts
@@ -17,7 +17,7 @@ import { DEFAULT_ERROR_BUDGET, type AgentError, type ErrorBudget } from '../erro
 import { AutonomyService } from '../autonomy/autonomy-service.js';
 import { formatTimeContextBlock } from '../time/time-context.js';
 import type { OfficeIdentityService } from '../identity/service.js';
-import type { ExecutiveProfileService } from '../executive/service.js';
+import { compileWritingVoiceBlock, type ExecutiveProfileService } from '../executive/service.js';
 import { formatBullpenContext, type BullpenService } from '../memory/bullpen.js';
 import type { AgentRegistry } from './agent-registry.js';
 
@@ -242,10 +242,13 @@ export class AgentRuntime {
     // The executive's display name comes from the contact system (not the profile)
     // so that identity data has a single source of truth.
     if (executiveProfileService) {
+      // get() throws only when initialize() was never awaited — this is a startup
+      // programming error and should fail loudly, not be swallowed.
+      const executiveProfile = executiveProfileService.get();
       try {
         effectiveSystemPrompt = effectiveSystemPrompt.replace(
           '${executive_voice_block}',
-          executiveProfileService.compileWritingVoiceBlock(executiveDisplayName ?? 'the executive'),
+          compileWritingVoiceBlock(executiveProfile, executiveDisplayName ?? 'the executive'),
         );
       } catch (err) {
         logger.error({ err, agentId }, 'Failed to compile executive voice block — ${executive_voice_block} placeholder left in prompt');
--- a/tests/unit/agents/runtime.test.ts
+++ b/tests/unit/agents/runtime.test.ts
@@ -7,6 +7,7 @@ import type { ExecutionLayer } from '../../../src/skills/execution.js';
 import { createLogger } from '../../../src/logger.js';
 import { WorkingMemory } from '../../../src/memory/working-memory.js';
 import type { AgentError } from '../../../src/errors/types.js';
+import type { ExecutiveProfile } from '../../../src/executive/types.js';
 
 // Minimal provenance block for mock LLM responses — satisfies the required field
 // without tying tests to specific model names.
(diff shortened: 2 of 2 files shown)


josephfung (commented):
## Overview

This PR fixes a silent error-swallowing bug in `AgentRuntime`. Previously, if `ExecutiveProfileService` was passed but never initialized, the `get()` throw would be caught and logged as a compilation error — execution would continue with the unfilled placeholder in the prompt, invisibly broken. The fix moves `get()` outside the `try/catch` so an uninitialized service propagates and fails the task loudly.

## Production code

The change is small, surgical, and correct:

- `get()` is outside the `try/catch` — a startup programming error surfaces as a task failure rather than a swallowed warning
- Switching to the pure exported `compileWritingVoiceBlock(profile, name)` removes instance method indirection — simpler call site
- The comment on the guard explains the *why* clearly

No issues with the production change.

## Tests

**Test 1 — uninitialized service path**: Solid. Explicitly mocks `get()` to throw, verifies the task response is an error and `provider.chat` is never called.

**Test 2 — compilation-throws path**: Has a hidden fragility. The test constructs `{} as ExecutiveProfile` and relies on `compileWritingVoiceBlock` throwing a `TypeError` when it accesses `.writingVoice.tone` on `undefined`. That works today, but if `compileWritingVoiceBlock` is ever hardened to guard against missing fields, the function won't throw, the `loggerErrorSpy` assertion will fail confusingly, and the test silently stops covering the intended path.

The more durable approach is to mock the module-level function explicitly so the throw is injected rather than accidental:

```ts
vi.mock('../../../src/executive/service.js', async (importOriginal) => {
  const mod = await importOriginal<typeof import('../../../src/executive/service.js')>();
  return {
    ...mod,
    compileWritingVoiceBlock: vi.fn(() => { throw new Error('compile error'); }),
  };
});
```

**Minor nit**: Both new tests use verbose inline dynamic imports for the type cast (`as unknown as import('...').ExecutiveProfileService`). Since `ExecutiveProfileService` is already referenced elsewhere in the test file, adding it to the top-level type imports would be cleaner.

## Summary

| Area | Status |
|---|---|
| Production code correctness | ✅ Correct |
| Error surfacing behaviour | ✅ Fixed |
| Test 1 (uninitialized guard) | ✅ Solid |
| Test 2 (compilation-throws) | ⚠️ Fragile — worth a follow-up |

Production fix is correct and the test fragility is a follow-up concern, not a blocker.

PR #614, opened 2026-05-16
docs: fix migration number and conversation ID format in spec 17

josephfung:
## Summary

Fixes two documentation inaccuracies in the spec 17 (meeting debrief) docs before implementation begins.

**Migration number:** The plan referenced migration prefix `029`, but the latest on main is `041`. Updated to `042`.

**Conversation ID format:** The spec and plan documented Signal conversation IDs as `signal:<phone>:<thread>`, but the actual Signal adapter produces `signal:<phone>` for 1:1 and `signal:group=<base64>` for groups. There is no `:<thread>` component — Signal doesn't have explicit thread IDs. Updated the spec example JSON and the plan's implementer notes to match reality. The claims model works correctly with the actual format (all 1:1 messages from the same phone share one conversation ID).

## Test plan

- [ ] Verify no other references to `029` or `signal:<phone>:<thread>` remain in the docs
- [ ] Confirm migration 042 prefix is still available at implementation time (rebase hazard)

### DIFF
--- a/docs/specs/17-meeting-debrief.md
+++ b/docs/specs/17-meeting-debrief.md
@@ -86,7 +86,7 @@ The agent's `agent_tasks.progress` JSON field tracks:
   "pendingDebriefs": {
     "nylas_event_abc": {
       "promptedAt": "2026-04-28T14:05:00Z",
-      "conversationId": "signal:ceo:xyz",
+      "conversationId": "signal:+15550001111",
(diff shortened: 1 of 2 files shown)


josephfung:
Superseded by #617 — the spec sections this PR corrected have been rewritten as part of the context bridging v2 redesign.

PR #639, opened 2026-05-20
fix: migration 042 — cancel stale T2125 declarative job after cron change

josephfung:
## Summary

- Adds migration `042_cancel_stale_t2125_enrichment_cron.sql` to cancel the `*/30 * * * *` declarative job row for `T2125-expense-tracker` that was left in `pending` status when the enrichment cron was updated.

**Why this is needed:** `upsertDeclarativeJob` matches on `(agent_id, cron_expr, task_payload::text)`. Changing only `cron_expr` creates a new row rather than updating the old one. Without this migration both expressions fire in parallel indefinitely.

The migration is idempotent — if the old row was already cancelled or never existed (fresh install), the UPDATE touches zero rows and succeeds.

**Deploy order:** merge/apply this migration alongside or before the companion curia-deploy PR that changes the cron expression.

## Test plan

- [ ] Confirm migration applies cleanly on a DB with the old row present — `status` flips to `cancelled`, no error
- [ ] Confirm migration is a no-op on a fresh DB — zero rows affected, no error
- [ ] After both PRs are deployed: confirm only the `*/30 6-17 * * *` row is in `pending` status for `T2125-expense-tracker`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`T2125-expense-tracker` scheduler** — migration 042 cancels the stale `*/30 * * * *` declarative job row left behind when the enrichment cron was narrowed to the 6 am–6 pm window; without this, both expressions fire in parallel. (#623)
+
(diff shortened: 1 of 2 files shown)


josephfung:
Dropping in favour of a manual cancel via the web UI — migration not needed.

Issue #721, opened 2026-05-26
bullpen skill: timeout semantics are unsafe and need a redesign


josephfung:
## Summary

The `bullpen` skill's 10-second timeout (`skills/bullpen/skill.json:23`) creates a false-negative failure mode: a `post` action that times out has **already created the thread, persisted the message, published the discuss event, and woken every mentioned agent** — but the caller sees `<skill_error>Skill 'bullpen' timed out after 10000ms</skill_error>` and treats it as a clean failure to be retried. This caused a duplicate-Signal-message incident on 2026-05-26.

## Incident evidence

Today's `meeting-debrief` cron fire at 11:45 UTC discovered a Coffee With Nadiya event and tried to post a Bullpen thread to coordinator asking for a Signal debrief prompt. Five sequential `bullpen.post` invocations all timed out at exactly 10 seconds, but each one **fully succeeded** in creating a thread:

| t (UTC) | event | thread |
|---|---|---|
| 11:49:50.228 | `agent.discuss` published, coordinator mentioned | `078aa818…` |
| 11:50:00.224 | **`skill.result` returns** `Skill 'bullpen' timed out` | (same) |
| 11:50:16.492 | Retry → new `agent.discuss` published | `3b9cd799…` |
| 11:50:26.489 | Timeout | (same) |
| 11:50:42.980 | Retry → new thread | `8fc1e57b…` |
| 11:50:52.978 | Timeout | (same) |
| …5 threads total… |

Each thread independently woke the coordinator; the coordinator's reply task fired `signal-send`. Net effect: **4 identical Signal messages to the CEO**. Coordinator's own internal narration confirmed the cause: *"All five duplicate bullpen threads from meeting-debrief were asking the same thing"*.

## Code references

- `skills/bullpen/skill.json:23` — `"timeout": 10000`
- `skills/bullpen/handler.ts:71-101` — `post` action: `openThread()` + `bus.publish(agent.discuss)`, returns `{ thread_id, message_id }`. **No reply-waiting; the handler is conceptually fire-and-forget.**
- `src/skills/execution.ts:770-825` — generic `Promise.race(handler.execute, timeout)`. On timeout, the in-flight handler is *not* cancelled — it keeps running and any side-effects it has already committed (DB row, bus publish) persist.

## Why this is hard (deeper thoughts)

[...]

PR #732, opened 2026-05-26
fix: meeting-debrief no longer retries bullpen on timeout (#722)

### LINKED ISSUE #722: meeting-debrief: do not retry bullpen on timeout
## Summary

When `meeting-debrief` invokes `bullpen.post` and gets back `<skill_error>Skill 'bullpen' timed out after 10000ms</skill_error>`, the LLM-driven agent currently retries — creating a duplicate Bullpen thread on each retry. Because timed-out `bullpen.post` calls **still create the thread** (see #721), each retry produces one additional Signal message to the CEO.

## Evidence

Same incident as #721. 2026-05-26 11:49–11:51 UTC: five duplicate Bullpen threads created in 70 seconds, all from `meeting-debrief`, all with topic "Debrief prompt: Coffee With Nadiya". Net: 4 duplicate Signal messages.

## Root cause

`agents/meeting-debrief.yaml:228-262` instructs the agent to "open a Bullpen thread" but says nothing about timeout handling. The `error_budget.max_turns: 25` (line 42) gives plenty of room for retry loops. When the LLM sees `<skill_error>...timed out</skill_error>`, "retry" is its default heuristic.

## Fix

Add an explicit instruction to the system prompt around line 244 (right after the `bullpen` call instructions), e.g.:

```yaml
  ### Bullpen timeout handling

  If a bullpen `post` call returns a timeout error, DO NOT retry. The thread
  may have been created on the server side even though you saw a timeout —
  retrying will create a duplicate thread and the CEO will receive duplicate
  prompts. Instead: record the meeting in `pendingDebriefs` with
  status="prompt_unconfirmed" and continue to the next meeting. On the next
  tick, scheduler-list will show whether a thread exists; reconcile then.
```

This is a stopgap until #721 (proper contract) lands. Cheap to do, narrow scope.

## Acceptance criteria

- [ ] System prompt instructs the agent to not retry on `bullpen` timeout.
- [ ] An offline LLM eval (input: timeout error after a `post`) shows the agent moves on rather than retrying.
- [ ] `CHANGELOG.md` entry under **Changed** referencing the incident.

josephfung:
Closes #722

## Summary

- Adds a **Bullpen timeout handling** subsection to the `meeting-debrief` system prompt telling the LLM not to retry `bullpen.post` on `<skill_error>...timed out</skill_error>`. The thread is created server-side even on timeout (the #721 contract bug), so retrying produces duplicates (5 threads in 70s on 2026-05-26).
- Agent now records the meeting with status `prompt_unconfirmed` and stops. It does **not** re-post on subsequent ticks — the bullpen skill has no `list`/`search` action, so any reconciliation that "checks if a thread exists" is currently unimplementable. Re-posting would risk the exact duplicate this rule prevents.
- Step D5 reply-handling now transitions to `completed` from either `prompted` or `prompt_unconfirmed`, so a CEO reply on the "ghost" thread still resolves cleanly.
- Step 8 `scheduler-report` summary now calls out the `prompt_unconfirmed` count so the situation is observable.

Stopgap until #721 lands a proper post contract.

## Acceptance criteria

- [x] System prompt instructs the agent to not retry on `bullpen` timeout.
- [x] CHANGELOG.md entry under **Changed** referencing the incident.
- [ ] Offline LLM eval (input: timeout error after a `post`) shows the agent moves on rather than retrying.

The third criterion is partially addressed by a regression-guard unit test asserting the prompt contains the no-retry instruction (`tests/unit/agents/loader.test.ts`). A full behavioral LLM eval would require extending the smoke harness to support scheduled-mode task injection with mocked skill responses — out of scope for this stopgap. Once #721 lands the proper post contract, the eval should target the new contract directly.

## Trade-off

A genuinely-failed `post` (no thread created server-side) will now silently miss the CEO prompt until the entry is pruned at TTL. Accepted because:
- A missed prompt is recoverable (CEO can ask about the meeting; debrief preferences can be set per-recurring-meeting).
- A duplicate prompt train is not (the CEO got 4 duplicate Signal messages in 70 seconds in the original incident).
- The `prompt_unconfirmed` status + Step 8 summary makes the situation observable.

## Test plan

- [x] `tests/unit/agents/loader.test.ts` — new regression test asserts the prompt contains "Bullpen timeout handling" + a no-retry instruction + `prompt_unconfirmed` status.
- [x] `pnpm test tests/unit/agents/` passes (91 tests).
- [x] `pnpm run typecheck` clean.
- [ ] After deploy: monitor the next meeting-debrief timeout incident — confirm no duplicate threads and `prompt_unconfirmed` count appears in scheduler logs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ## [Unreleased]
 
+### Changed
+
+- **`meeting-debrief`** — system prompt now instructs the agent not to retry `bullpen.post` on timeout; on `<skill_error>...timed out</skill_error>` it records the meeting as `prompt_unconfirmed` and reconciles on the next tick. Stopgap until #721 lands a proper contract. (#722)
+
 ### Fixed
 
 - **Email reply quoting** — natural agent-response replies (no skill invocation) now include the quoted original message, matching the skill-driven paths. `buildReplyQuote` moved to `src/skills/_shared/` so the email channel adapter can share it. (#720)
--- a/agents/meeting-debrief.yaml
+++ b/agents/meeting-debrief.yaml
@@ -264,6 +264,37 @@ system_prompt: |
   entry for this meeting:
   - title, endedAt, promptedAt (now), attendees, status: "prompted"
 
+  ### Bullpen timeout handling
+
+  If a bullpen `post` call returns a timeout error (e.g.
+  `<skill_error>Skill 'bullpen' timed out after 10000ms</skill_error>`),
+  DO NOT retry the call. The thread may have been created on the server
+  side even though you saw a timeout — retrying will create a duplicate
+  thread and the CEO will receive duplicate prompts. Instead: record the
+  meeting in `pendingDebriefs` with status "prompt_unconfirmed" (title,
+  endedAt, attendees, promptedAt = now, reminderSentAt = null) and
+  continue to the next meeting.
+
+  On subsequent ticks, do NOT re-post for `prompt_unconfirmed` entries —
+  there is no skill action to query Bullpen for an existing thread, so
+  re-posting would risk the duplicate the rule above is meant to prevent.
+  Leave the entry as-is. One of two things will happen:
+  - The CEO replies (because the thread did get created server-side) —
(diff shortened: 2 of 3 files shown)


josephfung:
@coderabbitai review

PR #1142, opened 2026-06-23
fix(ceo-inbox): archive done-state emails in the overflow path (#1123)

### LINKED ISSUE #1123: fix: ceo-inbox overflow path skips archive — emails labeled but left in inbox
## Summary

When the inbox has more than the overflow threshold (10) unread messages, the agent splits handling into two paths: inline triage for the first N messages, and task-creation for the rest. The inline path correctly calls `ceo-inbox-archive`. The overflow/deferred path does not — it calls `task-create`, `ceo-inbox-mark-read`, and `ceo-inbox-label`, but never `ceo-inbox-archive`. The result is emails with disposition labels (e.g. "✔️ Cleared") sitting in the inbox indefinitely.

## Root Cause

Observed in the 2026-06-22 10:00 AM run (`conversation_id` ending in `95f7942b`). 13 unread dependabot PR notifications were processed:

- **First 5 (inline triage):** `ceo-inbox-archive` → `ceo-inbox-mark-read` → `ceo-inbox-label "✔️ Cleared"` ✅
- **Last 8 (overflow-deferred):** `task-create` → `ceo-inbox-mark-read` → `ceo-inbox-label "✔️ Cleared"` — **no archive** ❌

All `ceo-inbox-archive` calls return `success: true` when made — the API works fine. The step is simply absent from the overflow path.

Because the emails are also marked read, subsequent runs using `unread_only: true` will never see them again. The tasks created contain no message IDs and have no mechanism to archive the source emails. The 8 affected emails will remain in the inbox permanently unless manually archived.

## Scope

This is not just a "✔️ Cleared" problem. The same gap applies to any disposition the agent applies in overflow mode:

| Disposition | Should archive? | Notes |
|---|---|---|
| ✔️ Cleared | Yes | Done — no inbox presence needed |
| ✅ Handled | Yes | Reply sent — email is resolved |
| 📌 Seen | No | Needs the CEO's attention |
| ⚠️ Stuck | No | Blocked, awaiting input |
| 🚨 Urgent | No | High priority, needs action |
| ✍️ Drafted | No | Draft pending review/send |

The overflow deferred path should apply the same archival logic as the inline path — archive immediately when the classification is a "done" state, regardless of whether the underlying action (e.g. review a PR) is deferred to a task.

## Fix

Decouple "archive the email" from "triage the email inline". The archive step should fire at **classification time** for any done-state disposition, not only when the full triage happens inline.

Options:
1. **In the agent prompt / overflow instructions:** Explicitly instruct the agent that archiving is always required for Cleared/Handled emails, even in overflow mode. The task captures the action; the archive cleans the inbox.
2. **In the `ceo-inbox-archive` skill wrapper:** After `task-create` for a Cleared/Handled classification, automatically call archive (i.e. make it a post-step of the overflow skill path rather than leaving it to the LLM).

Option 2 is more robust — it removes the LLM as a point of failure for the archive step entirely.

## Acceptance Criteria

- [ ] Overflow-deferred emails classified as ✔️ Cleared or ✅ Handled are archived at classification time, before or immediately after the task is created
- [ ] Overflow-deferred emails classified as 📌 Seen, ⚠️ Stuck, 🚨 Urgent, or ✍️ Drafted remain in the inbox (no regression)
- [ ] Inline-triaged emails continue to be archived as before (no regression)
- [ ] Audit log contains a `skill.invoke` for `ceo-inbox-archive` for every email that receives a done-state label in any processing path
- [ ] A batch of 15+ unread emails can be processed with no emails left in the inbox that carry a Cleared or Handled label

josephfung:
## Summary

Fixes #1123. The ceo-inbox overflow path (>10 unread) created a deferred task and marked each message read but **never archived** it. Emails classified as a "done" state (✔️ Cleared / ✅ Handled) therefore stayed in the inbox permanently — and because they were also marked read, later `unread_only` runs never saw them again.

The root cause is purely in the agent's `system_prompt` (`agents/ceo-inbox.yaml`): the overflow loop deferred *everything* to a task without ever classifying or archiving. This reworks that loop so archiving is bound to the **classification**, not the processing path.

## What changed

`agents/ceo-inbox.yaml` overflow loop (step 3) now makes a lightweight done-state vs keep-in-inbox decision per deferred message (subject/sender/snippet only — no body read, no entity-context, no drafting):

- **✔️ Cleared** → archive + mark-read, **no task** (nothing to defer).
- **✅ Handled** → create a deferred task (the handoff is deferred to the task, not a live bullpen thread, so the action isn't lost), then archive + mark-read.
- **📌 Seen / ✍️ Drafted / 🚨 Urgent** → deferred task, **stay in the inbox** (no archive). "When in doubt, keep-in-inbox."

Added an explicit **Archive invariant**: any email assigned a done-state disposition MUST be archived via `ceo-inbox-archive` in *any* path (inline or overflow); a 📌/✍️/🚨/⚠️ message is never archived.

Also:
- mark-read is now gated on archive success — a failed archive leaves the email unread/visible (it won't re-appear because the watermark already advanced).
- New `Overflow archived: N` line in the run summary for observability.
- `version` 0.10.0 → 0.10.1, CHANGELOG entry under Fixed.

## Acceptance criteria

- [x] Overflow ✔️ Cleared / ✅ Handled archived at classification time
- [x] Overflow 📌 Seen / ⚠️ Stuck / 🚨 Urgent / ✍️ Drafted remain in the inbox
- [x] Inline-triaged emails continue to be archived (step 4h unchanged)
- [x] `skill.invoke` for `ceo-inbox-archive` fires for every done-state-labelled email in any path (enforced by the Archive invariant)
- [x] A 15+ unread batch leaves no Cleared/Handled-labelled email in the inbox

## Testing

Prompt-only change; no executable code touched. `tests/unit/agents/loader.test.ts` + `tests/unit/startup/validator.test.ts` pass (42 tests), confirming the agent config still loads and validates.

## Reviewer note

In overflow, ✅ Handled becomes a deferred CEO-owned task rather than a live specialist bullpen handoff (the load-shedding tradeoff — the email leaves the inbox, the task is the durable record). This matches the issue's "the task captures the action; the archive cleans the inbox" model. If you'd prefer overflow ✅ Handled to stay in the inbox instead (no archive until a real handoff happens), that's a one-line change — flag it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 - **Bullpen long-thread context** — agents woken by a new reply in a thread with more than 15 messages now always see the original request (first message) in their ambient context, alongside the 14 most recent messages; the first message no longer silently falls off the window. (#1090)
 - **Scheduler duplicate-fire race condition** — claim-time `next_run_at` advancement and null-safe `rowCount` guard prevent duplicate cron fires. (#1124)
+- **ceo-inbox overflow archive gap** — the overflow path now classifies each deferred email and archives done-state ones (✔️ Cleared / ✅ Handled) instead of leaving them in the inbox; archiving is now an invariant of the classification, not the processing path. (#1123)
 
 ### Security
 
--- a/agents/ceo-inbox.yaml
+++ b/agents/ceo-inbox.yaml
@@ -1,5 +1,5 @@
 name: ceo-inbox
-version: "0.10.0"
+version: "0.10.1"
 role: specialist
 description: >
   Manages the CEO's personal email inbox. Handles standing-rule management,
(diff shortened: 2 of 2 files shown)


josephfung:
Superseded by #1145, which removes overflow mode entirely (the path this PR patched) in favor of read-status batch draining. Closing.

Issue #1125, opened 2026-06-22
Persist TaskOriginator on tasks; restore woken-task authorization via score-keyed bypass inheritance


josephfung:
> **Part of #1060 — step 1 of 3 (foundation).** This is the prerequisite for the other two follow-ups; both build on the lineage-on-task model and effective-standing computation introduced here.
>
> **Depends on:** — (none)
> **Blocks:** #1126 (redefine `elevated` / skill audit), #1127 (console originator stamping)

## Summary

Implements the woken-task half of the authorization model decided in the design note (resolves the core of #1060). Woken/deferred tasks currently lose all authorization provenance because (1) `TaskOriginator` is not persisted on the `tasks` row, and (2) `SchedulerService.enqueueTaskWake()` inserts a `scheduled_jobs` row without an `originator`. This issue fixes the data-model gap and introduces the **score-keyed inheritance ("bypass ladder")** that governs how much of a chain's standing a woken/derived execution retains.

This issue deliberately does **not** redefine the `elevated` gate or reclassify skills — that is #1126, which must land atomically with the skill reclassification to avoid regressing legitimate non-principal callers.

## Why the naive fix is wrong

Simply writing the original `originator` onto the wake (the obvious one-line fix) is the privilege-escalation-across-task-boundaries hazard #1060 warns about: it would let woken/derived contexts ride borrowed principal/system standing, and would re-open broad `elevated` access once the woken task fires with a populated `originator`. The gate reads must therefore consume **effective (post-ladder) standing**, not raw lineage. The ladder *is* the safe form of threading.

## Model (from the design note)

- **Lineage** — the chain's original `TaskOriginator`, stamped on the `tasks` row at creation, immutable. Pure audit + the *ceiling* for what may be inherited.
- **Effective standing** — computed at skill-invocation time from `(lineage, live autonomy score, execution context)`. The score can only ever **downgrade** standing, never upgrade it; a chain can never acquire standing its lineage didn't have.
- **Bypass ladder** (thresholds in autonomy config; defaults below):
  - `< 70` (posture A) — woken & derived → downgrade to agent (propose-only)
  - `70–89` (posture B) — same-task heartbeat wake keeps lineage standing; child task → agent
  - `≥ 90` (posture D) — same-task wake **and** child task keep lineage standing
- Pre-authorized *specific* deferred actions remain on the **scheduler** path (already preserves originator at `scheduler.ts:419`); the heartbeat ladder governs open-ended backlog wakes.

## Scope / changes

[...]

Issue #1060, opened 2026-06-19
Investigate: enqueueTaskWake drops task originator — woken tasks lose authorization provenance


josephfung:
## Summary (investigation / design needed)

`SchedulerService.enqueueTaskWake()` writes its `scheduled_jobs` row **without an `originator`**, so every heartbeat-woken task fires with no task provenance. Any elevated skill invoked from a woken task is then blocked, regardless of how the originating task was created. The *right* fix is non-obvious and has security trade-offs in both directions — hence this is an investigation issue, not a straightforward patch.

## How it surfaces

Discovered while investigating why the contacts specialist "can't perform admin merges" on a scheduled dedup review:

1. Weekly dedup scan (system-originated cron) runs `contact-find-duplicates` → files a `owner: 'curia'` review task. It does **not** merge (by design — contacts.yaml:197).
2. `BacklogHeartbeat` later wakes that review task via `enqueueTaskWake()`.
3. The INSERT (scheduler-service.ts:988-990) **omits the `originator` column** → `NULL`.
4. Scheduler fires it with `metadata: job.originator ? {...} : undefined` (scheduler.ts:401) → `undefined`.
5. Agent calls `contact-merge` → elevated gate (execution.ts:374) sees neither `isPrincipalOriginated` nor `isSystemOriginated` → blocked.

For the dedup case this is *masked* — `contact-merge` is supposed to require CEO confirmation anyway (contacts.yaml:166, 175-176), so the intended completion path is "agent asks CEO → CEO's reply is principal-originated → merge runs." But the underlying provenance drop is general.

## The latent general bug

Provenance is dropped for **all** woken tasks, so the failure is not dedup-specific:

- A **principal**-originated task that gets deferred (e.g. "email X tomorrow, once Y arrives") and is later re-woken by the heartbeat loses its principal rights and cannot run an elevated skill it was legitimately authorized for.
- A **system**-originated declarative job that defers work into a follow-up task loses its `system` standing on the wake.

The current failure mode is *over-restriction* (fails safe), but it silently breaks legitimate deferred elevated work.

## Why the obvious fixes are wrong

(Framing from @josephfung — captured so the discussion starts here, not from scratch.)

- **Naively thread the original originator through the wake** → risks cascading tasks inheriting rights they shouldn't. A task spawned by a principal-originated task is not itself principal-authorized; blanket inheritance would let agent-invented follow-ups ride on borrowed principal/system standing — a privilege-escalation-across-task-boundaries hazard.
[...]

PR #1156, opened 2026-06-24
feat: redefine 'elevated' as live-principal authority; audit & reclassify all elevated skills (#1126)

### LINKED ISSUE #1126: Redefine 'elevated' as live-principal authority; audit & reclassify all elevated skills
> **Part of #1060 — step 2 of 3.**
>
> **Depends on:** #1125 (the effective-standing computation must exist first). Do not start until #1125 has merged.
> **Blocks:** — (none)

## Summary

Completes the authorization model from the design note. Redefines `sensitivity: elevated` to mean **"requires a live principal turn"** — the current turn originated from a fresh principal inbound; never satisfiable by system, agent, or any inherited/woken standing — and **audits & reclassifies all 19 currently-`elevated` skills**, since the label today conflates three distinct intents. The gate change and the reclassification must land **atomically** (same PR or tightly sequenced) to avoid regressing legitimate non-principal callers (e.g. agent self-scheduling via `scheduler-create`).

## Why

- `elevated` was widened from principal-only to principal-or-system in `3bd3d224` **solely** to let a scheduled digest job call the read-only `list-pending-actions`. That over-generalized: it now lets any system-lineage context reach genuine authority primitives.
- Handler-level `isPrincipalOriginated` re-checks (and a now-stale comment on `contact-merge` claiming the gate is principal-only) duplicate the *old* definition and have drifted out of sync with the gate. The gate must be the single enforcement point — no handler re-checks.
- "Live principal" closes the **self-approval hole** with zero per-skill exceptions: a woken principal-*lineage* task can never approve its own pending action, because lineage ≠ a live turn.

## Taxonomy (rubric for the audit)

- **Stay `elevated` (live-principal authority primitive):** `approve-action`, `deny-action`, `dismiss-action`, `set-autonomy`, `approve-grant-recommendation`, `decline-grant-recommendation`.
- **→ `normal` + `action_risk` (autonomy-governed; inherits the ADR-018 approval flow for free):** `contact-merge` *(the auto-at-high-trust vs. always-confirm decision is made here)*, `contact-update`, `contact-rename`, `behavioral-preferences-update`, `executive-profile-update`, `scheduler-create`, `scheduler-cancel`.
- **→ `normal` + `action_risk: none` + `allowed_callers`:** `list-pending-actions` (a sensitive *read* — gate on *who may call*, not *who originated*; the digest job runs as the coordinator).
- **Decide explicitly (Bucket D):** `contact-grant-permission`, `contact-revoke-permission`, `contact-set-tier`, `contact-set-role` (these alter authorization itself — authority-primitive vs. high-risk mutation?); `web-browser` (an egress/injection gate, not authority or a mutation — likely `allowed_callers` + `action_risk`).

## Scope / changes

- Redefine the elevated gate (`execution.ts:553`) to require **live principal**; introduce the "live principal turn" signal (stamped only on fresh principal inbounds — never on wakes, derivations, or scheduler fires).
- Remove **all** handler-level standing/origination re-checks and the stale `contact-merge` comment.
- Set `action_risk` (and `allowed_callers` where applicable) on every reclassified skill, with `version` bumps per manifest.
- Resolve the Bucket-D borderlines (decision recorded in the PR).
- Update ADR-011, ADR-017 (+ ADR-018 cross-ref), spec 03, spec 14 to match shipped behavior.

## Acceptance criteria

- [ ] `elevated` is satisfiable only by a live principal turn; system, agent, and woken/inherited principal-lineage contexts all fail it. Covered by tests.
- [ ] No skill handler contains a standing/origination re-check; the gate is the sole enforcement point.
- [ ] Every one of the 19 currently-`elevated` skills is explicitly classified (table in PR), with `sensitivity`/`action_risk`/`allowed_callers` set accordingly and `version` bumped.
- [ ] `list-pending-actions` remains callable by the 8am digest job after moving to `normal` + `allowed_callers` (verified).
- [ ] `approve-action` / `deny-action` / `dismiss-action` cannot be invoked by a non-live-principal context (self-approval test).
- [ ] Bucket-D skills each have a documented decision.
- [ ] ADR-011 / ADR-017 (+018 xref) and specs 03 / 14 updated.

## References

- Design note: `docs/wip/2026-06-22-woken-task-authorization-design.md`
- Builds on #1125; #1060 (investigation); commit `3bd3d224` (the gate-widening this corrects)
- `src/skills/execution.ts:541,553`, `src/contacts/principal.ts`, `skills/*/skill.json`

josephfung:
## Summary

Completes the authorization model from the woken/derived-task design note (#1060, step 2 of 3; depends on #1125 which merged). Redefines `sensitivity: 'elevated'` to mean **"requires a live principal turn"** and audits & reclassifies the formerly-`elevated` skills. The gate change and reclassification land atomically.

Closes #1126.

## The live-principal signal — a distinct field threaded through synchronous delegation

`liveTurn` is a **distinct field** on the `agent.task` payload (→ `InvokeOptions.liveTurn` → `SkillContext.liveTurn`), deliberately **not** a key in the `taskMetadata` bag. This matters: the persistence skills (`scheduler-create`, `task-create`, `bullpen`, the wake path) forward `originator`/`metadata` *by name*, so keeping the signal off the bag makes **"never persisted" a structural guarantee**, not a per-skill discipline.

- The **dispatcher computes** it (`originator.systemRole === 'principal'`) — never copies it from inbound input, so it can't be forged from a message.
- The **`delegate` skill forwards it** across a *synchronous* delegation. So **"the CEO is live" = the whole synchronous call tree**: a specialist delegated-to inside the CEO's turn inherits it; a wake, scheduler fire, or persisted task never does (those construct their own `agent.task` without it).
- The gate is `isLivePrincipalTurn(options.liveTurn, effectiveTaskMetadata)` — requires **both** the strict-boolean flag **and** principal lineage (defence in depth). It is the **single enforcement point**; the handler-level origination re-checks on elevated skills were removed.

This closes the **self-approval hole** with zero per-skill exceptions: a woken principal-*lineage* task can never approve its own action, change autonomy, alter authorization, or mint a secret-capture link.

## Final classification

**`elevated` (live principal turn — CEO authority primitives):**
- approve-action, deny-action, dismiss-action, set-autonomy, approve/decline-grant-recommendation
- **contact-set-tier, contact-set-role, contact-grant-permission, contact-revoke-permission** — they alter *who-can-do-what*; require a live CEO, never autonomous. Pinned to the delegated `contacts` specialist, which works *because* the live signal threads through synchronous delegation.
- **system-secret-capture-request** — a secret-capture link can never be minted by a schedule/job. Used by the delegated `setup-wizard` (verified to run only on live CEO turns).

**`normal` + `action_risk` (autonomy-governed; inherit ADR-018 surface-and-confirm):**
- contact-merge (medium), contact-update/contact-rename/behavioral-preferences-update/executive-profile-update (low/medium), scheduler-create/scheduler-cancel (low)

**`normal` + `allowed_callers: [coordinator]`:** list-pending-actions (none), web-browser (medium). The 8am digest reads `list-pending-actions` because the coordinator is the caller.

### Bucket-D resolution (the key decision)

The 4 authorization-altering contact skills are **authority primitives, not mutations** — Curia should never change a contact's tier/role/permissions autonomously, even at max trust. The original blocker to classifying them `elevated` was that they're delegated (the contacts specialist's derived turn had no live signal). Threading the signal through synchronous delegation removes that blocker, so they're now correctly `elevated`. The grant-*recommendation* flow (`approve-grant-recommendation`, also elevated) is the "Curia proposes, CEO approves" path, so direct grants being live-only loses nothing.

## Handler re-checks: removed vs. retained

- **Removed** (gate / allowed_callers now covers them): approve/deny/dismiss-action, approve/decline-grant-recommendation, list-pending-actions, contact-set-tier; plus the stale contact-merge comment.
- **Retained** (ADR-017 Option C, NOT among the elevated set): `send-draft`, `calendar-list-events` — both `action_risk: 'none'` `normal` skills the autonomy gate cannot govern, so their handler origination checks are load-bearing.

## Tests

- `isLivePrincipalTurn` (both-required, forgery/downgrade defence, strict boolean).
- Execution gate: live turn passes; system/agent/woken-lineage blocked at any score; forged flag without principal lineage fails.
- **Dispatcher**: `liveTurn` is a distinct payload field, stamped only for the principal, computed (not copied) — a forged inbound `livePrincipal` yields `liveTurn=false` and is scrubbed.
- **Delegate**: forwards `liveTurn` on a live turn; does NOT manufacture it on a non-live turn.
- Integration (real Postgres): woken task blocked at the elevated gate at score 65 and 90; contact-merge surface-and-confirms below 70, auto at/above.

Typecheck clean, lint clean, **4317 passing** (only failure in a full local run is a pre-existing cross-test DB-cleanup flake in `setup-routes.test.ts`, which passes in isolation; CI uses a fresh DB).

## Docs

ADR-011/017/018, specs 03 & 14, CHANGELOG (Changed + Security).

## Security review

A dedicated adversarial review verified all four invariants hold *structurally*: (a) `liveTurn` cannot reach a persisted/wakeable row, (b) cannot be forged, (c) `delegate` cannot manufacture liveness on a non-live turn, (d) no gate fails open and the elevated gate is not bypassed by `humanApproved`. A misleading comment in `bullpen-dispatcher` was corrected (bullpen, being async/persisted, correctly does NOT carry the live signal).

## Trust-model notes (pre-existing, not regressions)

- The live-principal gate inherits its trust from channel/route auth (the `web`/`cli` routes assert principal at the HTTP layer).
- `normal` skills inherit the #1125 fail-direction: a *woken* task with an unreadable autonomy score fails closed; a non-woken agent task fails open (a transient DB issue must not disable a human-directed agent).


josephfung:
@coderabbitai review

josephfung:
@Coderabbitai review

docs/adr/011-score-based-autonomy-engine.md
```diff
@@ -33,3 +33,9 @@ The five band names  are human-readable aliases for score ranges, documented in
...
 - The `action_risk` field is part of the skill manifest public API — renaming it or changing its semantics is a breaking change.
+
+## Update (#1125 / #1126): the score also governs inherited standing
+
+The bypass ladder (#1125) extends this ADR's thesis: the live autonomy score now also governs how much *standing* a woken/derived task inherits from its chain's lineage. The score can only ever **downgrade** inherited standing to `agent` (propose-only) for a non-live execution — never grant standing the lineage didn't have. See `docs/specs/14-autonomy-engine.md` (Effective standing) and `docs/wip/2026-06-22-woken-task-authorization-design.md`.
+
+#1126 completed the picture by decomposing the overloaded `elevated` sensitivity: CEO-authority primitives stay `elevated` (require a live principal turn — see ADR-017), while consequential autonomous actions move to `normal` + `action_risk`, governed by *this* engine and inheriting the ADR-018 approval flow. The practical effect: `system` standing now confers no gate power, and the ladder governs exactly one thing — whether a woken/derived `normal`-skill execution inherits the principal-bypass of Gates A and B.
```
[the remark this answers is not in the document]
josephfung: Good catch — fixed in 7384bc3. Reworded the line to start with `PR #1126 completed the picture…` so markdown no longer parses the leading `#1126` as an ATX heading.

docs/adr/018-curia-initiated-approval-requests.md
```diff
@@ -310,3 +310,29 @@ send mechanism.
 - The `approve-action`, `deny-action`, and `dismiss-action` skills must be
   `sensitivity: 'elevated'` (CEO-only) to prevent non-CEO contacts from
   approving actions. Same security model as `set-autonomy`.
+
+## Update (#1126): the approval loop gains its consequential actors, and `elevated` is sharpened
+
+#1126 makes this flow do more work. Consequential *mutations* that previously hid behind the
```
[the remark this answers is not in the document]
josephfung: Fixed in 7384bc3 — same treatment as ADR-011: the paragraph now opens with `PR #1126 makes this flow do more work.` so the leading hash is no longer read as a malformed heading.

docs/specs/14-autonomy-engine.md
```diff
@@ -218,21 +218,21 @@ When a task is **principal-originated** (initiated by a direct CEO message, as d
...
+The principal-bypass consumes **effective standing**, not raw lineage. (The elevated-skill gate does not: it requires a live principal turn, which no wake or derivation ever is — #1126.) For a **live turn** (a fresh inbound, a specific scheduler-create fire) effective standing *is* the task's lineage — behaviour above is unchanged. For a **heartbeat-woken** task the score-keyed **bypass ladder** governs how much of the chain's lineage standing the woken execution inherits. The lineage (`tasks.originator`, see [19-tasks-and-backlog.md](19-tasks-and-backlog.md)) is pure audit and the *ceiling*; the live score can only ever **downgrade** it to `agent` (propose-only), never grant standing the lineage didn't have.
```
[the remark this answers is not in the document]
josephfung: Agreed, that wording contradicted the new contract. Fixed in 7384bc3. The sentence now distinguishes a **non-heartbeat execution** (fresh inbound, synchronous delegation, or a scheduler-create fire — all of which take standing = lineage) from a **live principal turn** (only fresh inbounds and synchronous delegation). A scheduler fire is explicitly called out as a scheduled autonomous execution that is never live: it can drive the lineage-based bypass but never the elevated gate.

src/dispatch/dispatcher.ts
```diff
@@ -36,6 +39,8 @@ function mergeTaskMetadata(
     ...(injectionMetadata ?? {}),
     ceoInitiated: undefined,          // strip legacy untrusted channel value (after all untrusted spreads)
     originator: undefined,            // strip untrusted channel value (after all untrusted spreads)
+    livePrincipal: undefined,         // defensive scrub — legacy alias; the real signal is the distinct payload field (#1126)
+    liveTurn: undefined,              // defensive scrub — keep the real signal OFF the metadata bag so it can never be persisted (#1126)
```
josephfung, line 43: > Scrub `metadata.liveTurn` as well as the legacy alias.

Done in 7384bc3. The scrub now clears both `livePrincipal` (legacy alias) and the real `liveTurn` key from the metadata bag before the trusted `originatorMeta` spread, keeping the off-bag contract intact — the genuine signal only ever travels as the distinct `payload.liveTurn` field, never via metadata.

tests/unit/skills/execution.test.ts
```diff
@@ -167,11 +167,13 @@ describe('ExecutionLayer', () => {
...
-    // the authorization signal is the task originator, not the immediate caller.
-    // See docs/wip/2026-05-10-principal-identity-design.md
+    // #1126: the elevated-skill gate requires a LIVE PRINCIPAL TURN — the metadata must carry
+    // a principal originator, and the invoke options must carry the dispatcher-stamped `liveTurn`
+    // signal (a distinct off-bag field). The immediate caller.role is not consulted. System/agent
+    // lineage and woken principal-lineage all fail.
+    // See docs/wip/2026-06-22-woken-task-authorization-design.md §4, ADR-017.
 
-    it('allows elevated skill when caller has ceo role', async () => {
+    it('allows elevated skill on a live principal turn (liveTurn + principal originator)', async () => {
```
josephfung, line 176: > Rename the stale `livePrincipal` wording to `liveTurn`.

Fixed in 7384bc3 — the test name and the surrounding comment now describe the current contract: a principal originator in `taskMetadata` plus the distinct top-level `liveTurn` option. No more references to the old `livePrincipal` metadata key or to `caller.role`.

src/skills/execution.test.ts
```diff
@@ -1372,10 +1393,70 @@ describe('humanApproved on InvokeOptions', () => {
...
+          contactId: 'someone',
+          systemRole: 'agent' as const,
+          channel: 'internal',
+          initiatedAt: new Date().toISOString(),
+        },
+      },
+    });
+
+    expect(result.success).toBe(false);
+    if (!result.success) expect(result.error).toContain('live principal turn');
```
josephfung, line 1456: > Assert the forgery failure reason here too.

Done in 7384bc3 — added `expect(result.error).toContain('live principal turn')` so the forgery-defence case is pinned to the elevated live-turn gate, not just to the handler not being called.

src/skills/execution.test.ts
```diff
@@ -734,56 +735,73 @@ describe('autonomy gates', () => {
...
+      if (!result.success) expect(result.error).toContain('live principal turn');
+      expect(handler.execute).not.toHaveBeenCalled();
     });
 
-    it('elevated gate blocks a principal same-task wake once the score drops below 70', async () => {
+    // Defence in depth: even if the liveTurn flag somehow leaks onto a wake (it shouldn't — it is a
+    // distinct off-bag field the wake path never sets), the gate must still reject because the task
+    // carries wakeContext. This pins the self-approval-hole closure to the gate itself rather than to
+    // the dispatcher/scheduler behaving perfectly. (#1126)
+    it('elevated gate rejects a principal same-task wake even if liveTurn leaks in', async () => {
```
josephfung, line 762: > Add the leaked-`liveTurn` wake case; this only proves the flag is absent.

Good point — addressed in 7384bc3, and it required a real hardening, not just a test. The gate now rejects any task carrying `wakeContext` outright (see `src/skills/execution.ts`), so a wake fails even if `liveTurn` ever leaks onto it. This new test (`liveTurn: true` + `wakeContext`) exercises exactly that, and the existing absent-flag case is kept. The stale `livePrincipal` comment two lines up was renamed to `liveTurn` as well.

skills/approve-action/handler.ts
```diff
@@ -97,8 +93,9 @@ export class ApproveActionHandler implements SkillHandler {
       // Step 4: Publish human.decision audit event (best-effort).
       //
       // originator is stamped by the dispatcher on every task and carries the contactId
-      // and channel of whoever initiated the task chain. Must be present if we passed the
-      // principal-origin check above — if missing, that indicates a dispatch-layer bug.
+      // and channel of whoever initiated the task chain. For this elevated skill the
+      // execution-layer live-principal gate (#1126) already guaranteed a principal originator —
```
josephfung, line 97: > Update the later audit wording to match this new gate.

Fixed in 7384bc3 — the audit comment no longer references a removed "principal-origin check above"; it now states that for this elevated skill the execution-layer live-principal gate (#1126) already guaranteed a principal originator, so a missing one indicates a dispatch-layer bug.

skills/deny-action/handler.ts
```diff
@@ -46,8 +41,9 @@ export class DenyActionHandler implements SkillHandler {
       // Publish human.decision audit event (best-effort).
       //
       // originator is stamped by the dispatcher on every task and carries the contactId
-      // and channel of whoever initiated the task chain. Must be present if we passed the
-      // principal-origin check above — if missing, that indicates a dispatch-layer bug.
+      // and channel of whoever initiated the task chain. For this elevated skill the
+      // execution-layer live-principal gate (#1126) already guaranteed a principal originator —
```
josephfung, line 45: > Bring the audit comment below along for the ride.

Done in 7384bc3 — same fix as `approve-action`: the audit block now points at the execution-layer live-principal gate (#1126) instead of the removed handler-level principal-origin check.

skills/dismiss-action/handler.ts
```diff
@@ -45,8 +40,9 @@ export class DismissActionHandler implements SkillHandler {
       // Publish human.decision audit event (best-effort).
       //
       // originator is stamped by the dispatcher on every task and carries the contactId
-      // and channel of whoever initiated the task chain. Must be present if we passed the
-      // principal-origin check above — if missing, that indicates a dispatch-layer bug.
+      // and channel of whoever initiated the task chain. For this elevated skill the
+      // execution-layer live-principal gate (#1126) already guaranteed a principal originator —
```
josephfung, line 44: > Update the stale audit comment below before it misleads the next poor sod.

Fixed in 7384bc3 — the dismiss-action audit comment now describes the execution-layer live-principal gate (#1126) as the sole enforcement point; the "principal-origin check above" wording is gone.

tests/integration/woken-task-authorization.test.ts
```diff
@@ -110,11 +120,14 @@ describeIf('woken-task authorization (#1060 dedup scenario)', () => {
...
-  it('the woken derived task is BLOCKED below posture D and ALLOWED at it (surface-and-confirm)', async () => {
+  it('contact-merge (now normal+medium) on a woken dedup task: surface-and-confirm below 70, auto at/above', async () => {
     const taskId = await seedReviewTask();
     const candidate = (await selectHeartbeatCandidates(pool, {
       eligibleAgents: ['contacts'], idleThresholdHours: 4, staleWaitThresholdHours: 48, maxWakes: 5,
-    })).find((c) => c.id === taskId)!;
+    })).find((c) => c.id === taskId);
+    // Guard the lookup explicitly (CodeAnt, #1156): a missing candidate means the seed/selection
+    // path regressed — fail with a clear message rather than a bare non-null-assertion crash.
+    if (!candidate) throw new Error(`heartbeat candidate not found for seeded task ${taskId}`);
```
josephfung, line 130: **CodeAnt point 1 of 2** (non-null assertion after `.find(...)`, originally flagged at line 127):

Valid — fixed in 7384bc3. The `.find(...)!` is replaced with an explicit guard that throws a clear, descriptive error if no candidate matches, so a regressed seed/selection path fails loudly instead of crashing on a bare non-null assertion. Applied to the sibling elevated-gate test below as well (same pattern).

tests/integration/woken-task-authorization.test.ts
```diff
@@ -123,23 +136,57 @@ describeIf('woken-task authorization (#1060 dedup scenario)', () => {
       `SELECT originator, task_payload FROM scheduled_jobs WHERE id = $1`,
       [jobId],
     );
+    expect(rows).toHaveLength(1); // guard before indexing (CodeAnt, #1156): no row → clear assertion failure
```
josephfung, line 139: **CodeAnt point 2 of 2** (unguarded `rows[0]` access, originally flagged at line 136):

Valid — fixed in 7384bc3. Added `expect(rows).toHaveLength(1)` before indexing, so a missing row fails with a clear test assertion rather than dereferencing `undefined`. (Project convention does sanction `array[0]!` where existence is guaranteed, but in a test an explicit assertion gives a far better failure message, so the guard is the right call here.) Same fix applied to the sibling test below.

Issue #1051, opened 2026-06-18
Return actionable block reason from outbound content filter so agents can self-correct and retry


josephfung:
## Summary

When the outbound content filter blocks a message, `outbound-gateway.ts` returns a generic string that gives the agent's LLM no signal about *why* the block happened:

```typescript
// src/skills/outbound-gateway.ts L773
return { success: false, blockedReason: 'Content blocked by filter' };
```

The agent's tool-use loop sees a failed skill call with no correction path, so it cannot self-correct and retry naturally.

The fix is minimal. `reasonSummary` is already computed at L692 via `buildBlockReasonSummary()` — it's a principal-safe abstract reason (strips sensitive Stage-1 detail) that is currently used only for the CEO notification. Surfacing it, along with the rule name(s), in the skill return value gives the LLM enough context to rewrite the message and retry via its tool loop. No retry counter or retry machinery needed — the agent's existing tool-use loop handles retries organically, with the conversation budget and task timeout as the natural circuit breakers.

## Context

Caught in production: an agent's internal fallback reasoning leaked into an outbound email body. The LLM judge correctly blocked it (`llm-judge-audience-leak`). The agent received no feedback to self-correct — only the CEO was notified. The correct content (a meeting confirmation) was never sent.

### `src/skills/outbound-gateway.ts` — L773

```typescript
// Before
return { success: false, blockedReason: 'Content blocked by filter' };

// After — reasonSummary already computed at L692; zero extra work
return {
  success: false,
  blockedReason: reasonSummary,
  blockedRules: filterFindings.map((f) => f.rule),
};
```

`blockedRules` gives the LLM the specific rule category (e.g. `llm-judge-audience-leak`) so it can address the root cause, not just retry blindly.

### CEO notification subject line (same PR, optional)

Current: `'Action needed — blocked outbound reply'`

Since the agent may now self-correct and successfully resend, consider softening to `'FYI — outbound message blocked'` with a note that the agent may retry. The audit log is always the ground truth; the notification becomes informational rather than a call to action.

## What does NOT need to change

- No retry counter or retry state machinery — the agent's tool loop handles it
- CEO notification still fires on every block — audit trail integrity unchanged
- All other evaluate→block→notify patterns use intentional human-in-the-loop approval (autonomy gates, escalation judge) and are correct as-is

## Acceptance criteria

- [ ] `outbound-gateway.ts` returns `reasonSummary` and rule name(s) in the blocked skill result
[...]

PR #1161, opened 2026-06-24
fix: return actionable block reason from sendEmailDraft() content filter (#1158)

### LINKED ISSUE #1158: Return actionable block reason from sendEmailDraft() content filter so agents can self-correct
## Summary

Follow-up to #1051 (PR #1157). That fix made the **`send()`** path return the principal-safe `reasonSummary` plus the triggering rule name(s) in the blocked skill result, so the agent's tool loop can self-correct and retry. The parallel **`sendEmailDraft()`** block path was intentionally left out of scope and still returns the generic string:

```typescript
// src/skills/outbound-gateway.ts — sendEmailDraft(), second `if (!filterPassed)` block (~L1462 on main)
return { success: false, blockedReason: 'Content blocked by filter' };
```

So a draft send blocked by the content filter (e.g. via the `send-draft` skill) gives the agent no signal to fix the content, unlike a regular `send()`. This is the same defect #1051 fixed, on a lower-traffic path.

## Context

In the draft block path the gateway already computes `ruleNames` (rule names joined) and `fullReason` (rule + detail, for the audit event) but **does not** compute `reasonSummary` via `buildBlockReasonSummary()`, and surfaces neither to the caller. The CEO notification there is also still a call-to-action (`'Action needed — blocked draft send'`).

## What to change

Mirror the `send()` fix in `sendEmailDraft()`:

1. Compute `const reasonSummary = buildBlockReasonSummary(filterFindings);` in the draft block path (same helper, same per-rule safety contract).
2. Return `{ success: false, blockedReason: reasonSummary, blockedRules: filterFindings.map((f) => f.rule) }` instead of the generic string.
3. (Optional, matching #1051) Soften the draft block notification subject from `'Action needed — blocked draft send'` to an FYI and add a note that the agent may retry.

`OutboundSendResult.blockedRules` already exists (added in #1157), so no interface change is needed.

## Safety

`blockedReason` must reuse `reasonSummary` so it inherits `buildBlockReasonSummary()`'s contract: only `llm-judge-*` / `disclosure-tier-gate` findings surface their abstract detail; every other (Stage-1 deterministic) rule is rule-name-only. `blockedRules` carries rule names only. This is the same invariant #1051 verified — do not surface `fullReason` (which embeds matched fragments) to the caller.

## Acceptance criteria

- [ ] `sendEmailDraft()` returns `reasonSummary` and rule name(s) in the blocked skill result (not the generic `'Content blocked by filter'`)
- [ ] The draft-path LLM-facing payload is verified against `buildBlockReasonSummary()`'s safety contract — no Stage-1 finding detail (matched fragment) leaks into the result
- [ ] CEO notification still fires on every draft block (no behavior regression)
- [ ] Unit tests cover the updated draft-path return shape for both LLM-judge blocks and Stage-1 deterministic blocks (mirror the `send()` tests added in #1157)
- [ ] `send-draft` skill handler surfaces the new reason to the agent (verify it forwards `result.blockedReason`)

josephfung:
## Summary

Follow-up to #1051 (PR #1157). That fix made the **`send()`** content-filter block path return the principal-safe `reasonSummary` plus the triggering rule name(s) so an agent can self-correct and retry. This applies the same fix to the parallel **`sendEmailDraft()`** block path, which still returned the opaque `'Content blocked by filter'`.

A draft send blocked by the content filter (e.g. via the `send-draft` skill) now gives the agent the same actionable signal as a regular send.

Closes #1158

## What changed

In the `sendEmailDraft()` block path (`src/skills/outbound-gateway.ts`):

1. Compute `reasonSummary = buildBlockReasonSummary(filterFindings)` (the same helper and per-rule safety contract as `send()`).
2. Return `{ success: false, blockedReason: reasonSummary, blockedRules: filterFindings.map(f => f.rule) }` instead of the generic string.
3. Soften the CEO block notification from a call-to-action (`'Action needed — blocked draft send'`) to an FYI, add the reason, and bring the body to parity with the `send()` notification (Time / Channel / Audit event ID, with the search hint pointing at the audit event ID).
4. Update the `OutboundSendResult.blockedRules` JSDoc — it no longer notes the draft path as an outstanding gap.

No interface change (`blockedRules` was added in #1157). The `send-draft` skill handler already forwards `result.blockedReason` into the skill `error` field, so the reason reaches the agent with no handler change.

## Safety

`blockedReason` reuses `reasonSummary`, inheriting `buildBlockReasonSummary()`'s contract: only `llm-judge-*` / `disclosure-tier-gate` findings surface their abstract detail; every Stage-1 deterministic rule is rule-name-only, so no matched secret fragment leaks into the LLM-facing result. `fullReason` (rule + detail) stays confined to the audit bus event. `blockedRules` carries rule names only. Verified by the negative-leak and mixed-finding tests.

## Tests

Mirrors the `send()` tests from #1157 on the draft path:
- Stage-1 deterministic block → `blockedReason` is the rule name only, secret never present
- LLM-judge block → judge's abstract reason + rule name surface
- Negative-leak → a Stage-1 secret fragment never appears in the result
- Mixed findings → only the judge detail surfaces; `blockedRules` lists all rules in order

94/94 in `outbound-gateway.test.ts` pass; 30/30 in `send-draft/handler.test.ts`; typecheck clean.

## Acceptance criteria

- [x] `sendEmailDraft()` returns `reasonSummary` and rule name(s), not the generic string
- [x] Draft-path payload verified against `buildBlockReasonSummary()`'s contract — no Stage-1 detail leaks
- [x] CEO notification still fires on every draft block (no regression)
- [x] Unit tests cover the updated draft-path return shape for LLM-judge and Stage-1 blocks
- [x] `send-draft` handler surfaces the new reason (already forwards `result.blockedReason`)


josephfung:
@coderabbitai review

josephfung:
@coderabbitai review

PR #1180, opened 2026-06-25
fix: thread TaskOriginator onto wake_at self-deferral wake jobs (#1153)

### LINKED ISSUE #1153: Thread TaskOriginator onto task wake_at self-deferral wake jobs
> **Part of #1060 — follow-up to #1125.** Independent of #1126 and #1127.

## Summary
A task that defers itself via its own `wake_at` column mints a one-shot wake job with no `originator` and no `standing` envelope (`src/db/task-repo.ts`, see the `@TODO(#1125/#1127)`), so `fireJob` fires it with `metadata: undefined` and it floors to agent / no-bypass standing.

## Why this is wrong
A `wake_at` time is *pre-chosen*, which under the design note (§3) makes it pre-authorized — the same category as a `scheduler-create` job, which **keeps** its originator at fire time. It must **not** be subject to the heartbeat bypass ladder (the time was already decided). As shipped, a principal-lineage task that self-defers loses its principal-bypass entirely on wake. Fail-closed (safe), but a real asymmetry: heartbeat wakes are laddered, scheduler-create keeps principal, wake_at floors. For agent-lineage tasks it's a no-op (floor is agent anyway); the blast radius is principal-lineage self-deferrals.

## Scope
- Persist the task's `originator` onto the `wake_at` wake job's `scheduled_jobs` row (the CTE in `createTask`).
- Do **not** write a `standing` envelope — these must keep the originator at fire time, not ladder.
- `fireJob` already threads `job.originator` when no envelope is present, so the fire side likely needs no change — verify.

## Acceptance criteria
- [ ] A `wake_at` wake job carries the task's `originator` on its `scheduled_jobs` row.
- [ ] On fire, a principal-lineage `wake_at` task is treated as principal-originated for the **autonomy** principal-bypass (`normal` skills), with no ladder applied; `elevated` skills remain blocked (the fire is not a live turn — see Update).
- [ ] The `@TODO(#1125/#1127)` in `task-repo.ts` is removed.
- [ ] Test covering a principal-lineage task self-deferring via `wake_at` and waking with the autonomy bypass intact.

## Update (2026-06-24 — #1126 shipped)

#1126 redefined `elevated` as a **live principal turn** (a distinct off-bag `liveTurn` field; the gate also rejects any task carrying a `wakeContext` marker outright). This **sharpens but does not change** this issue's scope:

- "Bypass intact" here means the **autonomy** principal-bypass (for `normal` skills), **not** the `elevated` gate. A `wake_at` fire carries no `liveTurn`, so `elevated` authority primitives (approve/deny/dismiss-action, set-autonomy, the contact-auth skills, secret-capture) stay **blocked** on a self-deferral wake. That is correct and intended: a pre-chosen deferral may resume CEO-authorized *work*, but must not *exercise authority* without the CEO present.
- The fix stays "thread the **originator** only, no `standing` envelope." Do **not** thread `liveTurn` onto the wake job. Carrying no `standing` envelope also means no `wakeContext` marker, so the `wake_at` fire is not caught by the gate's defence-in-depth `wakeContext` rejection — consistent with treating it like a `scheduler-create` fire for the autonomy bypass.

See design note §3b ("Interaction with #1126").

josephfung:
## Summary

Closes #1153. Follow-up to #1125 (depends on it); independent of #1126/#1127.

A task that defers itself via its own `wake_at` column minted a one-shot wake job with **no** `originator` and no `standing` envelope, so `fireJob` fired it with `metadata: undefined` and it floored to **agent / no-bypass**. A `wake_at` time is *pre-chosen*, which under the design note (§3) makes it **pre-authorized** — the same category as a `scheduler-create` job, which keeps its originator at fire time. As shipped, a principal-lineage task that self-deferred silently lost its principal-bypass on wake (fail-closed, but a real asymmetry: heartbeat wakes ladder, scheduler-create keeps principal, `wake_at` floored).

## What changed

Persist the task's `originator` onto the `wake_at` wake job's `scheduled_jobs` row at **both** mint sites in `src/db/task-repo.ts`, while writing **no** `standing` envelope:

- **`createTask` CTE** (`_wake_job`) — now `SELECT … new_task.originator` (definitionally the same parent-capped lineage stamped on the task).
- **`updateTask` `_new_wake` arm** — the literal self-deferral path (a running task rescheduling itself via `task-update`); now binds `current.originator` (the immutable lineage from the pre-update row) as a `::jsonb` param.

The issue scoped only the `createTask` CTE, but the same drop exists in `updateTask`'s `_new_wake` arm — which is the more literal "a task defers itself" path — and AC #1 reads generically ("a `wake_at` wake job"). Fixing only one would leave the headline scenario broken, so both are addressed.

`fireJob` keys the `wakeContext`/bypass-ladder marker on the `standing` field being present, **not** on the originator. So: originator present + no standing → `fireJob` threads `job.originator` but mints no `wakeContext` → the task **keeps** its principal autonomy-bypass and the heartbeat ladder never runs. The fire side needed no change (verified). `elevated` authority primitives stay blocked (a wake carries no `liveTurn` — see §3b).

No migration needed: `scheduled_jobs.originator` already exists (migration 040).

## Acceptance criteria

- [x] A `wake_at` wake job carries the task's `originator` on its `scheduled_jobs` row.
- [x] On fire, a principal-lineage `wake_at` task is treated as principal-originated for the **autonomy** principal-bypass (`normal` skills), with no ladder applied; `elevated` skills remain blocked.
- [x] The `@TODO(#1125/#1127)` in `task-repo.ts` is removed.
- [x] Test covering a principal-lineage task self-deferring via `wake_at` and waking with the autonomy bypass intact.

## Tests

All against real Postgres (`curia-test-pg`); written first (red), then green.

- `task-repo-originator.test.ts` — new `#1153` block: `createTask` + `updateTask` `wake_at` jobs carry the originator with a bare `{type:'task-wake'}` payload (no `standing`); a null-lineage task writes a null originator (no regression).
- `woken-task-authorization.test.ts` — end-to-end fire-path: a principal-lineage task self-defers via `wake_at`, fed through the real `ExecutionLayer`; the principal-bypass fires for a `normal`+`critical` skill at score 75 (no ladder), while an `elevated` skill stays blocked at any score. The `metadataFromWakeRow` helper was corrected to mirror `fireJob` exactly (keys `wakeContext` on `standing` presence, not the payload type).

Full run: 12/12 new + 184 scheduler/heartbeat regression tests pass; `typecheck` + `eslint` clean.

## Docs

- CHANGELOG `## [Unreleased]` → Fixed.
- Design note `docs/wip/2026-06-22-woken-task-authorization-design.md` §3b marked resolved; the §3a table row and status list updated.

### DIFF
--- a/tests/integration/task-repo-originator.test.ts
+++ b/tests/integration/task-repo-originator.test.ts
@@ -25,6 +25,13 @@ const SYSTEM: TaskOriginator = {
 };
 
 async function cleanup(pool: pg.Pool): Promise<void> {
+  // scheduled_jobs.task_id is ON DELETE SET NULL, so the wake-job rows minted by the wake_at
+  // tests must be deleted FIRST (while the task link still exists) — otherwise they'd survive
+  // the task delete as task_id=NULL orphans, polluting the shared test DB (#1153).
+  await pool.query(
+    `DELETE FROM scheduled_jobs WHERE task_id IN (SELECT id FROM tasks WHERE title LIKE $1)`,
+    [`${PREFIX}%`],
+  );
   // Children reference parents via parent_task_id — delete all test rows in one statement.
   await pool.query(`DELETE FROM tasks WHERE title LIKE $1`, [`${PREFIX}%`]);
 }
@@ -70,5 +77,58 @@ describeIf('TaskRepo originator lineage (#1125)', () => {
       parentTaskId: parent.id, originator: PRINCIPAL,
     });
     expect(child.originator).toBeNull();
+  });
+
+  // #1153: a wake_at deferral is pre-chosen, so its wake job must carry the task's originator
+  // (like a scheduler-create job) and write NO `standing` envelope — fireJob then threads the
+  // originator with no wakeContext, keeping the principal autonomy-bypass without the heartbeat ladder.
+  describe('wake_at deferral threads originator onto the wake job (#1153)', () => {
+    /** Fetch every wake job linked to a task (newest first), for asserting originator + payload. */
+    async function wakeJobsFor(taskId: string) {
+      const { rows } = await pool.query<{
+        originator: TaskOriginator | null;
+        task_payload: Record<string, unknown>;
+        status: string;
+      }>(
+        `SELECT originator, task_payload, status FROM scheduled_jobs WHERE task_id = $1 ORDER BY created_at DESC`,
+        [taskId],
+      );
+      return rows;
+    }
+
+    it('createTask with wake_at persists the task originator (no standing envelope)', async () => {
+      const task = await repo.createTask({
+        agentId: 'coordinator', title: `${PREFIX} wake-create`, source: 'coordinator',
+        originator: PRINCIPAL, wakeAt: new Date(Date.now() + 3_600_000),
+      });
+      const jobs = await wakeJobsFor(task.id);
+      expect(jobs).toHaveLength(1);
+      expect(jobs[0]!.originator).toEqual(PRINCIPAL);
+      // No standing envelope: payload is the bare task-wake marker. This is what keeps the wake
+      // out of the bypass ladder at fire time (the time was already decided).
+      expect(jobs[0]!.task_payload).toEqual({ type: 'task-wake' });
+    });
+
+    it('updateTask with wake_at persists the existing task originator onto the new wake job', async () => {
+      const task = await repo.createTask({
+        agentId: 'coordinator', title: `${PREFIX} wake-update`, source: 'coordinator', originator: PRINCIPAL,
+      });
+      const updated = await repo.updateTask(task.id, { wakeAt: new Date(Date.now() + 3_600_000) }, 'coordinator');
+      expect(updated).not.toBeNull();
+      const pending = (await wakeJobsFor(task.id)).filter((j) => j.status === 'pending');
+      expect(pending).toHaveLength(1);
+      expect(pending[0]!.originator).toEqual(PRINCIPAL);
+      expect(pending[0]!.task_payload).toEqual({ type: 'task-wake' });
+    });
+
+    it('a null-lineage task mints a wake job with null originator (no regression)', async () => {
+      const task = await repo.createTask({
+        agentId: 'coordinator', title: `${PREFIX} wake-nolineage`, source: 'coordinator',
+        wakeAt: new Date(Date.now() + 3_600_000),
+      });
+      const jobs = await wakeJobsFor(task.id);
+      expect(jobs).toHaveLength(1);
+      expect(jobs[0]!.originator).toBeNull();
+    });
   });
 });
--- a/tests/integration/woken-task-authorization.test.ts
+++ b/tests/integration/woken-task-authorization.test.ts
@@ -18,6 +18,7 @@ import pg from 'pg';
 import pino from 'pino';
 import { selectHeartbeatCandidates } from '../../src/db/queries/tasks.js';
 import { SchedulerService } from '../../src/scheduler/scheduler-service.js';
+import { TaskRepo } from '../../src/db/task-repo.js';
 import { makePrincipalOriginator } from '../../src/contacts/principal.js';
 import { makeWakeContext } from '../../src/autonomy/effective-standing.js';
 import { ExecutionLayer } from '../../src/skills/execution.js';
@@ -55,14 +56,19 @@ function makeManifest(
   };
 }
 
-/** Rebuild the agent.task metadata exactly as Scheduler.fireJob does from a persisted wake row. */
+/** Rebuild the agent.task metadata exactly as Scheduler.fireJob does from a persisted wake row.
+ *  Mirrors fireJob precisely: the wakeContext marker is keyed on the `standing` envelope being
+ *  PRESENT, not merely on the task-wake payload type. A wake_at self-defer job (#1153) carries a
+ *  task-wake payload with NO standing, so it threads the originator but mints no wakeContext — i.e.
+ *  it keeps its originator like a scheduler-create fire and is not subject to the bypass ladder. */
 function metadataFromWakeRow(row: { originator: Record<string, unknown> | null; task_payload: Record<string, unknown> }) {
-  if (!row.originator) return undefined;
-  const meta: Record<string, unknown> = { originator: row.originator };
-  if (row.task_payload?.['type'] === 'task-wake') {
-    const standing = (row.task_payload as { standing?: { derived?: boolean } }).standing;
-    meta.wakeContext = makeWakeContext(standing?.derived === true);
-  }
+  const standing = row.task_payload?.['type'] === 'task-wake'
+    ? (row.task_payload as { standing?: { derived?: boolean } }).standing
+    : undefined;
+  if (!row.originator && !standing) return undefined;
+  const meta: Record<string, unknown> = {};
+  if (row.originator) meta.originator = row.originator;
+  if (standing) meta.wakeContext = makeWakeContext(standing.derived === true);
   return meta;
 }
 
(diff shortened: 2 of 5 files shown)


josephfung:
@Coderabbitai review

josephfung:
@coderabbitai review

tests/integration/woken-task-authorization.test.ts
```diff
@@ -55,14 +56,19 @@ function makeManifest(
...
 function metadataFromWakeRow(row: { originator: Record<string, unknown> | null; task_payload: Record<string, unknown> }) {
-  if (!row.originator) return undefined;
-  const meta: Record<string, unknown> = { originator: row.originator };
-  if (row.task_payload?.['type'] === 'task-wake') {
-    const standing = (row.task_payload as { standing?: { derived?: boolean } }).standing;
-    meta.wakeContext = makeWakeContext(standing?.derived === true);
-  }
+  const standing = row.task_payload?.['type'] === 'task-wake'
+    ? (row.task_payload as { standing?: { derived?: boolean } }).standing
+    : undefined;
```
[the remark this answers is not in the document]
josephfung: Good catch — fixed in b6dac380. Narrowed through `unknown` first per the repo's `Record<string, unknown>` rule:

```ts
? (row.task_payload as unknown as { standing?: { derived?: boolean } }).standing
```

(For the record this is only a style-consistency nit here: the target type has all-optional properties, so the direct cast already compiled — and the production source it mirrors, `fireJob` in `src/scheduler/scheduler.ts`, uses the same direct cast. Aligned the test to the guideline anyway since it's zero-risk.) typecheck + eslint clean, 12/12 tests still green.

tests/integration/task-repo-originator.test.ts
```diff
@@ -25,6 +25,13 @@ const SYSTEM: TaskOriginator = {
 };
 
 async function cleanup(pool: pg.Pool): Promise<void> {
+  // scheduled_jobs.task_id is ON DELETE SET NULL, so the wake-job rows minted by the wake_at
+  // tests must be deleted FIRST (while the task link still exists) — otherwise they'd survive
+  // the task delete as task_id=NULL orphans, polluting the shared test DB (#1153).
+  await pool.query(
+    `DELETE FROM scheduled_jobs WHERE task_id IN (SELECT id FROM tasks WHERE title LIKE $1)`,
+    [`${PREFIX}%`],
+  );
```
josephfung, line 34: **CodeAnt finding 1/6 — declining (false positive on test code).**

This is the `cleanup()` DELETE that runs in `beforeEach`/`afterAll`. In Vitest an unguarded awaited rejection *is* the failure path: it propagates to the runner and fails the test with the full stack, which is exactly what we want for a setup/teardown query. A `try/catch` here would either swallow that signal or just re-throw (a no-op).

It would also diverge from every other integration test in the repo — including the pre-existing helpers in this same file (the original `cleanup`, the seed helpers in `woken-task-authorization.test.ts`), which all use unguarded awaits. The repo error-handling rule (CLAUDE.md: no empty catch, every catch logs/audits/propagates) governs production code, not tests. Leaving as-is.

(This `custom_rule` is misfiring on test files generally — worth scoping it to exclude `tests/**` in `.codeant/` so the other five don't recur.)

tests/integration/task-repo-originator.test.ts
```diff
@@ -71,4 +78,57 @@ describeIf('TaskRepo originator lineage (#1125)', () => {
...
+      const { rows } = await pool.query<{
+        originator: TaskOriginator | null;
+        task_payload: Record<string, unknown>;
+        status: string;
+      }>(
+        `SELECT originator, task_payload, status FROM scheduled_jobs WHERE task_id = $1 ORDER BY created_at DESC`,
+        [taskId],
+      );
+      return rows;
+    }
```
josephfung, line 97: **CodeAnt finding 2/6 — declining (false positive on test code).**

This is the `wakeJobsFor()` read helper. Same reasoning as finding 1/6: an unguarded awaited `pool.query` in a test surfaces a rejection as a test failure with full context — the desired behavior. Wrapping it in `try/catch` would either swallow that or re-throw with no added value, and diverges from the unguarded-await convention used throughout the repo's integration tests. No production code path is involved. Leaving as-is.

tests/integration/task-repo-originator.test.ts
```diff
@@ -71,4 +78,57 @@ describeIf('TaskRepo originator lineage (#1125)', () => {
...
+        agentId: 'coordinator', title: `${PREFIX} wake-create`, source: 'coordinator',
+        originator: PRINCIPAL, wakeAt: new Date(Date.now() + 3_600_000),
+      });
+      const jobs = await wakeJobsFor(task.id);
+      expect(jobs).toHaveLength(1);
+      expect(jobs[0]!.originator).toEqual(PRINCIPAL);
+      // No standing envelope: payload is the bare task-wake marker. This is what keeps the wake
+      // out of the bypass ladder at fire time (the time was already decided).
+      expect(jobs[0]!.task_payload).toEqual({ type: 'task-wake' });
+    });
```
josephfung, line 110: **CodeAnt finding 3/6 — declining (false positive on test code).**

This is a test-case body (the `createTask` wake_at assertion). The awaited calls (`repo.createTask`, `wakeJobsFor`) are meant to fail the test if they reject — that is the assertion mechanism Vitest provides. A `try/catch` would suppress or merely re-emit the failure. This matches the repo's established test style (no integration test guards its awaits) and the CLAUDE.md error-handling rule applies to production code, not tests. Leaving as-is.

tests/integration/task-repo-originator.test.ts
```diff
@@ -71,4 +78,57 @@ describeIf('TaskRepo originator lineage (#1125)', () => {
...
+      const task = await repo.createTask({
+        agentId: 'coordinator', title: `${PREFIX} wake-update`, source: 'coordinator', originator: PRINCIPAL,
+      });
+      const updated = await repo.updateTask(task.id, { wakeAt: new Date(Date.now() + 3_600_000) }, 'coordinator');
+      expect(updated).not.toBeNull();
+      const pending = (await wakeJobsFor(task.id)).filter((j) => j.status === 'pending');
+      expect(pending).toHaveLength(1);
+      expect(pending[0]!.originator).toEqual(PRINCIPAL);
+      expect(pending[0]!.task_payload).toEqual({ type: 'task-wake' });
+    });
```
josephfung, line 122: **CodeAnt finding 4/6 — declining (false positive on test code).**

This is a test-case body (the `updateTask` wake_at assertion). Same as 3/6: unguarded awaits are the intended failure path in a Vitest test; wrapping them in `try/catch` swallows or re-throws the very signal the test exists to produce. Consistent with every other integration test in the repo; the error-handling rule is for production code. Leaving as-is.

tests/integration/task-repo-originator.test.ts
```diff
@@ -71,4 +78,57 @@ describeIf('TaskRepo originator lineage (#1125)', () => {
...
+
+    it('a null-lineage task mints a wake job with null originator (no regression)', async () => {
+      const task = await repo.createTask({
+        agentId: 'coordinator', title: `${PREFIX} wake-nolineage`, source: 'coordinator',
+        wakeAt: new Date(Date.now() + 3_600_000),
+      });
+      const jobs = await wakeJobsFor(task.id);
+      expect(jobs).toHaveLength(1);
+      expect(jobs[0]!.originator).toBeNull();
+    });
```
josephfung, line 132: **CodeAnt finding 5/6 — declining (false positive on test code).**

This is a test-case body (the null-lineage regression check). Same reasoning as 3/6 and 4/6: the awaited `createTask` / `wakeJobsFor` calls should fail the test on rejection, which Vitest already does for an unguarded await. A `try/catch` adds nothing and would diverge from repo test convention. Leaving as-is.

tests/integration/woken-task-authorization.test.ts
```diff
@@ -257,4 +264,52 @@ describeIf('woken-task authorization (#1060 dedup scenario)', () => {
...
+    const repo = new TaskRepo(
+      pool,
+      { publish: async () => {}, subscribe: () => {} } as unknown as EventBus,
+      logger as never,
+      'UTC',
+    );
+    const task = await repo.createTask({
+      agentId: 'coordinator', title: `${PREFIX} wake_at self-defer`, source: 'ceo',
+      originator: CONSOLE_PRINCIPAL_LINEAGE, wakeAt: new Date(Date.now() + 3_600_000),
+    });
```
josephfung, line 281: **CodeAnt finding 6/6 — declining (false positive on test code).**

This is the awaited `repo.createTask` (and the following `pool.query` SELECT) in the new fire-path test. As with the other five: in a Vitest test an unguarded awaited rejection is the failure path — it propagates with the full stack and fails the test, which is the behavior we want. Wrapping in `try/catch` would swallow or re-throw it, and would diverge from the unguarded-await style used across the repo's integration suite. The CLAUDE.md error-handling rule targets production code, not tests. Leaving as-is.

PR #1182, opened 2026-06-25
fix: render markdown links in outbound HTML

### LINKED ISSUE #1169: Email markdown links render literally; consolidate the two partial markdown→HTML converters into one canonical sanitized converter
## Summary

When Curia drafts/sends an email, markdown link syntax like `[Michaël Bijaoui](https://www.linkedin.com/in/michaelbijaoui)` is delivered to the recipient **verbatim** instead of as a clickable link. Root cause: the email markdown→HTML converter is a hand-rolled, deliberately-partial converter that never implemented link syntax (and doesn't autolink bare URLs either). This issue fixes the bug *and* removes the underlying fragility by replacing both of Curia's divergent hand-rolled converters with a single canonical converter built on `markdown-it`, with sanitized output.

## Root cause

Outbound email flows: LLM writes markdown → `markdownToHtml()` → Nylas receives an HTML `body`. The converter at [`src/channels/email/markdown-to-html.ts`](https://github.com/josephfung/curia/blob/main/src/channels/email/markdown-to-html.ts) only handles inline code, bold, italic (see `applyInline`, lines 73–92). It has **no rule for `[text](url)`** and **no bare-URL autolinking**, so link markdown matches nothing and passes through as literal text.

Compounding this, there is a **second, divergent converter** at [`src/utils/markdown-to-html.ts`](https://github.com/josephfung/curia/blob/main/src/utils/markdown-to-html.ts) used by the web-console/SSE rendering path. It *does* autolink bare URLs but *also* lacks `[text](url)` support. Two converters with different behavior = the "rats nest" this issue untangles.

Conversion logic (`looksLikeHtml()` guard + `markdownToHtml()` call) is additionally **copy-pasted across four email call sites**: `outbound-gateway.ts` (send + draft) and the three `ceo-inbox-draft-*` skills.

## Decisions (already agreed)

1. **Fix as rich HTML, not plaintext.** Curia already sends every email as HTML and depends on it for bold/lists/paragraphs. Switching to Nylas `isPlaintext: true` to get Gmail's auto-linking would regress all existing formatting. So: keep sending HTML, emit a proper `<a href>` for links. The name stays the clickable label.
2. **The Nylas adapter stays a pure transport.** [`src/channels/email/nylas-client.ts`](https://github.com/josephfung/curia/blob/main/src/channels/email/nylas-client.ts) continues to take an HTML `body` and is **not** modified. Conversion is a content concern that happens upstream, not in the adapter. We do **not** set `isPlaintext` — Nylas auto-generates the plaintext alternative from the HTML.
3. **Use `markdown-it`, not an extended hand-roll.** A real CommonMark implementation eliminates the entire class of "markdown feature X was never implemented" bugs and correctly handles the edge cases the hand-rolled code special-cased (e.g. intra-word `some_var_name` is not emphasis under CommonMark flanking rules).
4. **Consolidate both converters into one module.** Email and chat/SSE rendering share one converter so they can never drift again.

### New canonical module

Create `src/format/markdown-to-html.ts` (final location TBD by reviewer) exposing one function, e.g.:

```ts
export function markdownToHtml(input: string, opts?: { wrap?: boolean }): string
```

Pipeline:

1. **HTML passthrough guard.** If `looksLikeHtml(input)` (LLM wrote HTML directly), skip markdown conversion and use the input as-is. (Keep/relocate the existing `looksLikeHtml` helper.)
2. **Convert** with `markdown-it` configured: `html: false` (don't pass raw HTML through the markdown parser — the guard above handles intentional HTML), `linkify: true` (autolink bare URLs), `breaks: true` (single newline → `<br>`, matching today's email behavior).
3. **Sanitize the output** with `sanitize-html` (already a dependency). Whitelist the inline/block tags we emit (`a, p, strong, em, code, pre, ul, ol, li, br, blockquote, hr, h1–h6`), allow `a[href]` only for `http`/`https`/`mailto`, and strip `javascript:`/`data:` URLs. Reuse/derive from the existing `QUOTE_SANITIZE_OPTIONS` in [`src/skills/_shared/reply-quote.ts`](https://github.com/josephfung/curia/blob/main/src/skills/_shared/reply-quote.ts) so the project has one sanitize policy. **Sanitizing the output is the standardizing move** — it makes link injection safe regardless of how the HTML was produced.
4. **Wrap** (when `opts.wrap`): the styled `<div style="font-family: Arial…">` shell currently in the email converter. Email path passes `wrap: true`; chat/SSE passes `wrap: false`.

### Wire up call sites

- **Email — `src/skills/outbound-gateway.ts`** (lines ~1834 send-draft, ~1888 send): call the shared converter with `wrap: true`. Preserve the existing contract that `htmlQuote` is appended **after** conversion (it is pre-sanitized HTML built by `reply-quote.ts` and must not be re-escaped).
- **Email — the three CEO-inbox skills** (`skills/ceo-inbox-draft-reply/handler.ts`, `ceo-inbox-draft-compose/handler.ts`, `ceo-inbox-draft-edit/handler.ts`): replace their local `looksLikeHtml` + `markdownToHtml` dance with a single call to the shared module.
- **Chat/SSE — `src/channels/http/event-router.ts`** (~line 171) and **`src/channels/http/routes/kg.ts`** (~line 1347): call the shared converter with `wrap: false`. These now get output sanitization for free (a small XSS hardening win for the web console).
- Delete the now-dead logic in `src/channels/email/markdown-to-html.ts` and `src/utils/markdown-to-html.ts` (or have them re-export the shared module if import churn is a concern, but prefer deletion + updated imports).

### Dependency

Add `markdown-it` (+ `@types/markdown-it`) via `pnpm -C <worktree> add`. Normal dependency — does **not** go in the `pnpm-workspace.yaml` overrides block. Commit `package.json` + `pnpm-lock.yaml` together.

## Testing (TDD)

Write failing tests first, then implement. Cover:

- `[Name](https://example.com)` → `<a href="https://example.com">Name</a>`
- bare URL autolinks (email **and** chat paths — closes today's behavioral gap)
- `**bold**`, `*italic*`, unordered lists, multi-paragraph, soft-wrap `<br>` — no regression
- `some_var_name` is **not** italicized
- `[x](javascript:alert(1))` is neutralized in output (sanitize)
- `looksLikeHtml` passthrough: an LLM-authored HTML body is not double-escaped
- reply-quote `htmlQuote` still appended correctly and not re-escaped

Run `pnpm -C <worktree> run typecheck` and the full vitest suite green before PR.

## Acceptance criteria

- [ ] An email whose body contains `[Name](url)` is delivered with a clickable anchor; the name is the link text.
- [ ] Bare URLs autolink in both outbound email and web-console rendering.
- [ ] Existing formatting (bold/italic/lists/paragraphs/`some_var_name`) is unchanged.
- [ ] `javascript:`/`data:` URLs in links are stripped from rendered output.
- [ ] LLM-authored direct-HTML bodies still pass through without double-escaping; reply quotes render correctly.
- [ ] Exactly one markdown→HTML converter module exists; `src/channels/email/markdown-to-html.ts` and `src/utils/markdown-to-html.ts` no longer contain independent conversion logic.
- [ ] All six call sites (4 email + 2 chat/SSE) use the shared module; `nylas-client.ts` is unchanged.
- [ ] `markdown-it` added; lockfile synced; typecheck + tests green.
- [ ] `CHANGELOG.md` updated under `[Unreleased]`; affected skill/agent `version` fields bumped per repo policy.

## Out of scope

- Inbound `html-to-text` conversion ([`src/channels/email/html-to-text.ts`](https://github.com/josephfung/curia/blob/main/src/channels/email/html-to-text.ts)) — unchanged.
- Signal/CLI plaintext formatting — unchanged.
- Any `isPlaintext` / multipart MIME work on the Nylas adapter.

josephfung:
## Summary

Fixes outbound markdown links rendering literally by replacing Curia's two partial markdown-to-HTML converters with one shared sanitized formatter. The shared formatter now also preserves LLM-authored markdown and HTML tables.

## Changes

- Added a canonical `src/format/markdown-to-html.ts` formatter using `markdown-it`, bare URL linkification, direct-HTML passthrough, and `sanitize-html` protocol filtering.
- Updated email send/draft paths and web console/SSE rendering to use the shared formatter, with email bodies using the existing styled wrapper.
- Preserved safe table markup and table alignment/structure attributes for markdown tables and direct HTML table bodies.
- Removed the old divergent email and utility converter implementations.
- Added regression coverage for markdown links, bare URL autolinking, tables, sanitizer behavior, direct HTML passthrough, CEO inbox draft payloads, and outbound gateway sends.

## Related Issues

Closes #1169

## Checklist

- [x] CI passes

Verification:
- `pnpm exec vitest run src/format/markdown-to-html.test.ts`
- `pnpm exec vitest run src/format/markdown-to-html.test.ts skills/ceo-inbox-draft-compose/handler.test.ts skills/ceo-inbox-draft-reply/handler.test.ts skills/ceo-inbox-draft-edit/handler.test.ts tests/unit/skills/outbound-gateway.test.ts`
- `pnpm run typecheck`
- `pnpm test`

[Slack Thread](https://thefungs.slack.com/archives/C0BDTUW34QY/p1782386226180449?thread_ts=1782386226.180449&cid=C0BDTUW34QY)

<div><a href="https://cursor.com/agents/bc-2348e0fc-1afd-5888-b026-949f80691fb0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2348e0fc-1afd-5888-b026-949f80691fb0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


src/format/markdown-to-html.ts
```diff
@@ -0,0 +1,82 @@
...
+  const token = tokens[idx]!;
+  const href = token.attrGet('href') ?? '';
+  if (/^(https?:)?\/\//i.test(href)) {
+    token.attrSet('target', '_blank');
+    token.attrSet('rel', 'noopener noreferrer');
+  }
+  return defaultLinkOpen(tokens, idx, options, env, self);
+};
+
+const MARKDOWN_SANITIZE_OPTIONS: sanitizeHtml.IOptions = {
```
josephfung, line 31: While we are in here upgrading the formatting processing, we should add Support for HTML tables. That is also a common output from the LLM's, in our work.

PR #1194, opened 2026-06-26
fix: warn when skill directory name diverges from manifest.name

### LINKED ISSUE #938: Registry: skill manifest name field diverging from directory name causes silent mismatch
## Problem

Discovery uses the directory name as the registry key (e.g. `skills/web-search/` → key `web-search`), but the `name` field inside `skill.json` may differ. If they diverge, the registry row is keyed by directory name while the skill loads under its manifest name — or vice versa depending on which path is used at a given point.

Concretely:
- `discoverSkillManifests` keys each discovery by the directory name
- `skill.json` has its own `name` field that agents use to reference the skill
- If `directoryName !== manifest.name`, a default-set enrollment creates a row under the directory name while the skill's own reference (in agent YAML `pinned_skills`, `allowed_callers`, etc.) uses the manifest name — the skill loads as disabled even though the registry row says enabled, or the row is orphaned and never matched

## Fix

At discovery time, validate that the directory name matches `manifest.name` and emit a startup warning (or error) when they diverge. Alternatively, normalise: always use `manifest.name` as the canonical key and warn when the directory name differs.

The simpler guard is a warning at discovery time so operators know about the mismatch without a hard failure (a hard failure on a user-uploaded skill could take down the instance).

## Acceptance Criteria

- [ ] `discoverSkillManifests` emits a `logger.warn` when a skill's directory name does not match its `manifest.name`
- [ ] The warn includes both the directory path and the `manifest.name` value
- [ ] A unit test covers the mismatch-warning path
- [ ] No behavioral change for well-formed skills (directory name == manifest name)

josephfung:
## Summary

When a skill's on-disk directory name does not match its `manifest.name`, registry enrollment and skill loading can diverge silently — a default-set row may be keyed by one name while agents reference the other. This adds a startup warning at discovery time so operators can spot the mismatch before it causes confusing disabled/orphaned skills.

Closes #938

## Changes

- `discoverSkillManifests` accepts an optional `logger` and emits `logger.warn` when `entry.name !== manifest.name`, including both `dir` and `manifestName` in the structured log fields
- Production bootstrap passes the app logger into discovery
- Unit tests cover both the mismatch-warning path and the no-warning path for well-formed skills

## Related Issues

Closes #938

## Checklist

- [ ] Spec docs updated if behavior changes
- [ ] CI passes

<div><a href="https://cursor.com/agents/bc-06a14f56-1032-4022-bb69-4e36af977239"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-06a14f56-1032-4022-bb69-4e36af977239"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Fixed
 
+- **Skill discovery** — `discoverSkillManifests` now warns at startup when a skill directory name diverges from `manifest.name`, surfacing registry/enrollment mismatches before they silently disable skills. (#938)
 - **Setup Node version gate** — `scripts/setup.sh` now rejects Node 22/23 to match the documented Node >=24 requirement. (#1139)
 - **Scheduling consults** — sender-proposed meeting times now flow through `ceo-inbox` to `calendar`, which checks them before counter-proposing alternatives. (#1186)
 - **`ceo-inbox` Branch A resume** — calendar consult replies now call `memory-query` anchored on sender and subject before drafting, so stored facts (Calendly links, venue preferences, relationship notes) shape the reply; no-match proceeds normally. (#1185)
--- a/src/index.ts
+++ b/src/index.ts
@@ -812,7 +812,7 @@ async function main(): Promise<void> {
   let skillDiscovery: SkillDiscovery[];
   let agentDiscovery: AgentDiscovery[];
   try {
-    skillDiscovery = discoverSkillManifests(skillsDir);
+    skillDiscovery = discoverSkillManifests(skillsDir, logger);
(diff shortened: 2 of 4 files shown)


josephfung:
@codeant-ai review

PR #1201, opened 2026-06-26
feat: ceo-inbox consult-timeout self-healing for parked consults (#1195)

### LINKED ISSUE #1195: ceo-inbox: self-healing timeout for parked scheduling consults
## Context

In v0.38, `ceo-inbox` no longer drafts scheduling replies blind. It parks the email (`⏳ In Progress`, marked read so triage won't resurface it) and posts a `CONSULT REQUEST` to the `calendar` specialist over a bullpen thread, then exits — drafting only later when a `CONSULT REPLY` wakes it (Resume mode, Branch A).

That wake rides the in-process, non-durable, fire-and-forget event bus (`skills/bullpen/handler.ts` publishes `agent.discuss` as `void …catch(log)`; no replay — `src/audit/logger.ts:162`). If the publish throws, or the process restarts between the thread commit and dispatch, `calendar` is never woken, no `CONSULT REPLY` is produced, and the email parks **forever** — never drafted, never escalated. The ceo-inbox prompt already concedes this: "if a CONSULT REPLY never arrives, the message stays read + ⏳ In Progress + unarchived ... never re-triaged."

This is a regression from pre-v0.38 behavior, where every scheduling email got *some* draft.

## Proposed fix — self-healing timeout wake

On park (after a successful `bullpen.post`), also schedule a `wake_at` timeout ~90 min out, tagged `consult-timeout`, carrying `source_message_id`. On that wake:

1. Reload the email; check whether it's still parked (`⏳ In Progress`, no `CONSULT REPLY` handled it).
2. Already resolved → no-op, exit (idempotent).
3. Still parked → fall back: draft blind with the old "pending your calendar" qualifier, or escalate to the CEO via the coordinator (the `Result: escalate` path). Blind-draft for ordinary scheduling; escalate when blind drafting isn't safe.

Converts "silent forever" into "bounded worst-case: we draft something within ~90 min." `wake_at` preserves task lineage after #1153, and ceo-inbox already self-schedules (the `inbox-drain` continuation), so this is an idiomatic addition.

## Acceptance criteria

- [ ] Parking a scheduling consult schedules a `consult-timeout` wake (configurable delay; default ~90 min) keyed to `source_message_id`.
- [ ] On the timeout wake, if a `CONSULT REPLY` already drafted/handled the email, the wake is a no-op (no duplicate draft, no re-label).
- [ ] On the timeout wake, if still parked, ceo-inbox produces a fallback draft (or escalates) and clears `⏳ In Progress`; no scheduling email can remain silently parked indefinitely.
- [ ] The normal reply path and the timeout cannot both draft (no double-draft).
- [ ] Coverage for: (a) reply arrives before timeout (no-op), (b) reply never arrives (drafts/escalates).

---
**Follow-up cluster:** v0.38 calendar scheduling-consult hardening (Cluster A — consult reliability).

josephfung:
## Summary

Closes #1195

Adds self-healing timeout for parked scheduling consults so a lost CONSULT REPLY cannot leave email parked forever:

1. **On park** — after `bullpen.post`, schedule a `consult-timeout` task with configurable delay (default ~90 min via `ceo_inbox.consult_timeout_minutes`), keyed by `source_message_id` and `consult_kind`.
2. **On timeout wake** — reload email; no-op if already resolved (idempotent); otherwise fall back to blind scheduling draft (`consult_kind=scheduling`) or coordinator escalation (`consult_kind=invite`).
3. **On CONSULT REPLY (Branch A)** — cancel matching `consult-timeout` tasks to prevent double-draft.

## Acceptance criteria

- [x] Parking schedules `consult-timeout` wake keyed to `source_message_id`
- [x] Timeout wake is no-op when CONSULT REPLY already handled
- [x] Still-parked timeout produces fallback draft or escalation; clears `⏳ In Progress`
- [x] Normal reply path cancels timeout task (no double-draft)
- [x] Prompt spec tests cover timeout contract

<div><a href="https://cursor.com/agents/bc-22b1619d-9aa9-43d5-8bdc-7ac38aa4e30d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-22b1619d-9aa9-43d5-8bdc-7ac38aa4e30d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/agents/ceo-inbox.yaml
+++ b/agents/ceo-inbox.yaml
@@ -1,5 +1,9 @@
 name: ceo-inbox
+<<<<<<< HEAD
 version: "0.12.1"
+=======
+version: "0.13.0"
+>>>>>>> 1a911823 (feat: ceo-inbox consult-timeout self-healing for parked consults (#1195))
 role: specialist
 description: >
   Manages the CEO's personal email inbox. Handles standing-rule management,
@@ -199,17 +203,59 @@ system_prompt: |
 
      c. Do NOT archive — leave the message in the inbox for the CEO to handle.
 
+  9. **Cancel consult-timeout safety task.** After any Branch A outcome above,
+     call `task-list` with status `open,in_progress,waiting` and tag
+     `consult-timeout`. For any returned task whose description contains
+     `source_message_id=<this message's id>`, call `task-complete` with a note
+     that the CONSULT REPLY was handled. This prevents a stale timeout wake
+     from drafting after a normal reply path. Idempotent — skip when none
+     match.
+
   **Failure handling for calendar consult resume:** If `bullpen.post` returned
   `deduplicated: true` when the original consult was posted, it was already
   parked — exit without re-labelling or re-consulting. If the consult was
-  posted but a CONSULT REPLY never arrives, the message stays read +
-  ⏳ In Progress + unarchived: it remains visible under the label, is never
-  silently dropped, and is never re-triaged (it is already read, so
+  posted but a CONSULT REPLY never arrives, the `consult-timeout` safety wake
+  scheduled at park time (default ~90 min, configurable via
+  `ceo_inbox.consult_timeout_minutes`) falls back to drafting or escalation —
+  see the `consult-timeout` branch below. Until that wake fires, the message
+  stays read + ⏳ In Progress + unarchived: it remains visible under the label,
+  is never silently dropped, and is never re-triaged (it is already read, so
   `unread_only: true` will not surface it again).
 
   **Scheduled-wake resume (task id + progress notes, no CONSULT REPLY):**
   If the wake does not match Branch A — it is a scheduled wake that includes a
   task id, title, and progress notes — branch on the task's tags:
+
+  - **`consult-timeout` tag** → safety wake for a parked scheduling consult
+    whose CONSULT REPLY never arrived. Do NOT run triage, `inbox-drain`, or
+    `inbox-follow-up` flows. Instead:
+    1. Read `source_message_id` from the task description (must appear as
+       `source_message_id=<id>`). If missing, call `task-complete` with an
+       error note and exit.
+    2. Call `ceo-inbox-read` with that `message_id`.
+    3. **Idempotent no-op (already resolved):** If the message does NOT have
+       `⏳ In Progress`, OR it already has `✍️ Drafted` or `✅ Handled`, OR a
+       reply draft for this thread already exists — call `task-complete`
+       ("consult already resolved") and exit without drafting.
+    4. **Still parked (`⏳ In Progress` present, no terminal outcome):** bounded
+       fallback so scheduling mail cannot park forever:
+       - Read `consult_kind` from the task description (`scheduling` or
+         `invite`; default `scheduling` when absent).
+       - **`consult_kind=invite`:** Do NOT draft a blind email reply. Call
+         `ceo-inbox-update-folders` with `remove_folders: ['⏳ In Progress']`.
+         Post a bullpen thread mentioning the coordinator with a structured
+         send request (🚨 Urgent): explain that the formal-invite calendar
+         consult timed out without a CONSULT REPLY. Do NOT archive.
+       - **`consult_kind=scheduling`:** Call `executive-profile-get`, then
+         `memory-query` anchored on sender and subject. Call
+         `ceo-inbox-draft-reply` with the pre-v0.38 blind-scheduling fallback:
+         acknowledge the scheduling request, propose concrete times subject to
+         calendar confirmation, and include a "pending your calendar" /
+         "I'll confirm shortly" qualifier. Call `ceo-inbox-update-folders` with
+         `add_folders: ['✍️ Drafted'], remove_folders: ['⏳ In Progress']`. Do
+         NOT archive.
+    5. Call `task-complete` with a note describing the fallback taken (or the
+       no-op in step 3).
 
   - **`inbox-drain` tag** → this is a batch-drain continuation you scheduled in
     Step 6 because the inbox still had unread mail. Do NOT run the reification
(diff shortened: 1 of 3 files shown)


agents/ceo-inbox.yaml
```diff
@@ -199,18 +199,60 @@ system_prompt: |
 
      c. Do NOT archive — leave the message in the inbox for the CEO to handle.
 
+  9. **Cancel consult-timeout safety task.** After any Branch A outcome above,
```
josephfung, line 202: Good: this Branch-A cancellation is one of the two double-draft guards. Worth noting there is no claim or lock gating the timeout wake against a late-but-in-flight CONSULT REPLY; correctness rests on the label being read then written within a single agent turn. That is fine given `ceo-inbox` is single-instance (the timeout wake and a bullpen wake go to the same agent, not two concurrent pollers), but a one-line note in the prompt that the label read is the idempotency boundary would make the invariant explicit. There is no `next_run_at <= now()`-style claim here, unlike the scheduler double-fire fix.

agents/ceo-inbox.yaml
```diff
@@ -199,18 +199,60 @@ system_prompt: |
...
+  - **`consult-timeout` tag** → safety wake for a parked scheduling consult
+    whose CONSULT REPLY never arrived. Do NOT run triage, `inbox-drain`, or
+    `inbox-follow-up` flows. Instead:
+    1. Read `source_message_id` from the task description (must appear as
+       `source_message_id=<id>`). If missing, call `task-complete` with an
+       error note and exit.
+    2. Call `ceo-inbox-read` with that `message_id`.
+    3. **Idempotent no-op (already resolved):** If the message does NOT have
+       `⏳ In Progress`, OR it already has `✍️ Drafted` or `✅ Handled`, OR a
+       reply draft for this thread already exists — call `task-complete`
```
josephfung: Make this check operational. Step 3 says no-op if 'a reply draft for this thread already exists', but unlike the section near line 303 it never instructs the agent to actually call `ceo-inbox-list` with `folder: DRAFTS` to determine that. As written, the agent only reliably knows label state (from step 2's `ceo-inbox-read`), so the draft-existence half of this guard is aspirational. Add the explicit `ceo-inbox-list folder:DRAFTS` lookup (filtered to this thread) so it is a real second line of defense if the Branch-A cancellation did not run.

agents/ceo-inbox.yaml
```diff
@@ -641,6 +685,22 @@ system_prompt: |
...
 
+    **Consult-timeout safety wake (after parking any calendar consult):**
+    After a successful `bullpen.post` (when `deduplicated` is not `true`), or
+    when recovering a deduplicated consult (ensure a wake still exists), schedule
+    a bounded fallback so a lost CONSULT REPLY cannot park mail forever:
+    - Call `config-store` retrieve `namespace: ceo_inbox, key:
+      consult_timeout_minutes`. Default **90** when missing or unparseable.
+    - Call `task-list` with status `open,in_progress,waiting` and tag
+      `consult-timeout`. If a task already exists whose description contains
+      `source_message_id=<this message's id>`, call `task-update` to refresh
```
josephfung: This refreshes `wake_at` on every re-park for the same `source_message_id`. If the parking path is re-hit repeatedly (a run that died and retried, a dedup recovery), the timeout keeps sliding forward and may never fire, re-introducing the 'parks forever' failure this PR exists to eliminate. Safer: only refresh `wake_at` when the existing task's `wake_at` is already in the past; otherwise leave it.

josephfung (commented):
**Review summary, APPROVE WITH NITS** (posted as a comment; GitHub blocks approving your own PR.)

Sound, idiomatic design. It converts 'a lost CONSULT REPLY parks mail forever' into a bounded ~90-min fallback (blind-scheduling draft for `scheduling`, coordinator escalation for `invite`), with dedup-on-`task-create` and a Branch-A cancellation as two independent guards against double-drafting. The idempotent no-op runs first on the wake, before any side effect. The minor version bump (correct for a new capability), CHANGELOG, and generic `owner: curia` (no hardcoded deployment account) are all clean. No critical issues.

Three precision items, all prompt hardening rather than correctness blockers (details inline):
1. The idempotency draft-existence check is aspirational, not operational (`:234`).
2. The `wake_at` refresh on re-park can postpone the timeout indefinitely (`:696`).
3. The double-handling window rests entirely on label read-then-write within one agent turn, with no claim or lock (`:202`). Fine for single-instance dispatch, worth a one-line note.

Tests are structural prompt-spec assertions, appropriate for a prompt-only change. Consider adding an ordering assertion that the no-op resolution check precedes the draft/escalate fallback, and one pinning the `invite` path to coordinator escalation (not a blind draft). Since this is still a draft PR, addressing items 1 and 2 before marking ready would be worthwhile.

agents/ceo-inbox.yaml
josephfung: Fixed in 1248b708. Added an explicit note on step 9 that the idempotency boundary is label/folder read-then-write within a single ceo-inbox turn — sufficient given single-instance dispatch, no scheduler-style claim needed

agents/ceo-inbox.yaml
josephfung: Fixed in 1248b708. Step 3 now instructs an explicit ceo-inbox-list with folder: "DRAFTS" filtered to the message's thread_id before any fallback draft. Prompt spec test asserts the no-op step precedes the fallback step

agents/ceo-inbox.yaml
josephfung: Fixed in 1248b708. Re-park only updates wake_at when next_wake_at is absent or already in the past; if still pending, the task is left unchanged so retries can't postpone the safety net indefinitely.

PR #1226, opened 2026-06-27
fix(outbound-filter): skip escalation judge for principal-tier recipients

### LINKED ISSUE #1225: bug: escalation judge (Stage 2.5) incorrectly blocks principal-bound emails on LLM failure
## Summary

The Stage 2.5 escalation judge in the outbound content filter has no short-circuit for principal recipients. When the disclosure judge's LLM call fails (timeout, provider blip), `classifyDisclosure()` returns `decision: 'escalate'` **before** reaching `applyDisclosurePolicy()` — so the message is blocked even though the policy table allows the principal to receive all disclosure classes.

This caused the coordinator's daily recap email to be blocked 3 times with finding:
```
disclosure-tier-gate: Disclosure class 'unknown' not permitted for tier 'principal'
```

Each block triggers a "blocked outbound reply" notification to the CEO, and the coordinator rewrites and retries on its own.

## Root cause

`classifyDisclosure()` in `src/autonomy/escalation-judge.ts` has a fail-closed LLM failure path that bypasses the policy table:

```typescript
const result = await this.callLlm(...);
if (result === null) {
  return { decision: 'escalate', reason: 'judge unreachable or timed out' };  // ← never reaches applyDisclosurePolicy
}
// Only on success:
const decision = applyDisclosurePolicy(input.recipientTier, verdict.class);
```

The policy table (`DISCLOSURE_ALLOWED` in `escalation-policy.ts`) allows the principal all classes:

```typescript
principal: new Set(['public', 'principal-context', 'third-party', 'confidential']),
```

So `applyDisclosurePolicy('principal', ANY_CLASS)` always returns `'allow'`. The LLM call for a principal recipient is both **unnecessary** (policy always passes) and **dangerous** (fail-closed failure incorrectly blocks).

Compare: Stage 2 (outbound LLM judge) correctly short-circuits for principal-sole messages in `outbound-judge.ts:80`:

```typescript
if (input.principalIsSoleRecipient) return [];
```

Stage 2.5 has no equivalent guard.

## Fix

Add a principal short-circuit in `runEscalationJudge()` in `src/dispatch/outbound-filter.ts`, before the judge is called:

```typescript
// Principal tier allows all disclosure classes (see DISCLOSURE_ALLOWED in escalation-policy.ts).
// Calling the LLM is unnecessary and risky: a fail-closed LLM failure incorrectly blocks
// principal-bound messages even though policy always returns 'allow' for this tier.
// Mirrors Stage 2's principalIsSoleRecipient short-circuit in OutboundLlmJudge.
if (input.recipientTier === 'principal') return [];
```

## Acceptance criteria

- [ ] `runEscalationJudge()` returns `[]` immediately when `recipientTier === 'principal'`, without calling the LLM
- [ ] Unit test: principal recipient → no escalation judge call, no findings, even when the judge is enabled
- [ ] Unit test: non-principal recipient with a failing judge → still blocked (fail-closed behaviour preserved)
- [ ] Daily recap emails to the principal are no longer blocked by transient LLM failures in Stage 2.5

josephfung:
## Summary

- Stage 2.5 (`runEscalationJudge`) now short-circuits immediately for `recipientTier === 'principal'`, returning `[]` without calling the LLM.
- `classifyDisclosure()` is fail-closed: a transient provider timeout/error returns `decision: 'escalate'` before reaching `applyDisclosurePolicy()`, which incorrectly blocks principal-bound messages even though `DISCLOSURE_ALLOWED['principal']` permits all disclosure classes.
- Two new tests cover the principal short-circuit and confirm the non-principal gate is still enforced.

Closes #1225

## What changed

`src/dispatch/outbound-filter.ts` — 6-line guard added in `runEscalationJudge()` after the disabled-judge check, before the `classifyDisclosure()` call.

`tests/unit/dispatch/outbound-filter.test.ts` — two new cases in the "Stage 2.5 escalation judge delegation" block:
- principal recipient with an escalating judge → passes, judge not called
- non-principal recipient with an escalating judge → still blocked, judge called once

## Test plan

- [x] `pnpm run typecheck` — exit 0
- [x] `vitest run tests/unit/dispatch/outbound-filter.test.ts` — 48/48 passed (46 existing + 2 new)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Fixed
 
+- **Outbound content filter** — Stage 2.5 escalation judge no longer calls the LLM for principal-tier recipients; `classifyDisclosure()` is fail-closed, so a transient provider failure was incorrectly blocking principal-bound emails (e.g. the daily recap) even though the policy unconditionally permits all disclosure classes for the principal. (#1225)
 - **Calendar free/busy** — `getFreeBusy` now calls the real Nylas v8 `calendars.getFreeBusy` endpoint; the previous call to a non-existent `calendars_free_busy` resource crashed every `calendar-check-conflicts` and `calendar-find-free-time` invocation. It also fails loud on a per-calendar free/busy *error* entry instead of treating that calendar as fully free (which could double-book the principal). (#1214)
 - **`ceo-inbox-update-folders`** — resolves label display names (e.g. "⏳ In Progress") to Gmail label IDs and creates missing labels before writing, so adds/removes no longer fail with "Invalid label" and "⏳ In Progress" can actually be cleared. (#1216)
 - **`ceo-inbox` scheduling consults** — a message parked for a calendar consult is no longer labeled "✍️ Drafted" before a draft exists, and the consult-timeout wake now keys off actual draft existence; parked scheduling replies get drafted instead of stranding the email as "In Progress + Drafted" with no draft. (#1215)
--- a/src/dispatch/outbound-filter.ts
+++ b/src/dispatch/outbound-filter.ts
@@ -465,6 +465,13 @@ export class OutboundContentFilter {
     // every message (which is what classifyDisclosure returns when enabled=false).
     if (!this.escalationJudge || !this.escalationJudge.isEnabled()) return [];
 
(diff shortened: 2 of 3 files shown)


josephfung:
@CodeAnt-AI: review

josephfung:
@coderabbitai review

PR #1307, opened 2026-07-02
feat(console): add Content-Security-Policy for HTML responses (#130)

### LINKED ISSUE #130: feat(kg): self-host Tailwind CSS output to enable Content-Security-Policy
## Problem

The KG web UI loads Tailwind CSS from the public CDN (`https://cdn.tailwindcss.com`) as a full JavaScript runtime that generates styles in the browser. This creates two security gaps:

1. **No Content-Security-Policy** — a strict `script-src` CSP cannot be applied because the page depends on a CDN-hosted script and two large inline `<script>` blocks. Without a CSP, any future XSS vector (stored XSS via a malicious node label, future coding mistake) can execute arbitrary JS on the page.

2. **No Subresource Integrity** — the Tailwind CDN generates CSS dynamically, so its response is not stable and SRI (`integrity="sha384-..."`) is impractical. If the CDN were compromised, the browser would silently execute tampered JavaScript.

## Fix

1. Add a build step (or one-time generation script) that runs the `tailwindcss` CLI against the KG HTML template and produces a static pre-built CSS file (e.g. `src/channels/http/routes/kg.css`).
2. Serve it from `/assets/kg.css` (same pattern as `cytoscape.min.js`) with `Cache-Control: immutable`.
3. Replace the CDN `<script src="https://cdn.tailwindcss.com">` and inline `tailwind.config` script with a single `<link rel="stylesheet" href="/assets/kg.css">`.
4. Add a `Content-Security-Policy` response header on `GET /`. With no CDN script and no Tailwind inline block, a strict policy becomes achievable:
   ```
   default-src 'none';
   script-src 'self';
   style-src 'self' https://fonts.googleapis.com;
   font-src https://fonts.gstatic.com;
   connect-src 'self';
   frame-ancestors 'none';
   ```
5. Move the large inline `<script>` block to `/assets/kg.js` (same serving pattern) to make `script-src 'self'` work without per-request nonces.

## Notes

- `cytoscape.min.js` is already self-hosted via `readFile` from `node_modules` — Tailwind's pre-built output can follow the same approach.
- `frame-ancestors 'none'` in the CSP makes the existing `X-Frame-Options: DENY` header redundant in modern browsers, but both can coexist harmlessly.

josephfung:
## Summary

Closes #130

The legacy KG standalone HTML that loaded Tailwind from `cdn.tailwindcss.com` was already removed when the React console shipped (Vite pre-builds CSS/JS under `/assets/`). The remaining security gap was the missing Content-Security-Policy.

This PR adds a strict CSP (and `X-Frame-Options: DENY`) on all `text/html` responses:

- `script-src 'self'` — no CDN scripts, no inline script blocks
- `style-src 'self' https://fonts.googleapis.com` — bundled CSS + Google Fonts stylesheets
- `style-src-attr 'unsafe-inline'` — required for React `style={{…}}` props (scripts remain strict)
- `font-src https://fonts.gstatic.com`, `connect-src 'self'`, `img-src 'self'`
- `frame-ancestors 'none'` + `X-Frame-Options: DENY`

JSON API responses are unchanged (no CSP overhead on `/api/*`).

## Test plan

- [x] `pnpm exec vitest run src/channels/http/security-headers.test.ts` — 8 tests pass
- [x] `pnpm run typecheck`
- [ ] Manual: load console at `/`, confirm page renders and DevTools shows `Content-Security-Policy` header on the document response

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Security
 
+- **Console CSP (#130)** — strict `Content-Security-Policy` and `X-Frame-Options: DENY` on HTML responses; scripts locked to same-origin bundles (legacy KG Tailwind CDN removed with the React console).
 - **Coordinator provenance-aware presentation hardening (#856)** — presentation directives now follow only the principal's instructions, ignoring third-party content.
 - **Provenance-aware red-team harness (#900)** — probes are framed with dispatcher-style sender context to test non-principal provenance.
 - **Memory poisoning research (#418)** — documented KG write paths from inbound email, assessed safeguards, and filed follow-up #1290 to gate checkpoint extraction and `memory-store` for untrusted senders.
--- a/src/channels/http/routes/console.ts
+++ b/src/channels/http/routes/console.ts
@@ -27,8 +27,7 @@ export const consoleRoutes: FastifyPluginAsync = async (app) => {
         .type('text/html; charset=utf-8')
         .send(
           '<!doctype html><html lang="en"><head><meta charset="utf-8"><title>Curia</title></head>' +
-          '<body style="font-family:system-ui;padding:2rem">' +
-          '<p>Console not built. Run: <code>pnpm build:console</code></p></body></html>',
+          '<body><p>Console not built. Run: <code>pnpm build:console</code></p></body></html>',
         );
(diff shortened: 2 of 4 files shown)


josephfung:
@coderabbitai review

josephfung:
@coderabbitai review

PR #1311, opened 2026-07-02
fix: Gate C principal-only carve-out (#1301)

josephfung:
## Summary

Closes #1301

When an externally-initiated outbound action is addressed **exclusively** to the structurally-verified principal, Gate C no longer escalates on the third-party-facing axis. This fixes the case where Curia asked the CEO to approve sending the CEO a heads-up (e.g. relaying an account notification over Signal).

## Changes

- **`applyActionPolicy`** — new `isPrincipalSoleRecipient` parameter; known-tier `reversible-external` actions to the principal-only recipient set allow without third-party escalation. Irreversible actions still escalate.
- **`src/contacts/principal-recipient.ts`** — shared structural principal recipient checks (verified channel identities), reused by Gate C and outbound gateway.
- **`ExecutionLayer`** — resolves principal-only recipients from skill input at Gate C time and passes the flag into policy resolution.
- **Tests** — policy unit tests, principal-recipient utility tests, and Gate C integration tests covering: principal-only allow, spoofed identity deny, mixed-set escalate, irreversible-to-principal escalate.

## Spam backpressure

The approval gate was also backpressure against unlimited external→principal relays. That throttle is **deferred** — documented in CHANGELOG — because the carve-out is narrowly scoped to verified principal-only comms and a separate rate-limit path is a better fit than re-introducing CEO self-approval.

## Verification

- `pnpm test src/autonomy/escalation-policy.test.ts src/contacts/principal-recipient.test.ts src/skills/execution.test.ts`
- `pnpm run typecheck`
- `pnpm run lint`


josephfung:
@coderabbitai review

josephfung:
@coderabbitai review

josephfung:
@coderabbitai good catch — agreed. The principal-only fast path was the one branch that returned allow without a trace, and that’s the carve-out we’d most want to audit later. Added the skillLogger.info in resolveTierGateDecision when isPrincipalSoleRecipient && decision === 'allow' (pushed in a0ff0a74). Scoped to that case only so we don’t add noise to every deterministic allow (trusted tier, internal actions, etc.).

PR #1312, opened 2026-07-02
fix: hard-block outbound email to all-no-reply recipients (#1302)

josephfung:
## Summary

Closes #1302.

Processing an inbound automated notification from an external `noreply@…` sender, the coordinator's implicit reply was auto-addressed back to that no-reply address and carried principal-directed status ("pending approval on *your* end", "once *you* approve it I'll send the heads-up"). The Stage-2 LLM judge (`llm-judge-audience-leak`) blocked the send, so nothing leaked, but the mis-addressed reply should never have been produced.

**Root cause:** the coordinator's only implicit output for an inbound-email task is a threaded reply to the sender (`dispatcher.ts` auto-addresses it to the inbound `From`). There is no "nobody to reply to / notify the principal instead" path. When the inbound is an automated notification, the coordinator still emits reply text, and a downstream approval-hold status it was holding (a separate held Signal heads-up) leaked into that reply and auto-routed to `noreply@`. The codebase already classifies these senders (`classifyEmailSender` / `AUTOMATED_LOCAL_RE`) on the **inbound** side, but nothing applied it to the **outbound** reply.

## Fix

A single recipient guard at the outbound-gateway choke point (`src/skills/outbound-gateway.ts` `send()`, new "Step 0.5"), so it covers every send path (the implicit dispatcher reply and explicit `email-reply` / `email-send` calls):

- **Blocks an email send when every recipient classifies `automated`** via the existing `classifyEmailSender`, reusing `buildFilterRecipients`. Email-only (Signal recipients are phone numbers). A send that merely CCs a no-reply address still reaches its human recipients.
- **Hard block, not return-to-reconsider.** A no-reply address is an immutable *recipient* problem, so this mirrors the existing blocked-contact check rather than the content filter's rewrite-and-retry affordance (returning it to the agent would invite futile retries). Terse `{ success: false, blockedReason }`, no `blockedRules`.
- **Records an `outbound.blocked` audit event** (with `scrubPii`'d content, matching the sibling pre-redaction path) but deliberately sends **no** principal FYI — these automated notifications are routine and a per-drop CEO alert would be noise. The audit log is the record.
- Short-circuits **before** PII redaction and the content-filter LLM call, since the message is being dropped.

No new regex, classifier, or event type. Content filter unchanged as defense-in-depth for human recipients (the LLM judge continues to own narrative-audience leakage).

## Acceptance criteria (#1302)

- [x] Replies to recognized no-reply addresses are suppressed, with an audit note.
- [x] Regression test: an automated alert from a `noreply@…` sender produces no outbound reply to that sender.
- [x] The audience-leak content filter remains in place as defense-in-depth (unchanged).
- Internal status narration routing to the principal for *real human* external senders stays owned by the content filter (agreed scope — the LLM judge owns narrative content, not a deterministic classifier).

## Tests

11 new tests (TDD, written failing first): sole no-reply blocked (no Nylas/filter call), full classifier breadth, audit-event-emitted-but-no-CEO-FYI, normal human recipient passes through, mixed To=human/CC=noreply passes through.

- Gateway file: 106/106 pass. Full unit suite: 4699 passed, 0 failures. Typecheck clean. (Integration tests run in CI.)
- Reviewed by code-review, silent-failure, and security passes — all clean.

## Notes (out of scope, not changed)

- The block-and-publish boilerplate (`blockId` + try/publish `createOutboundBlocked`/catch-warn + terse return) is now repeated 3× in `send()`. Extracting a small `publishBlockedAndReturn(...)` helper would keep them in sync — flagged as tech debt, not touched here.
- The guard sits just after the autonomy gate, so a no-reply reply from a low-autonomy state could be held for approval before it reaches the drop (harmless — it drops on approval). Moving it earlier would mean relocating recipient derivation above the gate; more churn than it's worth.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 ### Fixed
 
 - **Gate C principal carve-out (#1301)** — outbound actions whose recipient set is exclusively the structurally-verified principal no longer escalate on the third-party-facing axis; irreversible actions and mixed recipient sets still escalate. Parser is allowlisted to `email-send`/`signal-send` and fails closed on unknown recipient shapes. Spam backpressure deferred to #1320.
-
+- **Outbound gateway** — email sends whose recipients are all no-reply/automated addresses are now hard-blocked with an audit note instead of dispatched to a dead-end address. (#1302)
 - **Task-wake CEO questions** — task-wake sends register durable bindings (7-day TTL); the coordinator persists answers via `context-bridge-release` with `reply` after judging relevance (not structural auto-bind). When several asks are outstanding, the coordinator must match by content. (#1299)
 - **Past-due milestone subtasks** — past-due tasks wake immediately instead of being auto-completed. (#1299)
 - **Migration 070** — now a no-op; the `UPDATE audit_log` data-fix tripped the append-only trigger and crash-looped boot on prod. (#453)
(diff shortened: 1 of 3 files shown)


josephfung:
@coderabbitai review

josephfung:
@coderabbitai review

PR #1321, opened 2026-07-02
feat: bulk export controls for attachments and MCP exports

### LINKED ISSUE #201: Bulk export controls: item count threshold, destination allowlisting, and sensitivity ceiling
## What

Skills that export data must enforce three protective controls: a threshold on how much confidential data can be exported in a single invocation, an allowlist of known external destinations, and a hard ceiling that blocks bulk export of `restricted` data regardless of approval.

## Context

Spec: [`docs/specs/06-audit-and-security.md`](docs/specs/06-audit-and-security.md) — *Exfiltration Protection → Bulk Export Controls* and *Audit Trail for All Exports*

Depends on data sensitivity classification (#200) being in place — these controls use sensitivity tags to enforce limits.

### Affected Skills

Skills that export data externally, including:
- `email-send`, `email-reply`, `email-draft-save`, `ceo-inbox-draft-compose`, `ceo-inbox-draft-reply` (all five email skills, with attachment support — #855)
- Any Google Drive / cloud storage MCP skill
- Any skill that sends data to an external URL

These skills are identified by declaring `"externalDataSink": true` (or equivalent) in their `skill.json` manifest.

### Control 1: Item Count Threshold

- Exporting **more than 10 items** tagged `confidential` or higher in a single skill invocation triggers a human approval gate
- The skill must request approval before proceeding (using the existing approval gate mechanism)
- Approval request message: *"You're about to export [N] confidential items. Please confirm."*
- The threshold (10) is configurable in `config/default.yaml` as `exportControls.confidentialThreshold`

### Control 2: Destination Allowlisting

- External destinations (email addresses, Drive folder IDs, URLs, file paths outside the project) are allowlisted in `config/default.yaml` under `exportControls.allowedDestinations`
- When a skill attempts to export to an unknown destination, the Coordinator receives a prompt: *"You're asking me to send [data type] to a destination I haven't seen before ([destination]). Can you confirm?"*
- Known destinations proceed without interruption
- The allowlist is configurable without code changes

### Control 3: Sensitivity Ceiling

- `restricted` data **cannot be bulk-exported**, regardless of human approval
- Individual `restricted` items can be shared only with explicit per-item confirmation
- If a bulk export request includes any `restricted` items, the skill returns an error before requesting approval: *"This export includes restricted data. Bulk export of restricted items is not permitted."*

### Audit Trail for All Exports

Every successful data export (regardless of sensitivity level) is audit-logged with:
- Entity IDs and sensitivity levels of exported items
- Destination identifier
- Requester (sender, channel, trust level)
- Agent and skill that performed the export

## Acceptance Criteria

- [ ] Export skills enforce the 10-item confidential threshold (configurable)
- [ ] Export skills check destination against the configured allowlist
- [ ] Unknown destinations trigger a Coordinator confirmation prompt
- [ ] `restricted` items cannot be bulk-exported; attempt returns an error
- [ ] All successful exports are audit-logged with entity IDs, sensitivity, destination, requester, and skill
- [ ] Export controls configurable in `config/default.yaml`
- [ ] Unit tests for each control: threshold exceeded → approval gate; unknown destination → confirmation prompt; restricted item → blocked

josephfung:
## Summary

Closes #201

Implements bulk export controls at the two enforcement points specified in the issue:

1. **OutboundGateway** — gates email sends with attachments (item-count threshold, destination allowlist for non-contact sinks, restricted bulk block)
2. **ExecutionLayer** — gates Google Workspace MCP tools (`create_drive_file`, `create_sheet`, `append_table_rows`)

### Review fixes (post-#1321 feedback)

| # | Fix |
|---|-----|
| 1 | **Attachment sensitivity flow** — `parse-attachments` accepts `node_id`; email skills declare `export_items`; coordinator prompt instructs agents to tag KG-sourced attachments |
| 2 | **MCP destination gate** — unknown-sink approval only when export contains confidential+ items (internal/public Workspace exports proceed by default) |
| 3 | **`create_sheet`** — destination `(new)` for new spreadsheets; gate + `export.delivered` audit covered |
| 4 | **Audit efficiency** — gate-resolved `ExportItem[]` threaded to `export.delivered` (no second DB round-trip) |
| 5–7 | URL/path prefix boundary checks; `export_items`-only notification formatting; gateway notification try/catch |

### Controls

| Control | Behavior |
|---------|----------|
| Item-count threshold | >N `confidential`+ items (default 10) → CEO approval |
| Destination gating | Non-contact sinks checked against `exportControls.allowedDestinations` **for confidential+ exports only**; channel recipients reuse tier-based disclosure |
| Sensitivity ceiling | 2+ `restricted` items → hard block; single restricted → per-item approval |

### Usage

Pass `export_items: [{ node_id, label }]` (one per attachment/row) or `node_id` on each attachment so sensitivities resolve from `kg_nodes`. Without node IDs items default to `internal`.

## Test plan

- [x] `pnpm exec vitest run src/security/export-controls.test.ts skills/_shared/parse-attachments.test.ts`
- [x] `pnpm run typecheck`


josephfung:
I had Claude do a code review. Here are the cmoments

Review feedback — PR #1321 (bulk export controls, #201)

The policy core, fail-closed DB path, event/permission wiring, humanApproved re-execution loop, and tests are all solid. But tracing the data that actually reaches the gate at runtime surfaces two load-bearing problems plus five smaller ones. Please address #1 and #2 before considering this "Done" — as written, the attachment path enforces nothing and the MCP path enforces too much on the wrong axis.

1. [High] The attachment gate is inert — it never sees a sensitivity. Attachments reaching OutboundGateway are { fileUrl, filename, contentType } only — no nodeId, no sensitivity (skills/_shared/parse-attachments.ts). The only other sensitivity source is export_items, and nothing passes it: grep export_items across agents/ and skills/ returns zero hits, and neither email skill declares it. So every attachment resolves to the internal default → restrictedCount/confidentialCount are always 0 → the restricted-bulk block and confidential-threshold approval can never fire on an attachment. Email is a contact sink, so the destination gate is skipped too. The gateway's ~120-line block/approval path is unreachable for real inputs; it only ever emits exportAudit. Yet docs/specs/06-audit-and-security.md is flipped to "Done." Fix: make sensitivity actually flow — enrich attachments with nodeId at parse time, and/or update the email skills' schemas + the agent prompts to emit export_items. Without that, the headline protection doesn't gate anything.

2. [High] The MCP destination gate is sensitivity-blind → every Workspace export needs CEO approval out of the box. src/security/export-controls.ts:163 fires on isNonContactSink && !isDestinationAllowlisted with no sensitivity condition, and the shipped default allowedDestinations is empty. So by default: create_drive_file to any folder (incl. the 'root' fallback at :340), and append_table_rows/create_sheet-to-existing-sheet, all return approval_required regardless of sensitivity. Concrete: appending one internal (or public) row to an existing team sheet is blocked pending approval. This makes the Workspace tools near-unusable by default and contradicts #201's "confidential+ data" scope. Either seed the allowlist + document the deny-all default, or gate on sensitivity (engage only at confidential+).

3. [Medium] create_sheet (new sheet) is neither gated nor audited. A new sheet has no spreadsheet_id, so extractDestinationFromInput('create_sheet', …) returns null; the gate falls back to a placeholder email destination (destination check skipped), and the post-success audit block at src/skills/execution.ts:1461 (if (rawItems.length > 0 && destination)) is false → no export.delivered emitted. So a new sheet full of confidential data exports with no destination gate and no audit trail, despite the PR listing create_sheet as audited.

4. [Low-Med] Redundant kg_nodes query per successful MCP export. src/skills/execution.ts:1456-1459 re-runs extractExportItemsFromInput + resolveItems to build the audit event, re-resolving what the gate already resolved. Thread the resolved ExportItem[] through instead of a second DB round-trip.

5. [Low] URL/file-path allowlist uses unbounded startsWith → prefix bypass (export-controls.ts:214-216). https://drive.google.com/x matches an allowlist entry against …/x-evil…; /exports matches /exports-secret. Latent (no extractor emits these destination kinds yet) but a real bypass once they exist. Match on a host/segment boundary.

6. [Low] Approval-notification array heuristic is too broad (src/autonomy/approval-notification.ts, formatFieldValue). Any array of objects where some element has a string label/sensitivity gets rendered in export-item format, so unrelated detail arrays can be misformatted.

7. [Low] Gateway approval-notification send isn't guarded. In the export-gate approval path the actionLogRepo.insert is wrapped, but the following await this.sendNotification(...) is not — if it throws, send() throws instead of returning { success: false, gated: true, actionRef }, losing the gated signal after the pending_approval row was written. (Unreachable today per #1, latent once sensitivity flows.)

Confirmed good, no change needed: evaluateExportGate precedence and tests; fail-closed→confidential on lookup failure (both sensitivity and archived_at columns verified present); humanApproved populated on ctx and re-invoked as true by approve-action, with restricted-bulk correctly staying non-bypassable (no infinite approval loop).

josephfung:
Comments on the fixes - Addressed all seven items from the Claude review. The two load-bearing gaps are fixed:

#1 — Attachment gate now sees sensitivity. Attachments accept node_id; email skills declare export_items; the coordinator prompt tells agents to tag KG-sourced attachments. extractAttachmentExportItems() merges export_items[i] with attachments[i] so confidential/restricted thresholds can actually fire. Without node_id/export_items, items still default to internal (by design — the gate can't infer sensitivity from raw bytes).

#2 — MCP destination gate scoped to confidential+. Unknown Drive/Sheets sinks only trigger CEO approval when the export contains confidential or restricted items. Internal/public Workspace operations proceed without approval on the default empty allowlist.

Also fixed: create_sheet → (new) destination + audit; gate-resolved items threaded to export.delivered (no duplicate DB query); URL/path prefix boundary checks; export_items-only notification formatting; gateway sendNotification try/catch.

30 unit tests passing; typecheck clean.

PR #1322, opened 2026-07-03
Ant Farm AF-1: audit-log timeline query (findTimeline + index)

josephfung:
## Summary

Closes #1314

Adds the Ant Farm data-layer foundation: `AuditLogRepo.findTimeline` for ordered audit-log replay windows, optional `findByEventTypes`, migration 071 indexes for task and conversation queries, and forward-population of the `task_id` column in `AuditLogger.log()`.

## Changes

- **`findTimeline`** — optional `from`/`to`/`conversationId`/`taskId`/`limit` filters; `taskId` matches `payload->>'taskId'` (not the column); returns `parentEventId`.
- **`findByEventTypes`** — optional event-type list filter using the existing `ANY($n::text[])` idiom.
- **Migration 071** — expression index on `(payload->>'taskId')` and composite `(conversation_id, timestamp)`.
- **`AuditLogger.log()`** — writes `task_id` from payload when present (additive, no backfill).

## Testing

- Integration test `tests/integration/audit-log-timeline.test.ts` (gated on `DATABASE_URL`) covers ordering, each filter, `parentEventId`, limit clamping, and `findByEventTypes`.
- `pnpm run typecheck` passes.
- Existing audit logger unit/integration tests pass.

<div><a href="https://cursor.com/agents/bc-23eb8ec6-bb59-47f3-9938-c70e8617f2e6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-23eb8ec6-bb59-47f3-9938-c70e8617f2e6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ## [Unreleased]
 
+### Added
+
+- **Ant Farm AF-1** — `AuditLogRepo.findTimeline` and timeline indexes on `payload->>'taskId'` and `(conversation_id, timestamp)` for replay queries. Closes #1314.
+
 ### Fixed
 
 - **`delegate` timeout storm** — timeouts now return structured, retry-capped failures instead of triggering duplicate re-delegations. Closes #1288.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Added
 
-- **Ant Farm AF-1** — `AuditLogRepo.findTimeline` returns paginated `TimelinePage` (`hasMore`, keyset `after` cursor) with scoped-query guard; timeline indexes on task id and `(conversation_id, timestamp)`. Closes #1314.
+- **Ant Farm AF-1** — `AuditLogRepo.findTimeline` / `findByEventTypes` return paginated `TimelinePage` (`hasMore`, keyset `after` cursor) with scoped-query guard; task filter uses `payload->>'taskId'` (index-aligned). Closes #1314.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Added
 
-- **Ant Farm AF-1** — `AuditLogRepo.findTimeline` and timeline indexes on `payload->>'taskId'` and `(conversation_id, timestamp)` for replay queries. Closes #1314.
+- **Ant Farm AF-1** — `AuditLogRepo.findTimeline` returns paginated `TimelinePage` (`hasMore`, keyset `after` cursor) with scoped-query guard; timeline indexes on task id and `(conversation_id, timestamp)`. Closes #1314.
 
 ### Fixed
 
--- a/src/audit/audit-log-repo.ts
+++ b/src/audit/audit-log-repo.ts
@@ -30,18 +30,35 @@ export interface TimelineAuditQuery {
   conversationId?: string;
   taskId?: string;
   limit?: number;
+  /** Keyset cursor — fetch rows strictly after this (timestamp, id) pair. */
+  after?: { timestamp: Date; id: string };
 }
 
-export interface EventTypesAuditQuery {
-  from?: Date;
-  to?: Date;
-  conversationId?: string;
-  taskId?: string;
-  limit?: number;
+/** Same filter shape as {@link TimelineAuditQuery}. */
+export type EventTypesAuditQuery = TimelineAuditQuery;
+
+export interface TimelinePage {
+  rows: AuditLogRow[];
+  hasMore: boolean;
+  /** Set when {@link TimelinePage.hasMore} is true — pass as `after` for the next page. */
+  nextCursor?: { timestamp: Date; id: string };
 }
 
 const DEFAULT_LIMIT = 100;
 const MAX_LIMIT = 500;
+
+function assertTimelineScope(query: TimelineAuditQuery): void {
+  if (!query.from && !query.to && !query.conversationId && !query.taskId) {
+    throw new Error(
+      'findTimeline requires at least one scope filter: from, to, conversationId, or taskId',
+    );
+  }
+}
+
+/** Match task id on the indexed column when populated, else payload (historical rows). */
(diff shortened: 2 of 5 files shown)


josephfung:
## Code review — AF-1 audit-log timeline query

Reviewed with a security-first lens, keeping in mind this is the Ant Farm data-layer foundation (a nice-to-have visualization). **Bottom line: clean, well-scoped, low-risk. No security issues and no correctness bugs found.** One item is worth deciding before the replay UI lands; the rest are optional.

### Security ✅
- All three queries are fully parameterized (`$n` placeholders + params array). The dynamic `WHERE` builder only ever concatenates fixed condition strings, never user input, so there's no injection surface. `event_type = ANY($1::text[])` and `payload->>'taskId' = $n` are both bound params.
- `audit_log.task_id` is `TEXT` with no foreign key (migration 001), so populating it from an arbitrary `payload.taskId` string in `AuditLogger.log()` can't raise an FK violation and take down audit writes. Confirmed safe.
- Migration 071 is pure DDL (`CREATE INDEX`), so it doesn't trip the append-only UPDATE trigger (migration 021) that has crash-looped boot on `audit_log` DML before. Down migration drops both indexes in reverse order. Good.
- The new 9-column INSERT lines up correctly: column list `(id, timestamp, event_type, source_layer, source_id, payload, conversation_id, task_id, parent_event_id)` matches the values array `$1`–`$9`, with `taskId` in the `task_id` slot.

### Worth deciding (not blocking)

1. **Silent truncation / no pagination in `findTimeline`.** The query is `ORDER BY timestamp ASC LIMIT $n` (default 100, max 500) with no cursor/offset and no `hasMore` signal. On a busy replay window with more than 500 events, the timeline is silently cut to the *earliest* 500 rows and the caller has no way to know it's incomplete, so the viz would quietly drop events. Since this is the foundation the replay/DVR UI sits on, worth deciding now whether the UI paginates (keyset on `(timestamp, id)`) or explicitly accepts a capped window. Related: an unfiltered `findTimeline({})` returns the oldest 100 rows in the whole table, which is probably never what a replay caller wants.

### Minor (optional cleanup)

2. `EventTypesAuditQuery` is an exact structural duplicate of `TimelineAuditQuery`. `type EventTypesAuditQuery = TimelineAuditQuery` (or a shared base) keeps the two from drifting apart.
3. `idx_audit_log_conversation_timestamp (conversation_id, timestamp)` overlaps the existing `idx_audit_conversation (conversation_id)` — the composite can also serve conversation-only lookups, so the older single-column index is now largely redundant. Fine to leave; just noting.
4. `AuditLogger.log()` now writes the `task_id` column, but `findTimeline`/`findByEventTypes` (and the new index) read `payload->>'taskId'`, so on this path the column is written but never read. Intentional per the PR description (covers historical NULL rows), just flagging that task id now lives in two places.
5. Boot note: non-concurrent `CREATE INDEX` takes an ACCESS EXCLUSIVE lock while it builds, so on a large prod `audit_log` this lengthens startup. Not an outage (single-process, runs pre-traffic, and `CONCURRENTLY` can't run inside node-pg-migrate's per-migration transaction anyway) — just expect boot to block for the build.

`findTimeline`/`findByEventTypes` have no callers yet (foundation only, exercised by the integration test) — expected for AF-1. Nice clean pattern-match to the existing `findSkillResults` shape.

josephfung:
## Follow-up on the AF-1 revisions

The pagination rework is correct: `limit + 1` to detect truncation, `ORDER BY timestamp ASC, id ASC` with a matching `(timestamp, id) > (…)` keyset predicate, and `nextCursor` set only when `hasMore`. No dupes or gaps across pages for millisecond-aligned timestamps, and the `assertTimelineScope` guard is a good add. Three notes from a closer pass, in priority order.

### 1. `COALESCE(task_id, payload->>'taskId')` bypasses the index this migration adds (worth fixing)

The task filter now uses `COALESCE(task_id, payload->>'taskId') = $n`. Postgres only uses an expression index when the query contains that exact expression, so `COALESCE(...)` can't use `idx_audit_log_payload_task_id` (indexed on `(payload->>'taskId')`), and there's no index on the bare `task_id` column. Net: a task-scoped query seq-scans `audit_log` (unless a `from`/`to` window lets it lean on the timestamp index), and the index this migration exists to provide is unreachable.

It's also redundant. `AuditLogger.log()` is the only writer of `audit_log`, and it sets `task_id` *from* `payload.taskId` while storing the full payload — so `payload->>'taskId'` is non-null exactly when `task_id` is, with the same value. The `COALESCE` covers zero additional rows.

Suggested: revert the predicate to `payload->>'taskId' = $n` (keeps the index, loses no coverage). If the plan is to make `task_id` the canonical column long-term, then add a plain index on `task_id` and query the column directly instead. Either way, the query and the index should agree.

### 2. Keyset cursor loses sub-millisecond precision (latent, worth a guard)

The cursor carries `timestamp` as a JS `Date` (millisecond precision), but the column is `TIMESTAMPTZ` (microsecond) with a `DEFAULT now()`. Today every writer inserts a `new Date()` (ms-aligned), so this never triggers. But any row written with sub-ms precision (the `now()` default if some future INSERT omits `timestamp`, or a raw-SQL / bulk import) could be duplicated or skipped at a page boundary, because the cursor can't represent the exact stored value. A short comment recording the "all timestamps are ms-aligned" assumption, or truncating to ms on write, keeps this from turning into a heisenbug later.

### 3. `findByEventTypes` ignores `after` and can still silently truncate (minor)

`EventTypesAuditQuery` now aliases `TimelineAuditQuery`, so it carries the `after` cursor field, but `findByEventTypes` doesn't read it and returns a bare array with no `hasMore`. So it (a) exposes a cursor knob that does nothing and (b) keeps the original silent-truncation-at-cap behavior that `findTimeline` just fixed. Fine if it's only ever used for small bounded lookups; otherwise it wants the same `TimelinePage` treatment.

None of these block merge for a nice-to-have viz. #1 is the one I'd change before the table grows, since it quietly undoes the migration's own reason for existing.

PR #1326, opened 2026-07-03
fix: calendar operates as the principal — bind the calendar client to the CEO grant (#1217)

### LINKED ISSUE #1217: calendar-respond-to-invite RSVP fails with Google 400 omittedAttendeesSpecified (and skill disabled at boot)
## Summary

Accepting calendar invites fails. The principal has a standing instruction to auto-accept invites from a specific sender; two such invites on 2026-06-27 were never accepted. There are two layers:

1. **Skill not loaded at boot.** `calendar-respond-to-invite` (and `calendar-create-hold`, `calendar-holds-sweep`) were enabled in the registry ~8 minutes *after* the container booted. Enabling a skill in the DB does not hot-reload a running process, so the running container did not have the tool. The first invite failed with *"Skill 'calendar-respond-to-invite' not found in registry."*
2. **RSVP API 400.** After the next restart loaded the skill, the second invite reached `sendRsvp` and Google returned:

```
400 — "The request specified attendees that should have been omitted."
  domain: calendar, reason: omittedAttendeesSpecified
```

So even with the skill loaded, invite acceptance is broken.

## Production evidence

- Invite #1 (11:46 UTC): `calendar-respond-to-invite` → "Skill not found in registry" → escalated, not accepted.
- Restart at ~11:51 UTC loaded the now-enabled skill.
- Invite #2 (12:17 UTC): `sendRsvp` → Google 400 `omittedAttendeesSpecified` (two attempts) → escalated, not accepted.

## Root cause

- `skills/calendar-respond-to-invite/handler.ts:88` → `NylasCalendarClient.sendRsvp` (`src/channels/calendar/nylas-calendar-client.ts:378`). The RSVP request shape is rejected by Google for a non-organizer calendar — the underlying call is sending attendee data that must be omitted on an RSVP. Needs the correct Nylas v8 RSVP semantics.
- Registry enrollment: these calendar skills are not picked up on a fresh boot until manually enabled + restarted. They should be in `config/registry-defaults.yaml` so a fresh deploy enrolls them (closing the deploy → enable → restart gap).

## Acceptance criteria

- [ ] `sendRsvp` accepts an invite on the principal's calendar without a Google 400 (correct RSVP request shape — no omitted-attendee data).
- [ ] The RSVP path is verified against a real invite (manual or integration test) for accept / decline / tentative.
- [ ] A unit test pins the corrected RSVP request shape (would fail against the old shape).
- [ ] `calendar-respond-to-invite`, `calendar-create-hold`, and `calendar-holds-sweep` are added to `config/registry-defaults.yaml` so a fresh boot enrolls them (no silent disabled state), OR the enable+restart requirement is documented if intentional.
- [ ] `pnpm run typecheck` and the affected tests pass.
- [ ] `version` bumped in `skills/calendar-respond-to-invite/skill.json`.

josephfung:
## Summary

Fixes the Google 400 `omittedAttendeesSpecified` when accepting calendar invites. The root cause is **identity**, not request shape: the calendar client was constructed at boot bound to the primary email account's Nylas grant (`resolvedEmailAccounts[0].nylasGrantId` = Curia's own mailbox), so Curia acted as a third-party delegate on the CEO's calendar. RSVP is first-person, so Google rejected it.

This binds the single boot-constructed `NylasCalendarClient` to the CEO's own grant (`ceo_nylas_grant_id`) — the same identity the CEO's inbox uses — via a new helper, and **fails closed** when no CEO grant is configured.

Closes #1217.

## Approach (decisions made with the issue owner)

- **Minimal re-source, not per-skill secrets.** Kept the existing `nylasCalendarClient` capability injection; changed only the grant the boot client binds to. The calendar *identity* is unified onto the CEO grant without touching the 10 calendar handlers/manifests.
- **Fail closed, no email fallback.** No `?? email[0]` fallback — that fallback is exactly the delegate-identity bug (a two-identity deployment that forgot `ceo_nylas_grant_id` would silently run calendar as Curia's mailbox again). No CEO grant configured means calendar is disabled and its skills return a clean "not configured" result. Single-identity/self-host deployments must set `ceo_nylas_grant_id`.
- **`config/registry-defaults.yaml` deliberately NOT modified.** The pinned issue comment is correct: adding `calendar-*` there contradicts the credential-gating design, is a no-op for existing prod (registry rows already exist and reconciliation never overrides them), and is inert on a fresh install (the calendar agent stays excluded, so nothing pins these skills). The real prod gap ("enable in DB then restart") is operational.

## Changes

- **`src/channels/calendar/resolve-calendar-grant.ts`** (new, unit-tested) — resolves `ceo_nylas_grant_id`; returns `undefined` on genuine absence, and **propagates** DB/decrypt read failures so they surface loudly rather than masquerading as "not set".
- **`src/index.ts`** — boot binds the calendar client to the resolved CEO grant. A read failure is caught and logged at `error` (a wrong `SECRET_ENCRYPTION_KEY` breaks every secret, not just calendar), while the "not configured" warning fires only on genuine absence. `primaryNylasClient` stays (still used by HealthService).
- **`src/channels/calendar/nylas-calendar-client.test.ts`** — pins the `sendRsvp` body to `{ status }` only (fails if attendee data leaks in).
- **`skills/calendar-respond-to-invite/{skill.json,handler.ts}`** — corrected identity wording to describe the CEO grant; version 0.1.1 → 0.1.2.
- **`agents/calendar.yaml`** — collapsed the "my vs your calendar" disambiguation into a single CEO-calendar identity; version 0.6.0 → 0.6.1.
- **`CHANGELOG.md`**, design memo (corrected to fail-closed + cutover runbook), and implementation plan under `docs/wip/`.

## Testing

- `pnpm run typecheck` clean (src + skills + tests).
- 170 tests pass across touched areas: 49 in `src/channels/calendar/` (resolver + client) and 121 across all `calendar-*` skills.
- Ran code-review, silent-failure, and security reviews in parallel. Security: clean (grant value never logged; fail-closed confirmed the secure choice). Silent-failure: one HIGH — the original grant-read `catch` swallowed a decrypt failure that `SecretsService` intentionally makes loud; fixed here (propagated + logged at `error`, "not configured" reserved for genuine absence).

## Follow-ups / handoff (not in this PR)

- **Live RSVP verification (issue AC2):** accept/decline/tentative against a real invite under the CEO grant must be verified post-merge + cutover (needs the live grant). Code is fallback-ready: if the SDK `sendRsvp` misbehaves under the CEO grant, switch to a raw `fetch` mirroring `CeoNylasClient`.
- **Cutover:** the CEO's `contact_calendars` entry must be re-registered under the new grant (calendar IDs are grant-scoped). Runbook is in the design memo.
- The 10 calendar skill guards say "Nylas credentials missing" — now slightly misleading (the common cause is an unset CEO grant); left untouched to honor the minimal-change decision.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Fixed
 
+- **Calendar operates as the CEO** — the calendar client now binds to `ceo_nylas_grant_id` (the same identity as the CEO's inbox), fixing RSVP failures (`omittedAttendeesSpecified`) caused by acting as a third-party delegate; fails closed when no CEO grant is configured. Closes #1217.
 - **`delegate` timeout storm** — timeouts now return structured, retry-capped failures instead of triggering duplicate re-delegations. Closes #1288.
 
 ### Security
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Fixed
 
-- **Calendar operates as the CEO** — the calendar client now binds to `ceo_nylas_grant_id` (the same identity as the CEO's inbox), fixing RSVP failures (`omittedAttendeesSpecified`) caused by acting as a third-party delegate; fails closed when no CEO grant is configured. Closes #1217.
+- **Calendar operates as the CEO** — binds the calendar client to `ceo_nylas_grant_id`, fixing RSVP `omittedAttendeesSpecified`; fails closed when unset. Closes #1217.
 - **`delegate` timeout storm** — timeouts now return structured, retry-capped failures instead of triggering duplicate re-delegations. Closes #1288.
 
 ### Security
--- a/agents/calendar.yaml
+++ b/agents/calendar.yaml
@@ -1,5 +1,5 @@
 name: calendar
-version: "0.6.0"
+version: "0.6.1"
 role: specialist
 description: >
   Calendar domain specialist — scheduling, free/busy, conflict resolution,
@@ -18,9 +18,11 @@ system_prompt: |
   to the coordinator for synthesis and delivery.
 
   ## Entity Resolution
-  Your contact ID is ${agent_contact_id}. Use this when "you" or "your" clearly refers
-  to you as the agent (e.g. "what's your calendar look like?"). This is the ONLY contact
-  ID you should EVER use directly — NEVER derive or guess contact IDs from names.
+  Your contact ID is ${agent_contact_id}. This is the ONLY contact ID you should EVER
+  use directly for your own agent identity — NEVER derive or guess contact IDs from
+  names. You do NOT operate a calendar of your own: the calendar client is bound to the
+  CEO's Nylas grant, so every calendar operation acts on the CEO's calendar (see
+  "Calendar Identity" below).
 
   When the coordinator passes `<resolved_entities>` from a prior contacts briefing,
   use those IDs directly — do not re-resolve names that already appear in the block.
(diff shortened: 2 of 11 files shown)


josephfung:
@coderabbitai review

josephfung:
Addressed the CI failure and CodeRabbit's nitpicks:

- **CI:** `calendar-prompt.test.ts` pinned the agent version to `0.6.0`; updated the tripwire to `0.6.1` to match this PR's patch bump (with a comment to bump it alongside the YAML next time).
- **Warning tense:** `ran` → `runs` — it's pushed before `sendRsvp` executes.
- **CHANGELOG:** trimmed the bullet to the one-clause house style.
- **index.ts silent path:** the calendar bootstrap no longer has a silent branch when `NYLAS_API_KEY` is unset — it logs at `debug` (the missing key is already warned by the email bootstrap) and guards on the key first, so we don't do a pointless grant read when Nylas isn't configured at all.

PR #1340, opened 2026-07-04
feat(antfarm): AF-7 real LimeZu art loader + safe serving

josephfung:
## Summary

Ant Farm AF-7 (#1335): render real licensed LimeZu pixel art in the Phaser office when present, served only behind session auth, with a clean fall back to procedural placeholders when absent.

Closes #1335

### Safe serving (Blocker 2)
- New `GET /api/antfarm/assets/*` route (`antfarm-assets.ts`) guarded by the existing `assertSecret` session check (same as `/api/antfarm/timeline`). Serves from `apps/antfarm/assets-licensed/` — **outside** the world-downloadable Vite `public/` root, so licensed bytes never hit the unauthenticated static surface.
- Path-traversal blocked; missing file/dir → 404 (open-core → placeholders). Non-ENOENT fs errors re-throw (surface, not swallow). Integration test covers 401 / 200 / 404 / traversal.

### Loader + fallback (Blocker 1)
- `OfficeScene.preload()` loads office singles, tileset/room-builder, and per-agent character sheets; a `loaderror` handler records absent assets (info-level, no console errors) instead of throwing.
- A single all-or-nothing `realArtReady` gate renders real art only when **every** licensed asset loaded; a partial/mispackaged drop degrades to the full procedural scene (no broken hybrid).

### Characters
- Deterministic 1-of-20 premade sheet per agent (same hash as the placeholder appearance map). Real 32×64 frames, idle + 4-direction walk animations.

### Office (art-directed)
- Desks composed from Modern Office **singles** (lower-left 32×48 of 3 color tiles), deterministic per-agent color; grey reserved for the coordinator (wide 4-tile desk + 3 monitors). Specialists get a computer (122) + a chair (random 329–336); coordinator chair 101.
- Explicit depth bands (chair < agent < desk < devices < overlays) so the desk front occludes the seated character.
- Floor + tasks-board still sourced from the tileset/room-builder. Claw, vacuum tubes, and scheduler stay procedural (no pack coverage).

### Deploy
- Curated runtime-PNG-only art (character sheets + office singles + tileset + room-builder) is staged in **curia-deploy** and layered into the auth-served dir at image build. See josephfung/curia-deploy PR (branch `feat/antfarm-real-art-1335`) — must ship together, since this route expects that dir.

## Testing
- `pnpm run typecheck`, `pnpm --filter @curia/antfarm run build`, and the antfarm + `antfarm-assets` integration tests all pass.
- Frame-slicing, walk directions, and the office composition were validated in a standalone Phaser harness.

## Follow-up (not blocking merge)
- **Live-scene pixel pass:** the composition offsets/spacing were tuned in a standalone harness; exact in-scene desk spacing (vs the 800px stage), seated offsets, and the walk/seat interaction want a light refinement against the running app.
- Left/right walk frames (62/68) are a 3/4-view best-guess (up/down are exact) — trivially swappable if they read wrong.

## Out of scope
- Generator-parts character compositing (deferred post-v1; layer library not staged).
- Custom robot sprites (deferred post-v1).


josephfung:
@coderabbitai review

josephfung:
@coderabbitai review

apps/antfarm/src/game/OfficeScene.ts
```diff
@@ -39,26 +82,88 @@ export class OfficeScene extends Phaser.Scene {
...
+    // Character sheets are loaded here too (added in Task 5).
+    const desks = this.registry.get('desks') as DeskSlot[] ?? [];
+    const sheetIndices = new Set(desks.map((d) => characterSheetIndexForAgent(d.agentId)));
+    for (const idx of sheetIndices) {
+      this.load.spritesheet(characterSheetKey(idx), characterSheetUrl(characterSheetFile(idx)), {
+        frameWidth: CHARACTER_FRAME.width,
+        frameHeight: CHARACTER_FRAME.height,
+      });
+    }
   }
```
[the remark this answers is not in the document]
josephfung: Fixed in e7efaf2b. `preload()` is now restart-safe:

- The `loaderror` listener is cleared with `this.load.off('loaderror')` before re-adding, so restarts (`updateLayout`/`resyncPlayback` call `scene.restart()`, and the `LoaderPlugin` instance persists across that) no longer stack duplicate handlers.
- Each `this.load.image`/`this.load.spritesheet` is now guarded by `!this.textures.exists(key)`, so already-loaded tileset, singles, and character-sheet keys are not re-queued.

Note I intentionally did not gate on `this.failedLoads`: keys that legitimately failed (absent licensed art in open-core) don't exist as textures, so they're re-attempted on restart — cheap, and it preserves recovery if art ever becomes available mid-session.

src/channels/http/routes/antfarm-assets.ts
```diff
@@ -0,0 +1,62 @@
...
+    try {
+      const s = await stat(absPath);
+      if (!s.isFile()) return reply.status(404).send({ error: 'Not found' });
+    } catch (err) {
+      // Missing file OR missing assets-licensed dir (open-core) → 404 → placeholders.
+      if ((err as NodeJS.ErrnoException).code === 'ENOENT') {
+        return reply.status(404).send({ error: 'Not found' });
+      }
+      throw err;
+    }
```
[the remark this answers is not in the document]
josephfung: Fixed in e7efaf2b. Non-ENOENT errors now log with context before propagating:

```ts
req.log.error({ err, absPath }, 'antfarm-assets: unexpected fs error serving licensed asset');
throw err;
```

On the `AgentError` normalization suggestion: I deliberately kept the bare `throw` rather than wrapping in `AgentError` here. `AgentError` is the normalization shape for the agent/skill/bus layers (`docs/specs/05-error-recovery.md`); this is a Fastify HTTP route, where the framework's error handler owns turning a thrown error into a 500. Wrapping it in `AgentError` would be a layer mismatch (no bus context, no skill result envelope) and none of the sibling HTTP routes do it. The guideline's intent — never swallow, always log-and-propagate — is now satisfied.

apps/antfarm/src/game/OfficeScene.ts
```diff
@@ -124,35 +238,264 @@ export class OfficeScene extends Phaser.Scene {
...
+  /** Overwrite an existing texture key with a cropped region of a loaded source image.
+   *  Call-sites that reference `newKey` (e.g. this.add.image(x,y,'desk')) then get real art
+   *  with zero changes. Uses a canvas texture so the crop is a standalone key. */
+  private cropToTexture(srcKey: string, newKey: string, sx: number, sy: number, sw: number, sh: number): void {
+    const src = this.textures.get(srcKey).getSourceImage() as HTMLImageElement | HTMLCanvasElement;
+    const canvas = document.createElement('canvas');
+    canvas.width = sw;
+    canvas.height = sh;
+    const ctx = canvas.getContext('2d');
+    if (!ctx) throw new Error('[antfarm] cropToTexture: no 2d canvas context'); // swapOfficeTextures keeps the placeholder
```
josephfung, line 403: [CodeRabbit nitpick — cropToTexture null-context] Fixed in e7efaf2b. Replaced `getContext('2d')!` with an explicit null check that throws a descriptive error, matching the sibling `composeDeskTexture`/`cropSingle` helpers. `swapOfficeTextures` still catches it and keeps the placeholder, so behaviour is unchanged — just consistent and honest about the failure mode.

apps/antfarm/src/game/world-layout.ts
```diff
@@ -1,8 +1,20 @@
...
+export const STAGE_HEIGHT = 960;
+
+// --- Specialist ("floor") desk layout ---------------------------------------
+// Specialists are split across two rows so they neither crowd each other nor
+// collide with the coordinator's desk above. A composed desk renders ~192px wide
+// on the world canvas (3 tiles × 32px × the ×2 sprite scale), and a full station
+// (monitors above, chair/label below) is ~180px tall — so spacing must clear the
+// width and the rows must sit well below the coordinator (boss desk at y:200,
+// see buildWorldLayout). Tune freely.
+const FLOOR_ROW1_Y = 340;        // first specialist row — clears the coordinator's station (~y:200 + ~90 below)
```
josephfung, line 14: [CodeRabbit nitpick — comment drift on coordinator y] Fixed in e7efaf2b. The comments now reference the actual boss-desk position (`y:200`, set in `buildWorldLayout`) instead of the stale `y:110`/`y~160`. Both `FLOOR_ROW1_Y` and the block comment now point at the same number.

apps/antfarm/src/game/world-layout.ts
```diff
@@ -32,19 +44,32 @@ export function buildWorldLayout(desks: DeskSlot[]): WorldLayout {
...
+    const rowWidth = spacing * agents.length;
+    const startX = (STAGE_WIDTH - rowWidth) / 2; // center the row on the stage
+    agents.forEach((desk, index) => {
+      deskPositions.push({
+        agentId: desk.agentId,
+        row: 'floor',
+        x: startX + spacing * (index + 0.5),
+        y,
+      });
     });
```
josephfung, line 71: [CodeRabbit nitpick — no lower bound on spacing] Acknowledged, leaving as-is intentionally.

`spacing` is `min(FLOOR_MAX_SPACING, avail / n)` with no floor, so a very large single-row roster could pack desks tighter than ~192px and overlap. But:

- The realistic roster is a handful of specialists (coordinator + a few skill agents), split across two rows — nowhere near the ~7-per-row where 240px spacing starts compressing.
- Adding a minimum-spacing floor without wrap logic would just push desks off the stage edges instead (overflow rather than overlap) — no better, and the two-row split already halves the per-row count.
- Real wrap-to-N-rows is a layout feature beyond this PR's scope (real-art swap for AF-7).

If the roster ever grows enough to need it, the fix is dynamic row count (`rows = ceil(n / perRow)`), which I'd rather do deliberately than bolt on a silent clamp now. Flagging as a known ceiling, not fixing.

src/channels/http/routes/antfarm-assets.ts
```diff
@@ -0,0 +1,77 @@
...
+        return reply.status(404).send({ error: 'Not found' });
+      }
+      // Anything else (EACCES, EMFILE, …) is a real fault, not the open-core absent-art path:
+      // log with context before propagating so Fastify's error handler can turn it into a 500
+      // that's actually diagnosable on-call, instead of an anonymous stack trace.
+      req.log.error({ err, absPath }, 'antfarm-assets: unexpected fs error serving licensed asset');
+      throw err;
+    }
+
+    // Strict allowlist — only known-safe types are served. Unknown extensions 404 rather than
```
josephfung, line 64: Addressed in cd8a60a7.

Context first: this route is already session-authed (`assertSecret`), path-traversal-guarded (`relative(...).startsWith('..')`), and serves only build-shipped licensed art from a directory outside the web root — there are no user-uploaded files and no HTML in it, and the Content-Type was never `text/html`. So the practical XSS exposure was low. But the finding points at a real hardening gap on the response side, so I closed it two ways (defense-in-depth):

1. **Strict content-type allowlist.** Unknown extensions now 404 instead of falling back to `application/octet-stream`. The body can only ever be labelled `image/png` or `application/json` — never `text/html`, and never a guessable stream. A `.html` file placed on disk is refused, not served.
2. **`X-Content-Type-Options: nosniff`** on every served response, so a browser can't sniff a PNG's bytes into an executable type regardless of the declared Content-Type. This is the direct mitigation for `direct-response-write`.

`resp.render()` (the rule's suggested fix) doesn't apply — this streams binary image bytes, not templated HTML.

Tests added: assert `nosniff` on a served PNG, and that a `.html` file present on disk 404s (strict allowlist).

PR #1349, opened 2026-07-07
feat: add AWS Bedrock as a configured LLM provider

AnshulaChowdhury:
Summary
     
  Adds AWS Bedrock as a third LLMProvider (alongside Anthropic and OpenRouter), via
  Bedrock's Converse API. This lets a deployment run entirely on AWS Bedrock instead
  of, or alongside, the direct Anthropic/OpenAI APIs — useful for operators who
  want to consolidate billing/infra under an existing AWS account.

  The model actually shipped for the standard/powerful tiers
  (anthropic.claude-3-sonnet-20240229-v1:0, served via Bedrock) is the result of a
  live model search, not a single choice — the first three candidates tried each
  failed in a different, instructive way, documented in full in
  docs/adr/023-aws-bedrock-mistral-llm-provider.md:
  
  1. An invalid model ID → rejected by Bedrock outright.
  2. A working Mistral Large model → passed every isolated test, but silently 
  narrated tool calls as text instead of invoking them once the coordinator's real
  system prompt (~11.8K tokens) and full tool count (~48 skills) were both in play —
  confirmed 5/5 failure under that load, and once in a real production call where
  the coordinator told the CEO it would check something and never actually did. This
  is the finding I'd most want a reviewer to sanity-check, since it's a
  silent-failure mode, not a crash.
  3. A newer Claude model on Bedrock → real tool-calling confirmed once, but gated
  behind an AWS Marketplace subscription permission beyond plain
  bedrock:InvokeModel, unavailable on this account.
  4. anthropic.claude-3-sonnet-20240229-v1:0 (shipped) — an older, native on-demand
  model with no Marketplace gate. Confirmed 6/6 across isolated + full
  production-shaped load — 100% real tool_use, zero narrated-call failures.

  Two follow-on issues were found and fixed along the way (both documented in the
  ADR):
  - The narrated-tool-call failure mode is now detected and logged loudly
  (detectUncalledToolIntent()), and gated by a new permanent live regression check:
  pnpm run smoke:bedrock-tools. 
  - A real production conversation later broke every subsequent turn with
  ValidationException: A conversation must start with a user message — traced to two
  pre-existing bugs in AgentRuntime/WorkingMemory (a failed LLM call doesn't
  persist an assistant-side turn; a mid-history summarization turn isn't hoisted
  into the system prompt before provider-level system-message extraction). Bedrock's
  Converse API is strict about message ordering in a way Anthropic/OpenAI
  apparently aren't, so it's the first provider to surface this. Fixed defensively
  at the Bedrock adapter boundary (normalizeMessageSequence()); the actual root
  cause in AgentRuntime/WorkingMemory is not fixed here — flagged explicitly in the
  ADR's Consequences as follow-on work, since it's out of scope for a provider-swap
  PR and affects Anthropic/OpenRouter too under the right conditions.

  Changes
  
  - src/agents/llm/bedrock-mistral.ts (new) — BedrockMistralProvider, implementing
  LLMProvider via Bedrock's Converse API. Despite the filename (predates the final
  model choice), the class isn't Mistral-specific — it works for any
  Converse-compatible Bedrock model. Includes detectUncalledToolIntent() (logs when
  a model narrates rather than invokes a tool call) and normalizeMessageSequence()
  (defensive fix for the malformed-history case above).
  - src/agents/llm/bedrock-mistral.test.ts (new) — 18 unit tests, mocking
  BedrockRuntimeClient (keeping the real ConverseCommand), mirroring the existing
  anthropic.test.ts/openrouter.test.ts pattern. Includes tests reproducing the exact
  production sequences that broke tool-calling and conversation history.
  - scripts/smoke-bedrock-tool-use.ts (new) + pnpm run smoke:bedrock-tools — live
  regression gate against real Bedrock, reproducing the production-shaped load
  (large system prompt + ~48 tools) that exposed the narrated-tool-call failure.
  Deliberately not part of pnpm test (needs live AWS credentials + network calls) —
  same reasoning as tests/integration needing Docker Postgres, or redteam needing
  its own setup.
  - src/index.ts — Anthropic, OpenRouter, and Bedrock are now registered
  symmetrically, each conditional on its own credentials. The previous unconditional
  ANTHROPIC_API_KEY boot requirement is gone; the existing "every tier-mapped model
  has a registered provider" validation loop is what actually enforces a deployment
  has what it needs.
  - src/config.ts, src/secrets/apply-vault-secrets.ts (+ test) — AWS credentials
  resolve from the encrypted secrets vault
  (aws_access_key_id/aws_secret_access_key), not .env, per ADR-021 (vault-only
  secret resolution) — consistent with every other provider key. Region/timeout are
  non-secret operational config and do live in .env.
  - scripts/seed-vault.ts — AWS keys added to the seed list; REQUIRED_SECRET_NAMES
  no longer hardcodes anthropic_api_key as universally required, since no single
  provider is unconditionally required anymore.
  - src/agents/llm/model-registry.ts — new entries for
  mistral.mixtral-8x7b-instruct-v0:1, mistral.mistral-large-2402-v1:0, and
  anthropic.claude-3-sonnet-20240229-v1:0.
  - config/default.yaml — model_routing.tiers: fast → Mixtral, standard/powerful →
  Claude 3 Sonnet via Bedrock. filter.llmJudge/escalation.judge stay pinned to
  Mixtral — since that's a genuinely different model family from the new Claude
  tiers, the model-diversity property those judges are designed around (different
  vendor than the coordinator, so a prompt injection that fools one shouldn't fool
  both) is intact, not degraded.
  - .dockerignore — unrelated fix hit while rebuilding locally: excludes macOS
  AppleDouble (._*) sidecar files, which were failing the Docker build context
  loader outright.
  - docs/adr/023-aws-bedrock-mistral-llm-provider.md (new), docs/adr/README.md,
  docs/specs/02-agent-system.md, CHANGELOG.md, .env.example,
  package.json/pnpm-lock.yaml (@aws-sdk/client-bedrock-runtime, @smithy/types) —
  supporting docs and dependencies.

  Related Issues

  None filed — happy to open one if maintainers prefer that workflow before merging.

  CONTRIBUTING.md checklist

  - [x] Branch naming — feat/aws-bedrock-llm-provider
  - [x] Commit message — Conventional Commits (feat: ...)
  - [x] Code standards — TypeScript ESM throughout; no any introduced; no empty
  catch {} (every catch classifies via classifyError() and returns a typed
  AgentError, matching anthropic.ts/openrouter.ts); pino only, zero console.*; no
  raw SQL in this change
  - [x] Testing — 18 new unit tests + 2 new tests on apply-vault-secrets.test.ts;
  full suite run repeatedly during development (3815/3822 passing, rest skip on
  missing DATABASE_URL same as before this change); pnpm run typecheck, pnpm run 
  lint, pnpm run build all clean
  - [x] ADR — docs/adr/023-aws-bedrock-mistral-llm-provider.md, included in this PR
  (not filed separately), indexed in docs/adr/README.md
  - [x] Spec docs updated — docs/specs/02-agent-system.md now lists the Bedrock
  provider alongside Anthropic/OpenRouter (was previously only mentioning the other
  two)
  - [x] CHANGELOG.md — [Unreleased] entries under Added and Fixed
  - [ ] CI passes — pending, will confirm once this PR is open
  - [x] AI-assisted contribution disclosure (per CONTRIBUTING.md's AI-Assisted
  Contributions section): this PR is predominantly AI-generated, built with Claude
  Code (Anthropic) across an extended session that included live debugging against a
  real AWS account — the model search above, the tool-narration bug, and the
  conversation-history bug were all found through actual failures against live
  infrastructure, not synthesized. I've reviewed the full diff and can speak to any
  of the design choices above.

  One thing worth a maintainer opinion

  The out-of-scope AgentRuntime/WorkingMemory bug flagged in the ADR (failed LLM
  calls not persisting an assistant turn; mid-history system turns not hoisted) is
  real and pre-existing — I chose not to fix it here since it's a different layer
  than a provider swap and affects Anthropic/OpenRouter too, just less visibly since
  their APIs are more lenient about message ordering. Happy to file a follow-up
  issue for it, or take a swing at it in a separate PR if that's preferred.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,14 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ## [Unreleased]
 
+### Added
+
+- **AWS Bedrock LLM provider** — `BedrockMistralProvider` adds a third `LLMProvider` (alongside Anthropic and OpenRouter) via Bedrock's Converse API. Fast tier runs Mixtral 8x7B; standard/powerful run Claude 3 Sonnet via Bedrock, chosen after live load testing showed a candidate Mistral Large model silently narrating tool calls as text instead of invoking them under real prompt/tool-count load. Anthropic is no longer unconditionally required at boot — provider requirements now follow from which models `config/default.yaml`'s `model_routing` tiers actually reference. AWS credentials resolve from the secrets vault (ADR-021), not `.env`. A new `pnpm run smoke:bedrock-tools` script guards against this failure mode regressing. (ADR-023)
+
+### Fixed
+
+- **Bedrock: malformed conversation history crashed every subsequent turn** — a conversation that had accumulated orphaned turns (from earlier failed LLM calls not persisting an assistant reply) or a mid-history summarization turn could violate Bedrock Converse's strict "starts with user, strictly alternates" requirement, throwing `ValidationException` on every message. `BedrockMistralProvider` now normalizes the sequence (drops leading non-user messages, merges consecutive same-role turns) before sending it. The underlying gap in `AgentRuntime`/`WorkingMemory` — not persisting an assistant turn after a failed call, and not hoisting mid-history system turns — still needs its own fix; this is a defensive mitigation in the Bedrock adapter only. (ADR-023)
+
 ### Changed
 
 - **Coordinator Drive moves** — pinned `update_drive_file` to the coordinator and documented reparenting (`add_parents`/`remove_parents`), so "move this doc into a folder" requests are performed instead of declined. (#1062)
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -5,11 +5,15 @@
 model_routing:
   tiers:
     fast:
-      model: claude-haiku-4-5
+      model: mistral.mixtral-8x7b-instruct-v0:1
+    # standard/powerful: mistral-large-2402 silently narrates tool calls as text
+    # instead of invoking them under the coordinator's real prompt/tool-count
+    # load (see docs/adr/023) — Claude 3 Sonnet via Bedrock confirmed 6/6 real
+    # tool_use under the same load and needs no extra IAM/Marketplace grant.
     standard:
-      model: claude-sonnet-4-6
+      model: anthropic.claude-3-sonnet-20240229-v1:0
     powerful:
-      model: claude-opus-4-6
+      model: anthropic.claude-3-sonnet-20240229-v1:0
   default_tier: standard
 
 channels:
@@ -291,14 +295,16 @@ intentDrift:
 filter:
   llmJudge:
     enabled: true
-    # Default works out of the box (claude-haiku-4-5 — smaller/cheaper than the opus coordinator).
-    #
     # >>> STRONGLY RECOMMENDED <<<: point this at a DIFFERENT vendor/family than the
     # fast/standard/powerful agent tiers above (e.g. an OpenRouter Gemini or DeepSeek
     # model such as 'google/gemini-3.1-flash-lite'). Model diversity is the security
-    # value here: an attack crafted to fool the Claude coordinator should not also fool
+    # value here: an attack crafted to fool the coordinator should not also fool
     # the reviewer. Requires the corresponding provider API key (e.g. OPENROUTER_API_KEY).
-    model: claude-haiku-4-5
+    #
+    # Pinned to Mixtral (fast tier) — a genuinely different model family from the
+    # standard/powerful tier's Claude 3 Sonnet, so this still satisfies the
+    # diversity property above despite both running on the same Bedrock account.
+    model: mistral.mixtral-8x7b-instruct-v0:1
     timeout_ms: 5000
     # split (default) | open | closed. See src/config.ts for semantics.
     failMode: split
@@ -316,7 +322,11 @@ filter:
 escalation:
   judge:
     enabled: true
-    model: claude-haiku-4-5
+    # Pinned to Mixtral (fast tier) — a genuinely different model family from the
+    # standard/powerful tier's Claude 3 Sonnet, so this still satisfies the
+    # diversity property recommended above despite both running on the same
+    # Bedrock account.
+    model: mistral.mixtral-8x7b-instruct-v0:1
     timeout_ms: 5000
 
 # Dream engine — background knowledge graph maintenance (issue #27).
--- a/docs/adr/023-aws-bedrock-mistral-llm-provider.md
+++ b/docs/adr/023-aws-bedrock-mistral-llm-provider.md
@@ -1,0 +1,211 @@
+# ADR-023: AWS Bedrock as the configured LLM provider
+
+Date: 2026-07-06
+Status: Accepted
+
+## Context
+
+ADR-007 established Anthropic as Curia's primary LLM provider behind a
+provider-agnostic `LLMProvider` interface specifically so that switching or
+adding providers would be a config change, not an architecture change. The
+operator running this deployment wants to run entirely on AWS Bedrock, serving
+Mistral models, instead of the Anthropic API — this allows for easier swapping between models.
+
+This is exactly the scenario ADR-007 anticipated: "different agents may
+benefit from different providers" and "switching the primary provider... requires
+only a config change, not code changes," backed by "a thin adapter per
+provider." `OpenRouterProvider` already demonstrated that a second adapter
+plugs in cleanly. This ADR records the decision to add a third.
+
+Two implementation questions were resolved before building:
+
+1. **Bedrock access path.** Calls go directly to Bedrock via `@aws-sdk/client-bedrock-runtime`.
+2. **Credential storage.** ADR-021 (vault-only secret resolution) requires
+   every provider credential except the four vault-bootstrap values to live in
+   the encrypted secrets vault, not `.env`. The AWS access key and secret
+   follow that path (`aws_access_key_id`/`aws_secret_access_key` vault keys,
+   resolved by `applyVaultSecrets()` like every other provider key). Region and
+   request timeout are non-secret operational config and stay in `.env`,
+   matching `TIMEZONE`/`HTTP_PORT`.
+
+## Decision
+
+Add `BedrockMistralProvider` (`src/agents/llm/bedrock-mistral.ts` — the
+filename predates the final model choice below; the class itself is not
+Mistral-specific, it works for any Converse-compatible Bedrock model)
+implementing `LLMProvider` via Bedrock's **Converse API**, register it in
+`providerRegistry` in `src/index.ts` alongside (not instead of, at the code
+level) Anthropic and OpenRouter.
+
+Final `config/default.yaml` `model_routing.tiers`:
+- **fast**: `mistral.mixtral-8x7b-instruct-v0:1`
+- **standard/powerful**: `anthropic.claude-3-sonnet-20240229-v1:0` — served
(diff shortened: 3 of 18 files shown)


config/default.yaml
josephfung, line 1: @AnshulaChowdhury _love_ that you're adding bedrock support! Rather than modifying the default.yaml, it would be better to override in your own local.yaml (not committed to git). That way you have bedrock support out of the gate, but it doesn't break new installs for others.

As-is, committing this change will break new installs - the install wizard currently only asks for anthropic credentials, so wouldn't be able to call these models.

docs/adr/023-aws-bedrock-mistral-llm-provider.md
josephfung, line 1: Thanks for the ADR - this is helpful 👍

src/agents/llm/model-registry.ts
```diff
@@ -130,6 +130,61 @@ const MODEL_REGISTRY: Record<string, ModelMetadata> = {
...
+  // above, but still worth reconfirming against the current Bedrock pricing page
+  // if this stops matching your invoice.
+  'anthropic.claude-3-sonnet-20240229-v1:0': {
+    provider: 'bedrock',
+    contextWindow: 200_000,
+    pricing: {
+      inputPerMToken: 3.00,
+      outputPerMToken: 15.00,
+    },
+    capabilities: ['vision', 'reasoning', 'coding'],
```
josephfung, line 185: I recommend adding `large_context` to this set of capabilities

Future design goals includes supporting auto-selection of models based on whether they are enabled and have the right capabilities.

src/agents/llm/model-registry.ts
```diff
@@ -130,6 +130,61 @@ const MODEL_REGISTRY: Record<string, ModelMetadata> = {
...
   },
 
+  // AWS Bedrock — models reachable via the Converse API (bedrock-mistral.ts;
+  // despite the filename, the provider class works for any Converse-compatible
+  // Bedrock model, not just Mistral — see docs/adr/023 for the model search
+  // that led here).
+  //
+  //
+  // mistral-large-2402 is kept registered (usable, real) but is no longer the
+  // coordinator's tier — see the Claude 3 Sonnet entry below for why.
```
josephfung, line 140: This comment seems to apply to your specific installation, not the the default/basic installation. Unless your intent was to actually change the default behaviour for all default installations - in which case I'd love to better understand the reasoning.

docs/adr/023-aws-bedrock-mistral-llm-provider.md
josephfung: I've finished reading through the ADR - any chance you can rework it to apply to all readers of the code? 

It's heavily anchored on your deployment, so mixes together broadly-applicable things (adding support for Bedrock and 3 new models) and deployment-specific things (configuring your coordinator and the 3 model tiers selected.

CHANGELOG.md
```diff
@@ -15,192 +15,28 @@ bus event types) are noted explicitly even in the `0.x` range.
...
-- **Ant Farm AF-6** — Docker image builds both SPAs; `antfarm-static` route serves `/antfarm/*` before the console wildcard. Closes #1319.
-- **Ant Farm real art** — the office renders licensed LimeZu tiles/furniture and animated premade character sprites when present, served only behind session auth (`/api/antfarm/assets/*`); falls back to procedural placeholders when absent. (#1335)
+- **AWS Bedrock LLM provider** — `BedrockMistralProvider` adds a third `LLMProvider` (alongside Anthropic and OpenRouter) via Bedrock's Converse API. Fast tier runs Mixtral 8x7B; standard/powerful run Claude 3 Sonnet via Bedrock, chosen after live load testing showed a candidate Mistral Large model silently narrating tool calls as text instead of invoking them under real prompt/tool-count load. Anthropic is no longer unconditionally required at boot — provider requirements now follow from which models `config/default.yaml`'s `model_routing` tiers actually reference. AWS credentials resolve from the secrets vault (ADR-021), not `.env`. A new `pnpm run smoke:bedrock-tools` script guards against this failure mode regressing. (ADR-023)
```
josephfung, line 18: It's a nit, but I've been trying to keep the CHANGELOG additions to <15 words where possible (I've found a lot of the coding harnesses get *super* wordy). Assuming you're using an agent, can you ask it to shorten these to <15 words?

CHANGELOG.md
```diff
@@ -15,192 +15,28 @@ bus event types) are noted explicitly even in the `0.x` range.
...
-- **Console task lineage** — tasks and jobs created from the console carry principal lineage, so console-originated work is no longer stranded propose-only when later woken. (#1127)
-- **`AGENTS.md`** — cloud-agent environment notes (Node 24 toolchain, Docker Postgres, vault secrets, and how to run lint/typecheck/test/build).
+- **Bedrock: malformed conversation history crashed every subsequent turn** — a conversation that had accumulated orphaned turns (from earlier failed LLM calls not persisting an assistant reply) or a mid-history summarization turn could violate Bedrock Converse's strict "starts with user, strictly alternates" requirement, throwing `ValidationException` on every message. `BedrockMistralProvider` now normalizes the sequence (drops leading non-user messages, merges consecutive same-role turns) before sending it. The underlying gap in `AgentRuntime`/`WorkingMemory` — not persisting an assistant turn after a failed call, and not hoisting mid-history system turns — still needs its own fix; this is a defensive mitigation in the Bedrock adapter only. (ADR-023)
```
josephfung, line 22: same as above

josephfung (changes requested):
Looks great - thanks for the contribution! 

At a high level, the one adjustment needed is to ensure this commit supports your deployment, but doesn't change the default behaviour for other deployments. The best way to do that is to *not* edit `default.yaml`, but to rather edit your own non-committed `local.yaml`

The other points are mostly nits in documentation to support the above.

If I'm misunderstanding your goal, and you actually want to change the default behaviour for all deploys, let me know. i'd love to better understand the thinking on that. 

Also, there are a number of conflicts to resolve after a rebase

josephfung:
@coderabbitai review

docs/adr/023-aws-bedrock-mistral-llm-provider.md
```diff
@@ -0,0 +1,211 @@
+# ADR-023: AWS Bedrock as the configured LLM provider
```
josephfung, line 1: @AnshulaChowdhury I also just realized this PR was numbered, before others were merged in. Can you renumber to 26, which I'll reserve for you?

josephfung:
Hey @AnshulaChowdhury, following up on this one

I'm hoping to close this PR but keep the work, and want to check with you beforehand rather than after.

Main has moved enough since that all 18 files conflict (also claude-3-sonnet-20240229 has aged out). So rather than let it sit, I've split it up:

- ADR-041 (#1768): your ADR, body unedited, renumbered (023 got taken) and re-statused to Proposed since there's no Bedrock provider in main yet. Credit is in the header, the commit, and both issues. It isn't in the git author field only because our DCO check won't let me sign off on someone else's commit.

- #1766: tracks the provider, with your design as the starting point. #1345 already listed this PR as a dependency, so closing it without a replacement would have broken my own backlog. Tagged help wanted, and it's yours first if you ever want to come back to it. No expectation either way.

- #1767: the orphan-turn bug - thanks for catching.

I'll leave this open for a week or so. If you'd rather rebase and finish it, say the word and I'll leave it open as long as you need! Otherwise I'll close it and everything above stays live. Thanks so much for the contribution

josephfung:
Closing this out as per my previous comment.

PR #1371, opened 2026-07-09
feat(install): unify installer version selection under one CURIA_VERSION knob

josephfung:
## Summary

`install.sh` had **two uncoordinated version axes**: `CURIA_REF` (default `main`) picked the git ref for the config/compose files, while `CURIA_IMAGE_TAG` (in `.env.example`, default `latest`) picked the container image. Net effect: **config files came from `main` HEAD but images from the latest release** — they could drift. This also surfaced as a README/docs inconsistency (README said download `install.sh` from `<vX.Y.Z>/…`, docs from `main/…`), and even the README's tag URL didn't actually pin because `install.sh`'s internal `CURIA_REF` defaults to `main` regardless.

This unifies both axes under a **single `CURIA_VERSION` knob (default `latest`)** that selects a release; the config files and the app image now come from the **same** release. No per-release edits to `install.sh` ever again.

## How it works

`CURIA_VERSION` resolves to a config source + an image tag:

| `CURIA_VERSION` | config source | image tag |
|---|---|---|
| `latest` (default) | `releases/latest/download/` | `latest` |
| `vX.Y.Z` / bare `X.Y.Z` | `releases/download/vX.Y.Z/` | `vX.Y.Z` |
| `edge` | `raw.../main` | `edge` |

- `CURIA_REF=<ref>` is retained as a **hidden advanced override** (raw ref; no image-tag pin).
- `install.sh` writes `CURIA_IMAGE_TAG` on **fresh install only**, so a re-run never clobbers an operator's pinned tag.
- `curia-postgres` stays on `CURIA_POSTGRES_TAG=pg16` — it publishes no semver tags, so it's an intentionally separate axis (documented in `.env.example`).

## Changes
- **`install.sh`** — `resolve_version` (with bare-version normalization + strict validation matching `release.yml`), `ASSET_MODE`-branched fetches, flat-asset→path map (`Caddyfile`→`deploy/Caddyfile`, `setup-common.sh`→`scripts/setup-common.sh`), fresh-install image-tag pin.
- **`release.yml`** — stages the install bundle with flat asset names, signs a single `SHA256SUMS` manifest (cosign keyless), attaches all **12** assets, and extends the re-download-verify gate to cosign-verify the manifest + `sha256sum -c` the bundle.
- **`tests/`** — 15 new `resolve_version` cases (66 pass).
- **`.env.example`** — documents the postgres separate-axis.

## Sequencing (important)
The docs URL flip is **deliberately not in this PR**. `releases/latest/download/install.sh` only resolves once a release carries the bundle, so:
1. Merge this PR.
2. Cut **v0.40.1** (carries the 12-asset bundle via the updated `release.yml`).
3. Smoke-test the CDN, then a follow-up PR flips README + curia-docs to `releases/latest/download/…` + documents `CURIA_VERSION`.

Until then docs keep pointing at the working `raw.../main/install.sh`.

## Verification
- `shellcheck` + `actionlint` clean (only pre-existing SC2034 on `REPO_ROOT`/`COMPOSE_FLAGS`).
- `bash tests/setup/test-setup-functions.sh` → **66 passed, 0 failed**.
- Reviewed by code-reviewer + silent-failure-hunter: no bugs, error handling sound, "green = present + verified" release invariant preserved.

Design notes: install.sh relies on TLS + `curl -fsSL` (a 404 on a missing asset fails loudly) rather than client-side cosign verification, keeping host deps to docker + curl; the signed `SHA256SUMS` is for CI's self-check + optional operator verification.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -22,7 +22,11 @@ HTTP_PORT=3000
 
 # Published image tags (Docker Compose pulls these). Pin to a vX.Y.Z for a
 # stable self-host; `latest` tracks the newest release, `edge` the newest main.
+# On a fresh install, install.sh sets CURIA_IMAGE_TAG from CURIA_VERSION so the
+# app image matches the release its config came from.
 CURIA_IMAGE_TAG=latest
+# CURIA_POSTGRES_TAG is a SEPARATE axis, keyed to the Postgres major (pg16) — the
+# curia-postgres image publishes no vX.Y.Z tags, so it is not driven by CURIA_VERSION.
 CURIA_POSTGRES_TAG=pg16
 
 # Public domain for automatic HTTPS. Only used with the TLS overlay (below).
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,6 +3,10 @@ name: Release Artifacts
 # Builds, signs, and attaches release artifacts on every published release:
 #   - sbom.spdx.json            (SPDX SBOM for the release commit)
 #   - curia-<tag>.tar.gz        (reproducible source tarball)
+#   - the operator install bundle (install.sh, docker-compose.yml,
+#     docker-compose.tls.yml, .env.example, Caddyfile, setup-common.sh) so
+#     install.sh can fetch a release's config from releases/<ref>/download/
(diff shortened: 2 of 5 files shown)


josephfung:
@coderabbitai review

PR #1395, opened 2026-07-13
fix(scheduler): liveness check no longer flaps fail on a healthy scheduler

josephfung:
## Summary

- `lastTickAt` — the timestamp `checkScheduler` reads for liveness — was stamped only by the 5-min watchdog interval, while `scheduler_max_tick_s` (120s) treats it stale after 2 minutes. That mismatch made `/api/health` report `scheduler: fail` for ~3 of every 5 minutes on a healthy, actively-polling scheduler, burning the signal value of `degraded` for anything alerting on it.
- Move the `lastTickAt` stamp to the top of `pollDueJobs` — the scheduler's real 30s liveness cadence — and leave the 5-min watchdog interval to own `recoverStuckJobs` only.

## Test plan

- [x] `pnpm run typecheck`
- [x] Full unit test suite (`vitest run`) — 4916 passed, 0 failed
- [x] Added test: `pollDueJobs` stamps `lastTickAt` on every call, including when the query throws
- [x] Added test: the watchdog interval alone (without `pollDueJobs` running) does *not* stamp `lastTickAt`
- [x] Existing `checkScheduler` unit tests (ok-within-grace, fail-past-grace, ok-when-recent, fail-when-stale) already cover the health-check logic itself — unchanged by this PR

Closes #1359

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 - **Coordinator** — external replies use first-person single-voice; internal specialist names forbidden. (#1354)
 - **Dispatcher** — content-filter blocks on relayed replies retry with reason, then salvage draft. (#1355)
+- **Scheduler** — liveness check no longer flaps `fail` every few minutes on a healthy scheduler. (#1359)
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


josephfung:
@coderabbitai review

PR #1449, opened 2026-07-19
fix(web-browser): survive frame.evaluate serialization (restore all page reads)

josephfung:
## Summary

Hotfix: the `web-browser` skill has been failing **every** page read in production since the element-refs deploy (`c40d6c4a`). The agent's self-diagnosis ("can't load the page") was wrong — navigation succeeds; content extraction throws.

## Root cause

The element-refs change refactored `extractFrameContent` to use two **named inner arrow helpers** (`roleOf`, `nameOf`). Prod loads skills via `tsx` (esbuild with `keepNames: true`), which wraps every named inner function in a module-scope `__name(...)` helper. Playwright ships `extractFrameContent` into the browser page by serializing it with `.toString()` (`frame.evaluate`), but `__name` doesn't exist in the page context, so **every** `frame.evaluate` threw `ReferenceError: __name is not defined`. `getCleanedContent` then saw all frames fail and threw `Failed to read page content: all frames errored during extraction`.

Verified in prod: the transpiled function contained `const roleOf=__name(el=>{…})`. The tunnel/proxy path and today's resolver/generation changes are fine — this broke all sites, not just the survey.

Why tests missed it: happy-dom tests call the function directly in Node (where `__name` exists), and vitest's own esbuild has `keepNames` **off**, so nothing exercised the serialize-into-browser path.

## Fix

Inline the role/name computation into the loop — no named inner functions — so the function stays self-contained when serialized. Behavior is unchanged (same role/name precedence). Anonymous `forEach` callbacks are fine (`keepNames` only wraps named bindings), which is why the original pre-refs extractor worked.

## Test

New regression test transpiles `dom-extract.ts` with `keepNames: true` (the prod `tsx` pipeline) and runs `extractFrameContent` in a fresh scope like `frame.evaluate` does. It fails with `ReferenceError: __name is not defined` before this fix and passes after. This is the first test to exercise the real serialization path.

48/48 `skills/web-browser/` tests pass; `pnpm run typecheck` clean.

## Deploy

Prod is currently degraded (all browser reads fail). Once CI is green, redeploy `:edge` to pick this up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Fixed
 
+- **`web-browser`** — fix a regression that broke every browser page read after the element-refs change.
 - **Scheduler** — a run finishing after a concurrent pause/cancel no longer overwrites that state. (#1409)
 - **Coordinator** — external replies use first-person single-voice; internal specialist names forbidden. (#1354)
 - **Dispatcher** — content-filter blocks on relayed replies retry with reason, then salvage draft. (#1355)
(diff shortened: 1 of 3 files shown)


josephfung:
@coderabbitai review

PR #1450, opened 2026-07-20
fix(browser): resolve bracketed element refs and make them document-stable

josephfung:
## Summary

Fixes a production regression in the `web-browser` skill's element-refs feature: on the 16personalities survey (and any page with custom radios), **every click hung until the 30s skill timeout**, and the keyboard fallback then acted on the wrong focus. Root-caused with prod logs and by driving a real browser at the target DOM.

Two defects, both from the element-refs change (#1449 and the original feature):

### 1. The agent passed refs *with* brackets; the resolver rejected them
The page shows refs as `[g5f0e12] radio "Agree"` and the skill description told the agent to use `[gNfNeM]`, so the model passed the token back **with brackets**. `resolveLocator`'s fast-path regex required a bare token, so a bracketed ref fell through to fuzzy role/label/text matching, found nothing, and handed `humanClick` a zero-match locator whose **untimed `boundingBox()` waited the default 30s**. Prod logs confirmed the agent sent `"[g1f0e55]"`.

**Fix:** strip a leading `[` / trailing `]` before the regex, so `g5f0e12` and `[g5f0e12]` both resolve.

### 2. Refs died faster than the agent could use them
Refs were renumbered on every extraction (a fresh `generation` per read), but **every action re-reads the page** — so a ref was invalidated within one action of being issued. Reading once and acting on several elements (the natural survey flow) failed closed on every action after the first.

**Fix:** an element keeps its ref across re-extractions; new elements get a fresh id from a per-document `seq`. Refs are stamped with a per-document identity `(epoch, frame)` adopted once, from a monotonic handler seed that never resets. A fresh document (new window post-navigation) takes the current epoch and pins its frame index; a re-read document keeps both. This keeps refs stable **within** a document and distinct **across** documents and frames — so a ref from a prior page (older epoch) or a reindexed sibling frame matches nothing (fail closed) rather than clicking the wrong element.

Ref format: `g<gen>f<frame>e<n>` → `g<epoch>f<frame>e<n>` (same shape; the `g` component is now a per-document epoch, not a per-extraction generation).

## Why it was invisible in tests
The original clickability was fine (verified: the opacity:0 native radios are the top hit-target at their own centre). The failure was purely ref resolution → a zero-match locator that hangs. Added tests now cover the actual failure modes.

## Testing
- `pnpm run typecheck` clean.
- **51/51** unit tests pass across `dom-extract.test.ts` and `handler.test.ts`, including new coverage for:
  - bracketed-ref resolution (the exact shape the agent copies),
  - cross-extraction ref stability (read once, act many),
  - cross-document fail-closed after navigation,
  - frame-index-shift collision avoidance,
  - the keepNames serialization guard (retained).
- Two rounds of adversarial review (code-reviewer + silent-failure-hunter); both findings from round 1 and the frame-reindex finding from round 2 are fixed with targeted tests.

## Note
This does not change the SSRF frame-filtering or the fail-closed resolver contract. `web-browser` 1.6.0 → 1.6.1 (patch: bug fix).


josephfung:
@coderabbitai review

PR #1481, opened 2026-07-20
fix(deps): pin adm-zip to >=0.6.0 (GHSA-xcpc-8h2w-3j85)

josephfung:
## Summary

Patches Dependabot alert #51. `adm-zip < 0.6.0` lets a crafted ZIP file trigger a ~4GB memory allocation (GHSA-xcpc-8h2w-3j85, **HIGH**).

It reaches us only as a **transitive dev dependency** of the optional `onnxruntime-node`:

```
adm-zip@0.6.0
└─ onnxruntime-node@1.24.3
   └─ @huggingface/transformers@4.2.0
      └─ promptfoo@0.121.18 (devDependencies)
```

## Fix

Pin `adm-zip: '>=0.6.0'` in the `overrides:` block of `pnpm-workspace.yaml` (the only place pnpm v10+ reads overrides from), then regenerate the lockfile. `0.6.0` is the current `latest`.

## Verification

- `pnpm install` applied the override (not "Already up to date").
- Lockfile now has the `overrides:` entry and resolves a **single** `adm-zip@0.6.0` (no `0.5.18` remaining).
- `pnpm why adm-zip` confirms the forced version and the dev-only consumer chain.

No source changes — dependency pin only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Security
 
+- **`adm-zip`** — pin transitive dev dep to `>=0.6.0`, clearing GHSA-xcpc-8h2w-3j85 (crafted ZIP 4GB allocation, HIGH).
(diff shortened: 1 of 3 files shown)


josephfung:
@coderabbitai review

PR #1619, opened 2026-07-28
feat(agents): opt AgentRuntime.handleTask into streaming turn primitive (#1563)

josephfung:
## Summary

Closes #1563.

Converges the text-channel tool loop onto the shared streaming primitive (`runStreamingToolLoop`) per ADR-038. Voice already used this path; text now does too via a **chat-compatible `openStream` adapter** over `chatWithRetry` — so retry/fallback, llm.call telemetry, error budgets, DelegationGuard, clarification short-circuits, and tool bus events stay intact while round-cap / invoke sequencing / message assembly cannot silently diverge.

> Note: #1563 was auto-closed when ADR-038 acceptance (#1618) landed with “Resolves #1563’s blocking question.” That PR only cleared the decision gate; this PR is the implementation.

### What changed

- **`streaming-turn.ts`** — `openStream`, `llmResponseAsStream`, `beforeRound` / `onToolUse` / `afterTools` hooks, `stopReason: 'stopped'`, optional `invokeTool` messages ctx for skill-activate splices
- **`AgentRuntime.processTask`** — with `executionLayer`, delegates the tool loop to `runStreamingToolLoop`; without EL, keeps a single `chatWithRetry` shot
- Round caps remain parameterized: text `error_budget.max_turns` via `onToolUse` turnsUsed semantics; voice `DEFAULT_STREAMING_MAX_ROUNDS=8`
- Docs: ADR-037 consequences + file headers updated

### Review follow-ups (`ae27853f`)

- Warn on defensive silent `'stopped'`; document inert AbortSignal / unreachable branches / post-loop messages shape
- Strengthen skill-activate + `openStream` snapshot unit coverage

### Test plan

- [x] `pnpm run typecheck`
- [x] `pnpm run lint`
- [x] vitest: `streaming-turn`, `tool-loop-messages`, `runtime` (unit + src), `turn-runner`

<div><a href="https://cursor.com/agents/bc-019faa28-80b7-755a-ac9a-10c98031a43b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-019faa28-80b7-755a-ac9a-10c98031a43b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/src/agents/runtime.ts
+++ b/src/agents/runtime.ts
@@ -1206,8 +1206,11 @@ export class AgentRuntime {
             }
             maybeAppendCheckpointBudgetNudge(workingMessages);
           },
-          onToolUse: async ({ toolCalls }) => {
+          onToolUse: async ({ toolCalls, messages: workingMessages }) => {
+            // Match former while-loop order: increment, nudge (uses updated
+            // turnsUsed), then enforce maxTurns before invoking tools.
             budget.turnsUsed++;
+            maybeAppendCheckpointBudgetNudge(workingMessages);
             if (budget.turnsUsed >= budget.maxTurns) {
               await this.handleBudgetExceeded(budget, taskEvent, 'maxTurns', budgetHandoff);
               earlyExitHandled = true;
--- a/src/agents/runtime.ts
+++ b/src/agents/runtime.ts
@@ -1156,12 +1156,17 @@ export class AgentRuntime {
       let lastLlmUsage: LLMUsage | undefined;
       let lastLlmProvenance: LLMCallProvenance | undefined;
 
+      // Text turns are intentionally non-cancellable today (same as pre-#1563):
+      // chatWithRetry does not take an AbortSignal, and openStream ignores
+      // params.signal. The controller is kept reachable as a required-arg
+      // placeholder for a future turn-level cancel/timeout.
+      const turnAbort = new AbortController();
       const streamResult = await runStreamingToolLoop(messages, {
         provider,
         model: streamModel,
         tools: workingToolDefs,
         maxRounds: budget.maxTurns,
-        signal: new AbortController().signal,
+        signal: turnAbort.signal,
         logger,
         openStream: async function* (this: AgentRuntime, params: StreamingTurnOpenStreamParams) {
           // NOTE: beforeRound already mutated working messages; params.messages is a
@@ -1327,7 +1332,9 @@ export class AgentRuntime {
             return 'continue';
           },
         },
-        invokeTool: async (toolCall, ctx) => {
+          invokeTool: async (toolCall, ctx) => {
+          // Prefer the primitive's working transcript (ctx.messages) so mid-turn
+          // splices (skill-activate) are visible to the next openStream snapshot.
           const transcript = ctx?.messages ?? messages;
 
           // When escalation is pending mid-batch, skip remaining tools (matches
@@ -1737,7 +1744,13 @@ export class AgentRuntime {
         },
       });
 
-      // Sync working transcript for empty-recovery / any post-loop message use.
+      // Sync the outer messages array from the primitive's working transcript.
+      // On message_end, streamResult.messages ends with the terminal assistant
+      // turn (streaming-turn appends it). Pre-#1563 the loop left messages at
+      // the last user tool_result — final assistant text was never pushed here.
+      // Harmless today: this array is not re-sent post-loop; memory persists
+      // responseContent separately; empty-recovery only runs when no terminal
+      // turn was appended (empty finalText). Do not double-append the assistant.
       messages.length = 0;
       messages.push(...streamResult.messages);
 
@@ -1746,6 +1759,12 @@ export class AgentRuntime {
       }
 
       if (streamResult.stopReason === 'stopped') {
+        // Defensive: every hook 'stop' sets earlyExitHandled above and has already
+        // published a response. Reaching here means a hook stopped silently.
+        logger.warn(
+          { agentId, conversationId },
+          'Streaming tool loop stopped without a published response — no reply sent to the user',
+        );
         return;
       }
 
@@ -1757,6 +1776,8 @@ export class AgentRuntime {
           provenance: lastLlmProvenance ?? fallbackProvenance,
         };
       } else if (streamResult.stopReason === 'exhausted') {
+        // Unreachable on the text path (defensive): onToolUse bails when
+        // turnsUsed >= maxTurns one round before this maxRounds cap would fire.
         // Treat like tool_use without completion — existing fallback / isResponseError.
         response = {
           type: 'tool_use',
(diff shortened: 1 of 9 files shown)


src/agents/runtime.ts
```diff
@@ -1112,398 +1103,597 @@ export class AgentRuntime {
...
+              );
+              earlyExitHandled = true;
+              return 'stop';
+            }
+
+            return 'continue';
+          },
+        },
+        invokeTool: async (toolCall, ctx) => {
+          const transcript = ctx?.messages ?? messages;
```
josephfung, line 1338: **Test coverage gap (non-blocking).** The skill-activate / reference-content splice now writes into `transcript = ctx?.messages` (the primitive's `workingMessages`) rather than the outer `messages`. This is correct — the snapshot the next `openStream` round takes includes the spliced system blocks — but it's a new indirection with no dedicated test. Suggest a runtime-level test: activate a skill mid-turn, then assert the instruction block appears in the messages the *following* LLM round is called with. Skill activation is load-bearing enough to warrant explicit coverage of the transcript it targets, since a regression here (splicing into the wrong array) would silently drop the skill instructions from subsequent rounds without failing any current test.

src/agents/runtime.ts
```diff
@@ -1112,398 +1103,597 @@ export class AgentRuntime {
...
+      let llmFailureHandled = false;
+      let lastLlmUsage: LLMUsage | undefined;
+      let lastLlmProvenance: LLMCallProvenance | undefined;
+
+      const streamResult = await runStreamingToolLoop(messages, {
+        provider,
+        model: streamModel,
+        tools: workingToolDefs,
+        maxRounds: budget.maxTurns,
+        signal: new AbortController().signal,
```
josephfung: **Minor: inert AbortSignal.** A fresh `new AbortController().signal` is passed but never aborted, and the `openStream` adapter ignores `params.signal` (it calls `chatWithRetry` without one). So the text path stays non-cancellable — same as before #1563 — but the code reads as if cancellation is wired. Either thread the task's real cancellation signal through, or add a one-line comment that text turns are intentionally non-cancellable and this is a required-arg placeholder.

src/agents/runtime.ts
```diff
@@ -1543,145 +1733,68 @@ export class AgentRuntime {
...
-      if (!suppressAmbientBullpen) {
-        await this.refreshBullpenContext(messages, bullpenInsertAt, agentId, injectedBullpenThreadIds);
+      if (streamResult.stopReason === 'message_end') {
+        response = {
+          type: 'text',
+          content: streamResult.finalText,
+          usage: lastLlmUsage ?? zeroUsage,
+          provenance: lastLlmProvenance ?? fallbackProvenance,
+        };
+      } else if (streamResult.stopReason === 'exhausted') {
```
josephfung, line 1778: **Defensively-unreachable on the text path — worth a comment.** With `maxRounds: budget.maxTurns`, `onToolUse` always bails (`turnsUsed >= maxTurns`) one round before the loop's own round cap would exhaust, so `'exhausted'` can't fire from the tool_use path here. Likewise `'stream_error'` (~L1768) is unreachable: `chatWithRetry` returns `null` on every failure (caught by `llmFailureHandled`), never a materialized `type:'error'` response for `llmResponseAsStream` to surface. Both are fine to keep as defense — a one-line "unreachable on the text path; defensive" note on each would save the next reader the trace. (If you'd rather make the intent explicit, setting `maxRounds` slightly above the `maxTurns` gate would make `onToolUse` the unambiguous sole cap.)

src/agents/runtime.ts
```diff
@@ -1543,145 +1733,68 @@ export class AgentRuntime {
...
-        const clarificationResponse = createAgentResponse({
-          agentId,
-          conversationId,
-          content: clarificationContent,
-          skillsCalled,
-          parentEventId: taskEvent.id,
-        });
-        await bus.publish('agent', clarificationResponse);
+      // Sync working transcript for empty-recovery / any post-loop message use.
+      messages.length = 0;
```
josephfung, line 1754: **Note: `messages` shape change (non-blocking).** After this sync, on the `message_end` success path `messages` ends with the terminal assistant turn (streaming-turn appends it to the returned transcript). The pre-#1563 loop left `messages` ending at the last user `tool_result` turn — the final assistant text was never pushed. Harmless today (the array isn't re-sent post-loop; memory persists `responseContent` separately, and empty-recovery only runs when no terminal turn was appended), but worth a comment so a future reader doesn't assume the old shape or double-append.

josephfung (commented):
## Review — architecture / robustness follow-ups (non-blocking)

I traced this against the pre-#1563 `while (response.type === 'tool_use')` loop and it's behavior-preserving: `turnsUsed`/`maxTurns` stay in lockstep, `refreshBullpenContext` timing (`beforeRound` gated on `round > 0`) maps exactly onto the old bottom-of-iteration refresh, the `consecutiveErrors`/`dbFailures` accounting and success-reset are intact, and `chatWithRetry`'s null-on-failure contract means the `llmFailureHandled` catch handles every failure. The 81 existing runtime unit tests (unchanged) densely cover the moved clarification / skill-activate / escalation / budget paths, so it's well regressed. The `openStream` adapter is a clean way to keep retry/fallback/telemetry on the chat path while sharing the loop with voice.

One behavior actually improves: the old mid-batch `break` on a pending escalation left some `tool_use` blocks with no matching `tool_result` (a latent invalid transcript that only survived because the turn ended); the new per-call `invokeTool` returns a "skipped" result for every remaining call, so every `tool_use` is now answered.

Four non-blocking follow-ups for the coding agent (inline). None need to gate the merge — flagging so they're tracked and a future reader doesn't have to re-derive the traces:

1. **Test gap** — skill-activate splice now targets the working-copy transcript; no dedicated test asserts it reaches the next round.
2. **Two branches unreachable on the text path** (`exhausted`, `stream_error`) — defensive, but undocumented.
3. **Inert AbortSignal** — text stays non-cancellable; the fresh controller reads as if cancellation is wired.
4. **`messages` local shape change** — now ends with the terminal assistant turn on success.

src/agents/runtime.ts
```diff
@@ -1543,145 +1733,68 @@ export class AgentRuntime {
...
-            reason: pendingDelegationEscalation.reason,
-            escalated: pendingDelegationEscalation.escalated,
-          },
-          pendingDelegationEscalation.escalated
-            ? 'Task stopped after non-retryable delegation failure — escalated to CEO backlog'
-            : 'Task stopped after non-retryable delegation failure — CEO backlog escalation failed',
-        );
+      if (streamResult.stopReason === 'stopped') {
         return;
       }
```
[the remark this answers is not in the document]
josephfung: Fixed in b0cfa49 / ae27853 — the defensive 'stopped' branch now logger.warns before returning so a future silent-stop hook can't strand the user with no audit trail.

Issue #1666, opened 2026-08-18
fix(skills): MCP callTool passes no timeout/signal — execution-layer timeout doesn't abort the MCP request


josephfung:
## Summary

The MCP tool handler calls `capturedSession.client.callTool({ name, arguments })` with **no `RequestOptions`** (no `timeout`, no `signal`) at `src/skills/mcp-loader.ts:569`. Meanwhile the execution layer races the handler against `manifest.timeout` (`src/skills/execution.ts:1627`) and, when that wins, rejects our promise **without cancelling the underlying MCP request**.

Found during the #1650 leak hunt as the initial MCP hypothesis. It was **not** the ~145 MB/hr leak (that is the Ajv-recompile path, #1663), but it is a real robustness gap: a slow/hung MCP subprocess leaves the SDK's pending-request entry (and its captured request/closures) alive after we have already timed out and moved on.

## Detail

- No `AbortSignal` is plumbed from the execution-layer timeout down into `callTool`, so a timed-out skill does not abort the in-flight MCP request.
- The SDK installs its own default request timeout that both rejects and deletes the pending handler, so this is bounded in practice — but our timeout firing first leaves the request running until the SDK's fires (or progress notifications keep resetting it).

## Fix

Plumb an `AbortController`/`AbortSignal` (and an explicit `timeout`) from the execution-layer timeout into `callTool(..., { signal, timeout })` so a timed-out skill also aborts the underlying MCP request.

## Acceptance criteria

- [ ] `callTool` is invoked with a `signal` (and explicit `timeout`) tied to the execution-layer timeout.
- [ ] When the execution-layer timeout fires, the underlying MCP request is aborted (verified via a mock MCP client asserting abort).
- [ ] Normal (non-timeout) tool calls are unaffected.

Found via #1650.

Issue #1665, opened 2026-08-18
chore(dispatch): RateLimiter.senderWindows never evicts idle senders (latent unbounded)


josephfung:
## Summary

`RateLimiter.senderWindows` (`src/dispatch/rate-limiter.ts:35`, `Map<senderId, WindowEntry>`) adds an entry per distinct inbound human sender and **never removes it**. The rate window is reset in place (`entry.count = 0`) but the Map entry persists for process lifetime.

Found during the #1650 leak hunt. **Not** the OOM cause (bounded by distinct-sender cardinality; scheduler-fired tasks bypass `checkSender` entirely), but a genuine unbounded-in-principle latent leak.

## Detail

- Inserted in `checkSender(senderId)` (`src/dispatch/dispatcher.ts:616`), once per first-seen sender.
- No `.delete`, TTL, or sweep anywhere in the class. Idle / one-shot senders accumulate forever.
- ~150-200 B/entry.

## Fix

Evict idle sender windows: delete an entry once its window has fully expired with no recent activity (either lazily on next access, or via a periodic sweep like the entity-context assembler's).

## Acceptance criteria

- [ ] Idle sender windows are evicted after their window expires (bounded memory under unbounded distinct senders).
- [ ] Active rate-limiting behavior within a window is unchanged.
- [ ] Test covers eviction of a stale sender window.

Found via #1650.

Issue #1664, opened 2026-08-18
chore(scheduler): burstCounts Map never evicts on completion/cancel (latent unbounded)


josephfung:
## Summary

`Scheduler.burstCounts` (`src/scheduler/scheduler.ts:137`, `Map<jobId, number>` on the singleton) accumulates entries that are never removed on normal completion, one-shot completion, cancellation, or job deletion.

Found during the #1650 leak hunt. This is **not** the OOM cause (growth is ~150 B per *distinct* drift-eligible jobId ever seen — orders of magnitude too small), but it is genuinely unbounded in principle and worth closing.

## Detail

- Inserted/incremented in `handleCompletion` on every successful drift-eligible completion (`scheduler.ts:608-609`), keyed by `jobId` (so N runs of one job → one entry).
- `.delete` exists only at `:687` (drift-pause) and `:912` (watchdog recovery of stuck jobs). There is **no** delete on normal completion, one-shot completion, cancel, or job delete.
- Result: an entry for a completed one-shot job, or a cancelled/deleted recurring drift-eligible job, lives for process lifetime.

## Fix

Add `this.burstCounts.delete(jobId)` on the normal-completion path (and on job cancel/delete) once the burst window is no longer relevant — mirroring the existing deletes at `:687`/`:912`.

## Acceptance criteria

- [ ] `burstCounts` entries are removed on normal completion of one-shot jobs and on job cancel/delete.
- [ ] Test: after a one-shot drift-eligible job completes, `burstCounts` has no entry for its jobId.
- [ ] Existing drift-pause / watchdog delete behavior unchanged.

Found via #1650.

Issue #1650, opened 2026-08-10
Prod OOM restart loop (~every 4h41m): heap leak on the non-streaming OpenRouter/OpenAI-SDK path


josephfung:
## Summary

Production Curia OOM-restarts on a steady **~4h41m cycle** (`RestartCount=62`; 36/36 boots preceded by a V8 `FATAL ERROR: JavaScript heap out of memory`; `OOMKilled=false` → V8 old-space, not the cgroup). ~**300 MB/hr** steady JS-heap growth, independent of chat load.

This issue tracks the **actual** leak. An earlier hypothesis (undrained *streaming* `Response` buffers, #1648/#1649) turned out **not** to match prod usage — see "Why not streaming" below. #1648/#1649 are defensive streaming-cleanup hardening; they do not fix this.

## What prod actually runs

- **100% OpenRouter** — models `google/gemini-3.1-flash-lite`, `deepseek/deepseek-v4-pro`, `claude-haiku-4-5` (all OpenRouter slugs). Zero native-Anthropic calls.
- **Non-streaming `provider.chat()`** — per #1563/ADR-038, text agents (all scheduled/dispatch traffic) run `runStreamingToolLoop` via a chat adapter over `chatWithRetry`, backed by non-streaming `chat()`, **not** `stream()`. Only voice opens `stream()` directly.

## Why not streaming

Over 8 days of prod logs, every streaming-only marker is absent — "streamed response truncated", "streaming API call", "stream returned no chunks" all **0** — while the non-streaming markers are present (6× "OpenRouter response truncated", 1× "OpenRouter API call failed"). So the leaking path is non-streaming `OpenRouterProvider.chat()` (`src/agents/llm/openrouter.ts:380`, `this.client.chat.completions.create(...)`).

## Hypothesis

Two exhaustive static sweeps already cleared curia's own code (all state Postgres-backed/bounded), so the retention is below app code — in the **OpenAI SDK / undici (fetch) layer** holding response buffers on plain non-streaming `chat.completions.create()` calls. Upstream [anthropics/claude-code#33380](https://github.com/anthropics/claude-code/issues/33380) reported the same class even on standard (non-streaming) calls (181 MB `Response` objects retained). The steady, load-insensitive rate matches the scheduler's around-the-clock non-streaming calls.

## Plan

[...]

PR #1667, opened 2026-08-18
fix(health): MCP liveness probe uses ping() not listTools() — stops ~145 MB/hr OOM leak

### LINKED ISSUE #1663: fix: MCP liveness probe (listTools per /api/health) leaks ~145 MB/hr via SDK Ajv recompile → OOM restart loop
## Summary

Production OOM-restarts every ~4h41m on a steady JS-heap leak of **~145 MB/hr**. Root cause is the **MCP liveness health probe**: `/api/health` calls `client.listTools()` on every MCP server, and the MCP SDK **recompiles an Ajv validator for every tool's `outputSchema` on every `listTools()` call**, accumulating compiled validator functions on the JS heap forever. The Docker healthcheck fires `/api/health` every 30s, so the leak is steady and load-insensitive.

`MemorySampler` telemetry (#1650, PR #1652) confirmed the shape: `heapUsed` climbs ~145 MB/hr while `external`/`arrayBuffers` stay flat — i.e. plain JS objects (compiled functions), not native/fetch buffers.

## The chain (each link verified against code)

1. **Docker healthcheck** curls `/api/health` every 30s (`compose.production.yaml` → `healthcheck.interval: 30s`).
2. `HealthService` runs `checkMcpServers(...)` per request (`src/health/health-service.ts:217`).
3. For each live server it calls **`session.client.listTools()`** as the liveness probe (`src/health/health-checks.ts:235`, and the deprecated `checkMcpGoogleWorkspace` at `:264`). Comment at `:44` states its purpose: "a lightweight round-trip that proves the subprocess is alive."
4. MCP SDK `Client.listTools()` calls **`cacheToolMetadata(result.tools)` on every response** (`@modelcontextprotocol/sdk/dist/esm/client/index.js:554-558`), which calls `getValidator(tool.outputSchema)` per tool.
5. The SDK's `AjvJsonSchemaValidator.getValidator` calls **`ajv.compile(schema)` unconditionally** for schemas without an `$id` — which tool `outputSchema`s are — on a shared, process-lifetime Ajv instance (`.../validation/ajv-provider.js:63-67`).
6. Ajv 8's compiled-schema cache is **`new Map()` keyed by the schema object reference** (`ajv/dist/core.js:100,447,453`). Each `listTools()` response is freshly JSON-parsed, so every schema object is a new reference → **cache miss every time** → each compiled validator (generated function + `SchemaEnv` + `ValueScopeName` + bytecode) is retained forever.

## Evidence it's this and not something else

- **Rate matches exactly:** ~120 health checks/hr × ~121 google-workspace tools × ~10 KB per compiled validator ≈ **145 MB/hr** — the measured slope.
- **Heap class matches:** compiled JS functions are pure `heapUsed`; `external`/`arrayBuffers` flat. The (process-mismatched but still indicative) prod heap snapshots were dominated by `SchemaEnv` / `ValueScopeName` / `BytecodeArray`.
- **Load-insensitive:** driven by the Docker healthcheck timer, not traffic — matches the observed steady climb.
- **Onset matches:** commit `a8ad2e28` (2026-07-27, "fail /api/health when enabled MCP servers have 0 tools") extended `listTools()` probing to **all** MCP servers; it shipped in the Aug 3 prod deploy, exactly when the trouble began.

A 4-subagent sweep of Curia's own code (scheduler, agent runtime/LLM, bus/dispatch/memory, skills/execution/MCP + custom overlay) found **no** unbounded accumulator in our code — the retention is entirely inside the SDK's Ajv cache, triggered by our probe.

## Fix

Switch the MCP **liveness** probe from `listTools()` to the MCP protocol's **`client.ping()`** RPC (`Client.ping()` → strict `EmptyResult`; no tools, no `outputSchema`, no validator compilation). `ping` is the protocol's purpose-built liveness utility and is handled at the SDK protocol layer, so all servers respond to it. Keep `listTools()` for boot-time discovery only.

The boot-time zero-tools gate is unaffected: `checkMcpServers` already fails on `boot.status === 'zero_tools' | 'unavailable'` **before** the runtime probe (`health-checks.ts:223-225`), so dropping the per-probe tool-count check does not regress zero-tools detection.

## Alternatives considered

- **Inject a memoizing `jsonSchemaValidator` into the SDK `Client`** — fixes the leak but keeps an unnecessary full `listTools()` round-trip every 30s and works around the SDK bug rather than avoiding the expensive call. `ping()` is simpler and lighter.
- **Raise `--max-old-space-size` / container memory** — only delays the OOM; the host is RAM-constrained. Not a fix.

The underlying SDK behavior is a genuine upstream bug (recompiling on a shared Ajv per `listTools`) and will be reported separately to `@modelcontextprotocol/sdk`.

## Acceptance criteria

- [ ] MCP liveness probe uses `client.ping()`; runtime `listTools()` is no longer called per `/api/health`.
- [ ] `/api/health` still reports a stalled/dead MCP subprocess as `fail` (ping times out) within the existing ~3s probe cap.
- [ ] Boot-time zero-tools / unavailable servers still report `fail` (unchanged gate).
- [ ] A test asserts the health probe calls `ping()` and not `listTools()` (guards against regression of the leak).
- [ ] No compiled-validator growth across repeated health probes (covered by the probe-path test; end-to-end confirmed by `MemorySampler` slope flattening after deploy).

Refs #1650.

josephfung:
## Summary

Fixes the production OOM restart loop (#1650, root-caused in #1663). The MCP liveness health probe called `client.listTools()` on every `/api/health` request; the Docker healthcheck hits `/api/health` every 30s. The MCP SDK recompiles an Ajv validator for every tool's `outputSchema` on **every** `listTools()` call and retains them on a shared, process-lifetime Ajv instance whose cache is keyed by schema object reference — each response is freshly parsed, so it's a cache miss every time. At ~121 tools × ~120 checks/hr that accumulated **~145 MB/hr** of compiled validator functions on the JS heap (`heapUsed` climbs, `external`/`arrayBuffers` flat), OOM-restarting prod every ~4h41m.

The fix switches the liveness probe to the MCP protocol's **`client.ping()`** (strict empty result: no tools, no `outputSchema`, no compilation). `listTools()` remains for boot-time discovery only.

Closes #1663.

## How it was found

`MemorySampler` (#1650 / PR #1652) confirmed the shape (JS-heap, not buffers) and the ~145 MB/hr slope over a 4h prod run. A parallel code sweep cleared all of Curia's own subsystems (scheduler, agent runtime/LLM, bus/dispatch/memory, skills/execution) — the retention was entirely inside the SDK's Ajv cache, triggered by our probe. The onset (`a8ad2e28`, which extended `listTools()` probing to all servers) matches the Aug 3 deploy when prod trouble began.

## Changes

- `src/health/health-checks.ts`: liveness probe → `client.ping()`; `McpSessionHealth.client` interface now declares `ping()`; doc comments explain the leak and the boot-only role of `listTools()`.
- Removed the dead, deprecated `checkMcpGoogleWorkspace` (zero callers, same leaky `listTools()` pattern).
- Rewrote `tests/unit/health/check-mcp-servers.test.ts` for `ping()` semantics, including a **regression guard** asserting the probe calls `ping()` and **never** `listTools()`, plus ping-reject and 3s-timeout cases.
- CHANGELOG entry; fixed a stale comment referencing the removed function.

## Behavior change (reviewed, accepted)

The probe no longer re-counts tools at runtime — a boot-`ok` server is `ok` as long as it answers `ping()` within 3s. The boot-time `zero_tools`/`unavailable` gate is unchanged, so zero-tools detection does not regress. The only dropped signal is a server that had tools at boot but later exposes **0** live tools; that's a *readiness/degradation* condition (not liveness — a container restart is the wrong remedy), it's rare for our static-tool-set servers, and for the highest-stakes server (google-workspace) it remains independently covered by the separate `canaryGoogleWorkspace` credential probe. If functional tool-presence ever needs monitoring, it belongs in a low-frequency readiness canary, never back on the 30s liveness path. Both a code review and a silent-failure review confirmed no dead/hung server is masked as `ok`.

## Upstream

The underlying SDK behavior (recompile-per-`listTools()` on a shared, reference-keyed Ajv cache) is a genuine bug affecting anyone who polls `listTools()`. It will be reported to `@modelcontextprotocol/sdk` separately.

## Test plan

- `pnpm run typecheck` — clean.
- `pnpm exec vitest run tests/unit/health/` — 68/68 pass.
- Regression guard proves `listTools()` is not called by the probe; ping-reject and fake-timer-timeout cases prove dead/hung servers still report `fail`.
- End-to-end: `MemorySampler` slope should flatten to ~0 after deploy.

Refs #1650. Follow-ups filed: #1664, #1665, #1666.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Fixed
 
+- **MCP health probe** — liveness uses `ping()` not `listTools()`, ending a ~145 MB/hr validator leak. (#1663)
 - **`classifyError`** — `AbortSignal.timeout` / abort DOMExceptions map to `TIMEOUT`. (#1597)
 - **`delegate`** — calendar briefs must match same-turn date-resolve output. (#1612)
 - **Coordinator** — ambient bullpen mentions no longer bleed into scheduler jobs and leak to the principal. (#1609)
--- a/src/health/health-checks.ts
+++ b/src/health/health-checks.ts
@@ -41,8 +41,19 @@ export interface BrowserServiceHealth {
 export interface McpSessionHealth {
   serverId: string;
   client: {
-    /** MCP SDK Client.listTools() — a lightweight round-trip that proves the subprocess is alive. */
-    listTools(): Promise<{ tools?: unknown[] }>;
+    /**
+     * MCP SDK `Client.ping()` — the protocol's liveness RPC. Proves the subprocess
+     * is alive with an empty-result JSON-RPC round-trip.
+     *
+     * Deliberately NOT `listTools()`: the SDK recompiles an Ajv validator for every
+     * tool's `outputSchema` on every `listTools()` call and retains it on a shared,
+     * process-lifetime Ajv instance (its cache is keyed by schema object reference,
+     * and each response is freshly parsed → cache miss every time). With the Docker
+     * healthcheck hitting `/api/health` every 30s, that leaked ~145 MB/hr and OOM-
+     * restarted prod (#1663). `ping()` returns no schemas, so it triggers no
+     * compilation. Tool discovery still uses `listTools()`, but only at boot.
+     */
+    ping(): Promise<unknown>;
   };
 }
 
@@ -203,10 +214,12 @@ export function checkBrowser(service: BrowserServiceHealth | undefined): CheckRe
  * Check every **enabled** MCP server that was attempted at boot (#1500).
  *
  * - Boot `zero_tools` / `unavailable` → fail (no live probe needed)
- * - Boot `ok` → live tools/list; fail if the call errors or returns 0 tools
+ * - Boot `ok` → live `ping()` liveness probe; fail if it errors or times out
  * - Disabled servers are absent from `serverStatuses` and never appear here
  *
- * Hard 3-second timeout per live probe.
+ * Hard 3-second timeout per live probe. The probe is `ping()`, not `listTools()`,
+ * so it triggers no MCP-SDK Ajv validator recompilation (the ~145 MB/hr leak, #1663).
+ * The boot-time `zero_tools` gate above still catches servers that expose no tools.
  */
 export async function checkMcpServers(
   serverStatuses: ReadonlyMap<string, McpServerBootStatus>,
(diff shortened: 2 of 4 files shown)


josephfung:
@coderabbitai review

PR #1671, opened 2026-08-19
fix(deps): make markdown-it 15 attrGet type-safe (supersedes #1640)

josephfung:
## Summary

Completes the markdown-it 14→15 upgrade that Dependabot proposed in #1640 and adds the small production-code fix that PR couldn't merge without.

Dependabot's #1640 bumped `markdown-it` but left CI red: v15 widened `Token.attrGet()`'s return type to `string | number`, so the auto-`target=_blank`/`rel` link-hardening rule in the outbound-email markdown renderer stopped type-checking:

```
src/format/markdown-to-html.ts(24,31): error TS2345: Argument of type 'string | number' is not assignable to parameter of type 'string'.
```

## Changes

- **`markdown-it` 14.3.0 → 15.0.0** (mirrors #1640; `linkify-it` 5 → 6 comes along transitively).
- **`src/format/markdown-to-html.ts`** — coerce `attrGet('href')` to a string (`String(... ?? '')`) before `RegExp.test()`. This is the outbound-email HTML renderer, so behavior is preserved exactly; only the type is tightened.
- **Removed `@types/markdown-it`** — v15 ships its own bundled TypeScript declarations, making the DefinitelyTyped package redundant (its stale v14 types would otherwise drift from the real API).

## Verification

- `pnpm run typecheck` — passes (exit 0), the TS2345 error is gone.
- `vitest run src/format/` — 19/19 tests pass; v15 produces identical rendered output for the existing fixtures.

Supersedes #1640 (closing that one in favor of this).


josephfung:
@coderabbitai review

Issue #1672, opened 2026-08-20
Signal voice-call channel: bridge signal-cli calls into VoiceRuntime


josephfung:
## Summary

Implement native Signal voice calls as a live transport into the existing voice pipeline: the CEO (and, policy-permitting, others) can Signal-call Curia and hold a real-time conversation through the same `VoiceRuntime` / Deepgram / Cartesia stack the console voice channel uses.

Feasibility was proven live in the #1602 spike (see [feasibility finding](https://github.com/josephfung/curia/issues/1602#issuecomment-5347032324) and [validation result](https://github.com/josephfung/curia/issues/1602#issuecomment-5357625022)): signal-cli ≥ 0.14.2 answers calls by spawning [`signal-call-tunnel`](https://github.com/visigoth/signal-call-tunnel) (RingRTC), exposes `subscribeCallEvents` / `acceptCall` / `hangupCall` over the JSON-RPC socket we already use, and moves call audio through per-call PulseAudio virtual devices (Opus 48 kHz mono, 20 ms frames). Two-way audio was confirmed end-to-end from a real phone.

## Architecture (validated in the spike)

```
 CEO phone ──Signal call──▶ signal-cli daemon ──spawns──▶ signal-call-tunnel (RingRTC/WebRTC)
                                │ JSON-RPC (existing socket):                │
                                │  callEvent / acceptCall / hangupCall      │ PulseAudio virtual devices
                                ▼                                           ▼
                          SignalAdapter ──▶ VoiceRuntime ◀── SignalAudioTransport (new)
                                              │  reads  <output>.monitor  (caller → Deepgram STT)
                                              │  writes sink_for_<input>  (Cartesia TTS → caller)
```

`VoiceRuntime` is already transport-agnostic behind the five-method `AudioTransport` seam (`src/channels/voice/audio-transport.ts`), exactly as ADR-037 §10 planned. STT/TTS accept per-transport rate hints, so the transport can advertise 48 kHz both ways — no in-process resampling expected.

### App (curia)

- [ ] `SignalRpcClient`: add `subscribeCallEvents` (must be sent on every (re)connect — signal-cli silently ignores incoming calls with no subscriber), `acceptCall`, `rejectCall`, `hangupCall`, and `callEvent` notification dispatch
[...]

PR #1674, opened 2026-08-20
fix(voice): ignore duplicate RINGING_INCOMING instead of self-rejecting the call

josephfung:
## Summary

Found live in prod validation of Signal voice calls (#1672): **every** inbound call was rejected as "busy" and none ever reached CONNECTED.

Root cause: **signal-cli re-emits `RINGING_INCOMING` for the same call while it rings.** The busy check in `handleRingingIncoming` treated the repeat as a competing call and sent `rejectCall` for the callId it was actively answering, tearing the call down. Proof from the prod logs: the very first call after startup was rejected as "busy" — impossible unless a duplicate same-call event tripped the gate.

## Fix

Guard `handleRingingIncoming` so a `RINGING_INCOMING` whose callId is already in flight (`active` / `pending` / `ringing` / `connecting`) is **ignored**, not busy-rejected. The one-call-at-a-time invariant is unchanged: a ring for a *different* callId while one is in flight is still rejected as busy.

The `connecting` slot matters for a subtle window: after `ENDED` lands mid-`startSession`, `active` is nulled and `pending` cleared, but the id stays in `connecting` until setup's `finally` runs — without that clause a duplicate in that gap would be freshly accepted for an already-ending call.

## Tests

Four regression tests, one per guard slot, each verified RED without the fix:
- `ringing` slot (duplicate while still resolving caller)
- `pending` slot (duplicate after accept, before CONNECTED)
- `active` slot (stray re-emit on the live connected call)
- `connecting` slot (duplicate in the ENDED-during-startSession gap — RED = `acceptCall` fires twice without the clause)

Full bridge suite: 79 passed. Typecheck clean. Reviewed by code-reviewer + silent-failure-hunter (both clean; the added `active`/`connecting` tests came from the code-reviewer's finding).

Relates to #1672. Merge + redeploy needed before the prod validation call can be retried.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Fixed
 
+- **Signal voice calls** — duplicate `RINGING_INCOMING` for one call no longer self-rejects it as busy. (#1672)
 - **Scheduler runtime errors** — fallback `agent.response(isError)` paths now always emit paired `agent.error`. (#1647)
 - **MCP health probe** — liveness uses `ping()` not `listTools()`, ending a ~145 MB/hr validator leak. (#1663)
 - **Scheduler** — `burstCounts` evicts one-shot completions inline and stale jobs on the watchdog sweep. (#1664)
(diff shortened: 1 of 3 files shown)


josephfung:
@coderabbitai review

PR #1686, opened 2026-08-24
fix: CC preamble omittedCount should ignore empty sanitized addresses

### LINKED ISSUE #464: CC preamble omittedCount overcounts when recipient addresses sanitize to empty string
## Summary

> **Location updated (2026-07-01):** this code moved out of `dispatcher.ts` during a refactor. It now lives in `buildCcPreamble()` at `src/dispatch/email-metadata.ts:132`. The bug is unchanged — it came along with the move.

In `src/dispatch/email-metadata.ts` (`buildCcPreamble`, ~line 126-132):

```typescript
const sanitizedRecipients = meta.primaryRecipientEmails
  .map((addr) => sanitizeEmailField(addr))
  .filter((addr) => addr.length > 0)
  .slice(0, MAX_RECIPIENTS_IN_PREAMBLE);

const omittedCount = Math.max(0, meta.primaryRecipientEmails.length - sanitizedRecipients.length);
```

`sanitizedRecipients` is built with `.filter((addr) => addr.length > 0)` (and then `.slice(0, MAX_RECIPIENTS_IN_PREAMBLE)`), which discards addresses that collapse to empty after stripping. These are **rejected/malformed** addresses — not legitimate recipients that couldn't fit due to the `MAX_RECIPIENTS_IN_PREAMBLE` cap. The current formula subtracts the filtered-and-capped list from the raw `primaryRecipientEmails.length`, conflating the two.

If any address sanitizes to empty, the `+N more` suffix shown to the coordinator implies N real recipients weren't displayed. This could cause the coordinator to misread the thread's recipient list (e.g. believing the email has more participants than it does).

Identified during review of PR #462.

## Example

- 3 addresses in `To:` field: `alice@example.com`, `<<<injection>>>` (sanitizes to empty), `bob@example.com`
- `sanitizedRecipients` = `["alice@example.com", "bob@example.com"]` (2 entries)
- `omittedCount` = 3 − 2 = 1
- Preamble shows: `alice@example.com, bob@example.com, +1 more`
- The coordinator sees a third recipient that doesn't exist

## Fix

Track filtered-out (sanitized-to-empty) addresses separately from addresses dropped by the cap — split the filter step from the slice step:

```typescript
const sanitizedAll = meta.primaryRecipientEmails
  .map((addr) => sanitizeEmailField(addr))
  .filter((addr) => addr.length > 0);
const sanitizedRecipients = sanitizedAll.slice(0, MAX_RECIPIENTS_IN_PREAMBLE);
const omittedCount = Math.max(0, sanitizedAll.length - sanitizedRecipients.length);
```

## Acceptance criteria

- [ ] `omittedCount` reflects only addresses dropped by the `MAX_RECIPIENTS_IN_PREAMBLE` cap, not addresses that sanitized to empty
- [ ] Addresses that sanitize to empty are silently discarded (or logged at debug/warn) — they do not inflate `omittedCount`
- [ ] Unit test: 12 addresses where 2 sanitize to empty and 10 are valid — preamble shows 10 addresses with no `+N more` suffix
- [ ] Unit test: 11 valid addresses (none sanitize to empty) — preamble shows 10 with `+1 more`

VedantMadane:
## Summary

CC preamble omittedCount should ignore empty sanitized addresses

## Changes

- omittedCount only counts cap-truncated valid recipients, not empty sanitizations

Fixes #464

### DIFF
--- a/src/dispatch/email-metadata.ts
+++ b/src/dispatch/email-metadata.ts
@@ -131,13 +131,14 @@ export function buildCcPreamble(
   // from the Nylas To-field (attacker-controlled) and is injected into
   // taskContent after the injection scanner has already run on payload.content.
   // Without sanitization a crafted address could bypass Layer 1.
-  const sanitizedRecipients = meta.primaryRecipientEmails
+  const nonEmptyRecipients = meta.primaryRecipientEmails
     .map((addr) => sanitizeEmailField(addr))
-    .filter((addr) => addr.length > 0)
-    .slice(0, MAX_RECIPIENTS_IN_PREAMBLE);
+    .filter((addr) => addr.length > 0);
+  const sanitizedRecipients = nonEmptyRecipients.slice(0, MAX_RECIPIENTS_IN_PREAMBLE);
 
-  // Count items dropped by the empty-string filter or by the 10-item cap.
-  const omittedCount = Math.max(0, meta.primaryRecipientEmails.length - sanitizedRecipients.length);
+  // Only count valid addresses omitted by the display cap — empty/sanitized-away
+  // addresses must not inflate "+N more".
+  const omittedCount = Math.max(0, nonEmptyRecipients.length - sanitizedRecipients.length);
 
   // Only append "+N more" when there are named recipients to precede it —
   // "unknown recipients, +3 more" is misleading when every address was stripped.
--- a/tests/unit/dispatch/email-metadata.test.ts
+++ b/tests/unit/dispatch/email-metadata.test.ts
@@ -191,8 +191,8 @@ describe('buildCcPreamble', () => {
     expect(result).not.toContain('+');
   });
 
-  it('counts filtered-out (empty-after-sanitize) recipients in the omitted total', () => {
-    // '\n\r' sanitizes to empty and should be counted as omitted
+  it('does not count filtered-out (empty-after-sanitize) recipients in the omitted total', () => {
+    // '\n\r' sanitizes to empty and should not be counted as omitted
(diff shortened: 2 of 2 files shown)


josephfung (changes requested):
Thanks for picking this up, the logic is right 👍 

CI's red though: `tests/unit/dispatch/email-metadata.test.ts:181` asserts the old buggy behavior (expects +1 more). That test came in with #462 (same review that found this bug) so it basically encoded the bug as intent. In this PR please flip it rather than delete it: same inputs, expect alice@example.com with no +.

Also need the two tests from the acceptance criteria in #464:

- 12 addresses, 2 sanitize to empty, 10 valid → 10 shown, no +N more
- 11 valid → 10 shown with +1 more

The first one is the actual regression guard.

If you rebase on main while you're in there and it should go green. Thanks!

VedantMadane:
Hi @josephfung,

Thank you for the review!
- Rebased on upstream/main.
- Flipped the assertion in 	ests/unit/dispatch/email-metadata.test.ts so that empty/sanitized recipients do not increment the omitted count (expecting lice@example.com without +).
- Added the regression test for 12 recipients (2 empty, 10 valid) expecting 10 shown with no +N more.
- Added the test for 11 valid recipients expecting 10 shown with +1 more.
- Verified all 38 unit tests pass cleanly.

Ready for review!

josephfung (approved):
Thx! Test flipped rather than deleted, both ACs from #464 covered, CI green. Thanks for the rebase 🙏 Approving.

PR #1697, opened 2026-09-01
feat(channels): transcribe inbound Signal and Slack voice notes

### LINKED ISSUE #1600: Voice notes: inbound transcription on Signal + Slack
## Problem

A voice note sent to Curia over Signal/Slack is not transcribed; it never becomes an inbound message the agent can act on.

## Approach

In the Signal and Slack inbound paths, detect audio attachments, download the bytes, run batch STT via `SpeechMediaService`, and inject the transcript as normal inbound message content (tagged transcribed-from-audio for provenance).

The service shipped in #1621. Import from the barrel `src/speech/index.js`:

```ts
const result = await speechMediaService.transcribe({
  audio,          // Uint8Array — the downloaded attachment bytes
  contentType,    // MIME hint: Signal ≈ 'audio/ogg' (opus); Slack ≈ 'audio/mp4' / 'm4a' / 'webm'
  language,       // optional
});
// result: { ok: true,  value: { text, confidence?, durationSeconds? } }
//       | { ok: false, error: AgentError }
```

Key integration facts:

- **The service returns a result, it does not throw** — and it is not the skill `{ success }` shape. Branch on `result.ok`; on failure `result.error` is a structured `AgentError` (`.type`, `.retryable`, `.message`, `.context.status`).
- **Empty transcript is a success.** When the provider recognizes no speech, `transcribe` returns `{ ok: true, value: { text: '' } }`. Do **not** inject an empty inbound message — drop it, or surface a short "couldn't make that out" note.
- **Attachment bytes must be fetched first.** Signal via `src/channels/signal/signal-rpc-client.ts`; Slack file URLs require the bot-token `Authorization` header to download.
- **The DI seam already exists.** `speechMediaService` is constructed in `src/index.ts` (gated on Deepgram + Cartesia + voice id) and currently parked as `void speechMediaService;`. This issue threads it into the Signal/Slack adapter constructors and removes the `void`.
- **STT/TTS are coupled**: one credential gate builds both, so `speechMediaService` may be `undefined` (creds absent). Treat `undefined` as "voice notes disabled" and fall through to the existing text handling.
- Batch calls self-bound at 60s (`resolveBatchSignal`); pass your own `signal` only if you need earlier cancellation. Don't add a second timeout.
- Retryable vs terminal: `error.retryable` is true for `TIMEOUT` / `RATE_LIMIT` / `PROVIDER_ERROR` (abort/timeout now map to `TIMEOUT`), false for `AUTH_FAILURE` / `VALIDATION_ERROR`.

## Files

- `src/channels/signal/` (inbound attachment detect + download)
- `src/channels/slack/` (inbound file detect + authenticated download)
- `src/index.ts` (thread `speechMediaService` into the adapters; remove the `void`)

## Acceptance criteria

- [ ] A voice note over Signal produces a normal inbound message carrying the transcript, tagged transcribed-from-audio.
- [ ] Same for Slack.
- [ ] The attachment MIME type is passed as `contentType`.
- [ ] An empty transcript (`text: ''`) does not produce an empty inbound message.
- [ ] Transcription failure surfaces an honest error from `result.error` (type + message), never a silent drop.
- [ ] When `speechMediaService` is unavailable (creds not configured), the inbound path degrades to existing text behavior without error.

## Dependencies

- **#1597 — batch STT/TTS service (shipped in #1621).**

josephfung:
## Summary

Voice notes sent over Signal or Slack were dropped (empty text / file-only envelopes) and never became inbound messages the agent could act on. This wires `SpeechMediaService` (shipped in #1621) into the Signal and Slack adapters so audio attachments are downloaded, transcribed, and published as ordinary `inbound.message` content tagged `[transcribed-from-audio]`.

Closes #1600

## Changes

- **Signal:** `getAttachment` RPC downloads bytes; audio-only envelopes convert instead of being ignored; adapter transcribes via `speechMediaService`.
- **Slack:** `downloadFile` fetches private URLs with the bot-token Bearer header; `file_share` + audio files are accepted; adapter transcribes the same way.
- **DI:** `src/index.ts` passes the existing `speechMediaService` into both adapters (the previous `void` park is gone). Missing creds leave the service `undefined` and the inbound path stays on the existing text behavior.
- Empty transcripts (`text: ''`) publish a short "couldn't make that out" note — never an empty inbound. Transcription failures publish `error.type` + `error.message` instead of dropping.

## Related Issues

Closes #1600

## Checklist

- [x] CI passes


src/channels/signal/signal-rpc-client.ts
```diff
@@ -198,6 +198,31 @@ export class SignalRpcClient extends EventEmitter {
...
+    const params: Record<string, unknown> = {
+      account: this.config.accountNumber,
+      id: opts.id,
+    };
+    if (opts.groupId) {
+      params.groupId = opts.groupId;
+    } else if (opts.recipient) {
+      params.recipient = opts.recipient;
+    }
+    const result = await this.call('getAttachment', params);
```
josephfung: **Blocking — needs a size cap before this call.**

`getAttachment` returns the whole attachment base64 over the same JSON-RPC socket, and `handleData` accumulates it as a single line: `this.buffer += ...`, then `split('\n')`, then `line.includes('"callId"')`, then `JSON.parse`. Signal permits attachments up to 100 MB, so a worst case here is a ~133 MB single line through that path — a large memory spike plus a multi-second event-loop stall. Given the prod OOM history (#1650) that's the finding I'd hold the PR on.

The guard is free because `SignalAttachment.size` is already in the envelope: reject above a few MB **before** calling `getAttachment` (see my comment on the adapter) rather than defending in here. It also caps Deepgram spend per message.

Separately: `REQUEST_TIMEOUT_MS` is a shared 10s for every RPC call. That's fine for a local-socket disk read, but worth a comment on this method noting `getAttachment` now rides that budget so nobody tunes it down without thinking about attachments.

src/channels/signal/signal-adapter.ts
```diff
@@ -429,4 +455,54 @@ export class SignalAdapter implements Channel {
...
+    if (!audio || !speech) {
+      if (!converted.content) {
+        return { content: null, metadata };
+      }
+      return { content: converted.content, metadata };
+    }
+
+    let resolved;
+    try {
+      const bytes = await this.config.rpcClient.getAttachment({
```
josephfung: **Blocking — size guard belongs here, before the download.**

`audio.size` is on `SignalAttachment` already. Add a cap (a few MB is generous for a voice note) and short-circuit to a "voice note too long to transcribe" publish instead of downloading. That protects the RPC line buffer (see my comment on `signal-rpc-client.ts`) *and* bounds per-message Deepgram cost in one check.

Worth extracting the cap as a shared constant in `inbound-voice-note.ts` so Slack uses the same number.

src/channels/signal/signal-adapter.ts
```diff
@@ -429,4 +455,54 @@ export class SignalAdapter implements Channel {
...
+        id: audio.id,
+        recipient: converted.metadata.isGroup ? undefined : converted.senderId,
```
josephfung: **Should fix — `recipient` fallback for ACI-only senders.**

signal-cli requires exactly one of `recipient` / `groupId` (they're a required mutually-exclusive group upstream). The 1:1 branch passes `converted.senderId`, which is `envelope.sourceNumber` — empty for username/ACI-only senders as Signal moves off phone numbers. When that happens neither param is set, signal-cli errors, and the sender gets a "download failed" note.

`sourceUuid` is already on `SignalEnvelope` and signal-cli accepts it as `--recipient`; it just isn't plumbed into `converted.metadata` by the converter. Falling back to it makes this robust.

src/channels/signal/signal-adapter.ts
```diff
@@ -223,12 +234,27 @@ export class SignalAdapter implements Channel {
     }
 
     // ------------------------------------------------------------------
-    // Step 3: Sanitize and publish
+    // Step 3: Voice-note transcription (#1600)
```
josephfung: **Blocking — no tier gate on the 1:1 path.**

This new Step 3 runs after contact resolution, but hold/block policy lives in the dispatcher, not here. So any unknown *or blocked* 1:1 Signal sender triggers a download plus a paid Deepgram call on every voice note, unthrottled. Group traffic is protected by the Step 0 trust check; 1:1 isn't.

`isKnownSender` is computed two steps up for read receipts — gating transcription on the same signal (or `meetsMinimumTier(tier, 'known')`) is a couple of lines and matches existing policy precedent in this file. An unknown sender's voice note should degrade to the same "nothing to publish" path as STT-disabled, not to a billed API call.

src/channels/inbound-voice-note.ts
```diff
@@ -0,0 +1,144 @@
...
+export function resolveVoiceNoteInbound(opts: {
+  originalText: string;
+  result: SpeechMediaResult<TranscribeFileResult>;
+}): VoiceNotePublish {
+  if (!opts.result.ok) {
+    const { type, message, retryable } = opts.result.error;
+    return {
+      content: formatVoiceNoteContent(
+        opts.originalText,
+        `Voice note transcription failed (${type}): ${message}`,
```
josephfung: **Blocking — this leaks internal provider state to external senders.**

This string lands in `content`, which is what the coordinator reads and what it will paraphrase back to whoever sent the voice note. So a stranger who sends a voice note during a Deepgram outage gets told `Voice note transcription failed (AUTH_FAILURE): Deepgram STT request failed with HTTP 401`. That discloses provider identity and credential state, and burns an agent turn on an ops failure.

AC #5 ("surfaces an honest error … never a silent drop") is satisfiable without that:

- `content` → a user-safe body ("couldn't process that voice note, could you send it as text?")
- `metadata.transcriptionError` → keeps `type` / `message` / `retryable` (already done, no change needed)
- add a `logger.error` at the adapter call site so it's visible in prod logs and health signals

That's the one genuinely architectural change I'd ask for. It's confined to this function plus the two adapter test assertions that currently check for `AUTH_FAILURE` / `PROVIDER_ERROR` in the body — those should move to asserting on `metadata.transcriptionError` instead.

src/channels/inbound-voice-note.ts
```diff
@@ -0,0 +1,144 @@
...
+  if (!contentType) return false;
+  const mime = contentType.split(';')[0]!.trim().toLowerCase();
+  return mime.startsWith('audio/') || mime === 'application/ogg';
+}
+
+export function isAudioAttachment(att: AudioAttachmentHint): boolean {
+  return att.isVoiceNote === true || isAudioContentType(att.contentType);
+}
+
+export function findFirstAudioAttachment<T extends AudioAttachmentHint>(
```
josephfung: **Should fix — only the first audio attachment is ever transcribed.**

A Signal message carrying two voice notes silently drops the second, and nothing logs it. At minimum log the count when `attachments` has more than one audio item; better, append a short "1 more voice note not transcribed" line to the content so the principal knows something was skipped rather than wondering why Curia only answered half.

src/channels/slack/message-converter.ts
```diff
@@ -105,11 +108,19 @@ function isIgnorableMessage(event: SlackMessageEvent): boolean {
   // Bot messages / automation — never treat as human inbound.
   if (event.bot_id || event.bot_profile) return true;
   // Subtypes: message_changed, message_deleted, channel_join, bot_message, etc.
-  // Only plain user messages (no subtype) are accepted.
-  if (event.subtype) return true;
+  // Plain user messages (no subtype) and file_share (voice notes, #1600) are accepted.
+  if (event.subtype && event.subtype !== 'file_share') return true;
```
josephfung: **Blocking — this widens inbound scope well beyond #1600.**

This accepts *every* `file_share` subtype, not just audio-bearing ones. A PDF or screenshot uploaded with a caption was previously ignored; now it publishes an inbound message where the agent sees the caption and has no idea a file is attached — confusing replies, new inbound volume, and behavior the PR body and spec don't mention.

Either narrow it:

```ts
if (event.subtype && !(event.subtype === 'file_share' && hasSlackAudio(event.files))) return true;
```

or keep the widening as a deliberate decision and document it in the PR description and `docs/specs/04-channels.md` (in which case the agent should also decide what the content looks like when a non-audio file is attached, because a bare caption isn't enough context).

src/channels/slack/slack-client.ts
```diff
@@ -233,4 +235,27 @@ export class SlackClient extends EventEmitter {
       return null;
     }
   }
+
+  /**
+   * Download a Slack-hosted file. Private URLs require the bot token as a
+   * Bearer Authorization header — they are not publicly fetchable (#1600).
+   */
+  async downloadFile(url: string): Promise<Uint8Array> {
+    const response = await fetch(url, {
```
josephfung: **Should fix — three hardening gaps on this fetch.**

1. **Host assertion.** We attach `Authorization: Bearer xoxb-…` to whatever URL we're handed. The URL comes from Slack's own event payload so it isn't attacker-controlled today, but asserting the host is `files.slack.com` before sending the bot token is cheap defense in depth — and the security review will ask about it anyway.
2. **No timeout.** A hung Slack response pins this indefinitely; the STT 60s self-bound doesn't cover the download. Pass an `AbortSignal.timeout(...)`.
3. **No size cap.** Same OOM concern as the Signal path. Slack sends `size` on the file object, so the cheap version is to check it before calling this (see my note on `types.ts`); a `content-length` guard here is a reasonable second layer.

josephfung (commented):
## Changes requested — good bones, four things to fix first

_(Posted as a review comment rather than a formal `REQUEST_CHANGES` state: GitHub won't let the PR author request changes on their own PR. Treat every **Blocking** item below as changes-requested.)_

The shape is right: one shared `inbound-voice-note.ts` policy module, per-channel download in the client layer, DI threaded through `src/index.ts`, and every one of #1600's six acceptance criteria has a test behind it. I verified the signal-cli integration against upstream — `getAttachment` is a real JSON-RPC method and `{ data: <base64> }` is the correct response shape (`GetAttachmentCommand.java` / `JsonAttachmentData.java`), and `--recipient`/`--group-id` really are a required mutually-exclusive pair. No architectural rework needed.

Four blocking items, all localized:

1. **No size cap before download** (`signal-rpc-client.ts`, `slack-client.ts`) — Signal allows 100 MB attachments; base64 makes that a ~133 MB single JSON-RPC line through the string-accumulating buffer. Given our prod OOM history this is the one I'd hold on.
2. **Untrusted senders can burn STT credits** (`signal-adapter.ts`) — 1:1 has no tier gate; any unknown or blocked sender triggers a paid Deepgram call per voice note.
3. **Raw provider errors reach the LLM and the sender** (`inbound-voice-note.ts`) — `AUTH_FAILURE: Deepgram STT request failed with HTTP 401` ends up in `content`, which the coordinator will paraphrase back to an external contact.
4. **Unflagged scope widening on Slack `file_share`** (`slack/message-converter.ts`) — every `file_share` subtype now publishes, not just audio-bearing ones.

Plus a handful of should-fix items inline (first-attachment-only, ACI recipient fallback, Slack transcode 404 retry, `downloadFile` host assertion + timeout).

### Two other notes, no inline anchor

**Ordering.** Both adapters dispatch inbound fire-and-forget (`void handle…().catch(…)`), so a voice note followed immediately by a text now publishes out of order — the voice path awaits download + STT while the text path doesn't. Probably acceptable, but it's new behavior and should be stated in the PR body rather than discovered later.

**AC #6 degradation.** "Degrades to existing text behavior when creds absent" is met in letter, but for an audio-only message that means the message is dropped entirely and the sender gets silence. The spec doc says this outright so it's a conscious call, but "Curia silently ignores your voice note" is a rough first impression for a self-host without Deepgram configured. Consider a one-line "voice notes aren't enabled here" reply.

### Things that are right and I checked

- `vitest.config.ts` includes `src/**/*.test.ts`, so the co-located tests are fine (matches `signal-rpc-client-calls.test.ts`).
- No `package.json` bump, correctly — the CHANGELOG entry sits under `[Unreleased]` with the rest of v0.43.
- Transcription runs after the Step 0 group-trust check, so group traffic is already protected. It's only the 1:1 path that isn't.

src/channels/slack/slack-adapter.ts
```diff
@@ -356,4 +379,60 @@ export class SlackAdapter implements Channel {
...
+        converted.content,
+        new Error('Slack audio file is missing a download URL'),
+      );
+      metadata.transcribedFromAudio = true;
+      return { content: resolved.content, metadata };
+    }
+
+    let resolved;
+    try {
+      const bytes = await this.config.client.downloadFile(url);
```
josephfung: **Should fix — Slack audio clips can still be transcoding when this fires.**

`url_private_download` can 404 briefly right after upload while Slack processes the clip, which would surface a spurious "download failed" note to the user for a perfectly good voice memo. One retry with a short backoff on 404/403 would eliminate that class of failure.

Worth a comment either way so the next reader knows it was considered.

src/channels/slack/slack-adapter.ts
```diff
@@ -356,4 +379,60 @@ export class SlackAdapter implements Channel {
...
+      return { content: converted.content, metadata };
+    }
+
+    const url = slackFileDownloadUrl(audio);
+    if (!url) {
+      const resolved = voiceNoteDownloadFailure(
+        converted.content,
+        new Error('Slack audio file is missing a download URL'),
+      );
+      metadata.transcribedFromAudio = true;
```
josephfung: **Minor — inconsistent with the other failure paths.**

This branch sets `metadata.transcribedFromAudio` but not `metadata.transcriptionError`, unlike the catch below. A missing download URL is a real failure and should be observable in metadata the same way, so downstream consumers don't have to special-case it.

src/channels/slack/types.ts
```diff
@@ -15,17 +15,33 @@ export interface SlackMessageEvent {
...
+
+/** Hosted Slack file metadata (voice notes use audio mimetypes). */
+export interface SlackFile {
+  id: string;
+  mimetype?: string;
+  filetype?: string;
+  name?: string;
+  title?: string;
+  url_private?: string;
+  url_private_download?: string;
```
josephfung: **Should fix — add `size?: number`.**

Slack sends `size` on file objects and it's what makes a pre-download size cap possible on this side (see my comment on `slack-client.ts`). Right now the Slack path has no way to know how big the file is before pulling it into memory.

src/channels/slack/types.ts
```diff
@@ -15,17 +15,33 @@ export interface SlackMessageEvent {
...
+  subtype?: string;
+  media_display_type?: string;
 }
 
 /** app_mention event payload. */
 export interface SlackAppMentionEvent {
   type: 'app_mention';
   user: string;
-  text: string;
+  text?: string;
```
josephfung: **Nit — type erosion for no gain.**

`app_mention` always carries text (it contains the `<@U…>` mention by definition), so `!rawContent` can never be true on that branch and this loosening buys nothing while making every downstream use of `mention.text` optional. I'd leave it `string`.

josephfung:
@coderabbitai review

src/channels/signal/message-converter.ts
```diff
@@ -127,6 +136,7 @@ export function convertSignalEnvelope(
       signalTimestamp: envelope.timestamp,
       groupId: groupInfo?.groupId,
       isGroup,
+      sourceUuid: envelope.sourceUuid || undefined,
```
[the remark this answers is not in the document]
josephfung: Fixed in a67d70c3. The converter now derives one non-empty identity (`sourceNumber` then `sourceUuid`), drops the envelope if neither is usable, and uses that identity for `senderId` and 1:1 conversation IDs. Converter and adapter tests now assert the published sender and conversation identities, not just the `getAttachment` recipient.

src/channels/slack/slack-client.ts
```diff
@@ -233,4 +266,47 @@ export class SlackClient extends EventEmitter {
...
+      if (Number.isFinite(contentLength) && contentLength > MAX_VOICE_NOTE_BYTES) {
+        try {
+          await response.body?.cancel();
+        } catch {
+          // Best-effort drain so undici does not pin the socket on error paths.
+        }
+        throw new Error('Slack file download refused: content-length exceeds voice-note cap');
+      }
+    }
+    const bytes = new Uint8Array(await response.arrayBuffer());
```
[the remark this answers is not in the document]
josephfung: Fixed in a67d70c3. `downloadFile` now streams `response.body` and aborts as soon as accumulated bytes exceed `MAX_VOICE_NOTE_BYTES`, so a chunked response without Content-Length cannot buffer past the cap. The existing `SlackFile.size` and Content-Length checks are unchanged. Added a test that an oversize chunked body is cancelled mid-stream.

josephfung:
@coderabbitai approve

Issue #1694, opened 2026-09-01
KG node identity is keyed on display name — same-name contacts cannot hold nodes, facts, or context


josephfung:
## Context

Split out of #1625 / ADR-039, which explicitly names this as the part it does **not** fix.

KG node identity is keyed on `(lower(label), type)` — `idx_kg_nodes_label` plus the
one-contact-per-node partial unique index `idx_contacts_kg_node_unique`. When two distinct
people share a display name, only the first contact gets a KG node: `createContact` catches
the `23505` on `idx_contacts_kg_node_unique` and retries without the link, leaving
`kg_node_id = NULL` by design.

That NULL is not a bug in `createContact` — inventing a node for the second contact
recreates the collision (rejected in the PR #1624 design review). It is a limitation of
keying entity identity on the display name.

#1625 removed *dedup exclusions*' dependency on this by moving them to
`contact_dedup_exclusions`, a table with no node prerequisite. Nothing else was fixed.

## The remaining problem

For every contact with `kg_node_id = NULL`, all KG-backed capability degrades silently:

- **Facts** — `entityMemory.storeFact` needs an `entityNodeId`. Facts about the second
  Seth Berman have nowhere to live, so nothing is remembered about them.
- **Relationships** — entity-to-entity edges need nodes on both sides.
- **Entity-context enrichment** — assembly returns nothing for these contacts, so agents
  operate on them with no background at all.
- **Silence** — none of the above errors. The contact simply has no memory, and there is
  no signal distinguishing "we know nothing about this person" from "this person cannot
  hold knowledge".

## Design pass findings

Two things surfaced during the design pass that were not in the original write-up.

### 0. Measured population (production, 2026-09-02)

Run after PR #1701 shipped, via `scripts/kg-node-linkage-report.ts`:

| | |
|---|---|
| contacts total | 548 |
| holding no KG node | 14 (2.6%) |
| arm A — org re-link | **0** |
| arm B — mint a node | **14** |
| shadowed by a same-name contact that has a node | 12 |

By kind, the nodeless 14 are 12 `person` and 2 `automated`. `organization` (38),
`principal` and `agent` are fully linked. Across the whole table there are 14 same-name
clusters and the largest holds 3 contacts.

**Migration `085` therefore inserts 14 rows**, and 12 of the 14 are the same-name case.
That is far smaller and better-bounded than assumed when ADR-040 was accepted, and it
re-sequences increment 2 — see finding 1.

### 1. There are two nodeless populations, not one

[...]

Issue #1695, opened 2026-09-01
ContactService.mergeContacts write sequence is not transactional — a mid-merge failure can strand identities, overrides, and dedup exclusions


josephfung:
## Context

Surfaced during the #1625 review (dedup exclusions moving onto `contact_dedup_exclusions`).

`ContactService.mergeContacts` performs its write sequence as five independent `pool.query`
calls with no `BEGIN`/`COMMIT`:

```ts
await this.backend.reattachIdentities(secondaryId, primaryId);
await this.backend.reattachAuthOverrides(secondaryId, primaryId);
await this.backend.reattachDedupExclusions(secondaryId, primaryId);   // added in #1625
await this.backend.updateContact(updatedPrimary);
await this.backend.deleteContact(secondaryId);
```

The catch logs `'Contact merge write failed — DB may be in partial state'` and rethrows, so
the error is not swallowed. But the state it leaves behind is genuinely inconsistent, and
the backend already exposes `withTransaction` — it just is not used here.

## Why it matters more since #1625

The sequence now moves **CEO decisions** in step 3. If `updateContact` or `deleteContact`
fails, the secondary contact is still alive while its exclusion rows have already been
re-pointed onto the primary. The next dedup sweep then sees the secondary again with none
of its "not the same person" rulings attached, and can re-propose — or, for a structural
pair, auto-merge — a pair the CEO explicitly ruled apart.

The same window has always existed for identities and auth overrides:

- Identities re-pointed, secondary not deleted → the secondary is left with no channel
  identities and is unreachable/unresolvable.
- Auth overrides re-pointed, primary not updated → granted permissions land on a contact
  whose golden record was never written.

## Proposal

Wrap the write sequence in a single transaction using the existing
`ContactServiceBackend.withTransaction`, threading the client through each backend call
(several already accept an optional `client?: DbPoolClient`; `reattachIdentities`,
`reattachAuthOverrides`, `reattachDedupExclusions`, and `deleteContact` do not yet).

Out of scope for the transaction: the KG `mergeEntities` call above it, which is already
documented as best-effort and non-fatal, and the `onContactMerged` bus publish below it,
which must not run until the commit succeeds.

## Acceptance criteria

- [ ] `mergeContacts` executes `reattachIdentities`, `reattachAuthOverrides`,
      `reattachDedupExclusions`, `updateContact`, and `deleteContact` on one client inside
      one transaction, rolling back as a unit.
[...]

Issue #1647, opened 2026-08-10
scheduler: jobs stuck in 'running' — #801 fix missed the tool_use-exhausted and defensive-else error paths in runtime.ts


josephfung:
## Summary

The fix for #801 (jobs stuck in `running` when the runtime emits `agent.response({isError:true})` **without** a paired `agent.error`) only covered **one** of the three error-response paths in `AgentRuntime`. Two paths still set `isResponseError = true` and publish an error response **without** publishing `agent.error`, so a scheduler-originated task that hits either path stays wedged in `running` until the watchdog reclaims it (up to ~45–90 min later). This is a partial regression / gap-completion of #801.

## Background — how completion is signalled

The scheduler tracks completion via two bus subscribers in [`src/scheduler/scheduler.ts`](../blob/main/src/scheduler/scheduler.ts#L162-L191):

- `agent.response` subscriber **skips error responses**: `if (responseEvent.payload.isError) return;` (scheduler.ts:168)
- `agent.error` subscriber is what calls `handleCompletion()` on failure (scheduler.ts:180-191), which flips the job row out of `running` and deletes the `Scheduler.pendingJobs` / `Scheduler.burstCounts` entries.

So the invariant is: **every failure must publish `agent.error`.** The `agent.response(isError:true)` alone is deliberately ignored (to avoid a double `handleCompletion`). If a path emits the error *response* but not the error *event*, the job is never completed by either subscriber.

## The gap

In [`src/agents/runtime.ts`](../blob/main/src/agents/runtime.ts#L1829-L1939), three branches set `isResponseError = true` and fall through to publish `agent.response({ isError: true, parentEventId })` at lines 1929-1939:

1. **`tool_use` branch — runtime.ts:1829-1838** (`"Tool loop exhausted without a text response"` / `"LLM returned tool_use but no execution layer configured"`): sets `isResponseError = true`, **no `publishAgentError()`**. ❌
2. **Defensive `else` branch — runtime.ts:1917-1922** (`"LLM call failed after retries"`): sets `isResponseError = true`, **no `publishAgentError()`**. ❌
3. **Empty-text recovery-failed branch — runtime.ts:1877-1913**: sets `isResponseError = true` **and** calls `publishAgentError()` first (lines 1888-1904), with a comment explicitly describing this exact trap. ✅ — this is the path #801 fixed.

Paths (1) and (2) were missed by #801. They reproduce the original bug.

## Impact

[...]

PR #1703, opened 2026-09-02
fix(entity-context): surface transient DB failures in `failed` bucket, not `unresolved`

### LINKED ISSUE #1702: entity-context reports a DB failure as "contact not found" — errors share the `unresolved` bucket
Surfaced by review of PR #1701 (#1694 increment 1). Not introduced by that PR — pre-existing, but #1701 codified it in a header comment, so it should be written down rather than left implicit.

## Problem

`EntityContextAssembler.assembleMany` catches per-ID failures and pushes the ID into `unresolved`:

```ts
} catch (err) {
  this.logger.error({ err, entityId: id }, 'entity-context: assembleOne failed — treating as unresolved');
  unresolved.push(id);
}
```

`skills/entity-context` then hands that bucket to the LLM described as "IDs that could not be found in the KG or contacts table". Two categorically different answers share one label, inside `success: true`.

## Failure scenario

An agent calls `entity-context` with three contact UUIDs. Mid-batch the pool hits `53300 too_many_connections` (or a statement timeout, or `ECONNRESET` during a failover). `assembleOneClassified` throws, `assembleMany` logs at error and pushes the UUID into `unresolved`. The handler's outer `catch` never fires, because `assembleMany` does not throw.

The agent receives `{ success: true, entities: [2 items], unresolved: ['<real-uuid>'] }` and tells the CEO there is no record of that person. The correct action is to retry; the agent has no way to know that.

This is the same "absence of evidence read as evidence of absence" failure that #1694 exists to remove, applied to the error path instead of the nodeless path. PR #1701 fixed it for nodeless contacts and left it standing for transient failures.

## Suggested shape

A fourth bucket, symmetric with the `nodeless` one that just landed:

- `failed: { inputId, reason }[]` on `AssembleManyResult`, disjoint from `unresolved`.
- Surfaced by the `entity-context` handler with a reason saying the lookup failed and can be retried, and that it is not evidence the contact is unknown.
- Emitted before `entities` in the result object, for the same head-slice truncation reason as `nodeless`.
- A `warn` in the handler when non-empty.

## Acceptance criteria

- [ ] A DB error during `assembleMany` puts the ID in a `failed` bucket, not `unresolved`
- [ ] `unresolved` contains only IDs that genuinely matched no contact and no KG node
- [ ] The `entity-context` skill surfaces `failed` with a retryable-failure reason distinct from both `unresolved` and `nodeless`
- [ ] Tests cover a DB error mid-batch producing partial `entities` plus a populated `failed`, with `unresolved` empty
- [ ] `docs/specs/11-entity-context-enrichment.md` documents the full bucket contract

## Related

- #1694 / ADR-040, PR #1701 (the `nodeless` bucket this mirrors)

josephfung:
## Summary

Fixes #1702. When `assembleMany` hit a transient DB error mid-batch, the errored ID landed in `unresolved` — indistinguishable from "contact not found." This adds a `failed` bucket (symmetric with `nodeless` from #1694) with retry classification owned by the assembler and LLM-facing wording owned by the handler.

## Changes

- `EntityContextAssembler.assembleMany` returns `failed: { inputId, retryable }[]`; permanent errors (e.g. `42703`) are not classified retryable
- `entity-context` handler surfaces `failed` before `entities`, warns with counts (not raw emails), and returns `success: false` on total failure
- `entity_enrichment` logs failed lookups on their own warn line (with dormant-branch note)
- `tool.json` bumped to 1.2.0; spec 11 documents the four-bucket contract and total-failure behaviour
- Tests: non-retryable classification, all four buckets in one batch, handler warn line, `makeSequentialPool` reject steps

## Checklist

- [x] CI passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 - **`EntityContextAssembler.assembleMany`** — new `nodeless` bucket, disjoint from `unresolved`; logged at warn. (#1694)
 - **`entity-context`** — output gains `nodeless`; agents read a missing node as a capability gap. (#1694)
 - **`entity-context` output schema** — new `nodeless` key on `ToolResult.data` (public API surface). (#1694)
+- **`entity-context` output schema** — new `failed` key on `ToolResult.data` (public API surface). (#1702)
 - **`entity_enrichment`** — nodeless contacts get their own warn line instead of joining the unresolved-IDs line. (#1694)
 - **`contact-dedup-exclude`** — writes the exclusions table; outputs `created`, no `entityMemory` capability. (#1625)
 - **`contact-find-duplicates`** — loads exclusions in one query; `entityMemory` capability dropped. (#1625)
@@ -54,6 +55,8 @@ bus event types) are noted explicitly even in the `0.x` range.
 ### Fixed
 
 - **Signal inbound** — ACI-only senders keep a non-empty sender and conversation identity. (#1600)
+- **`entity-context`** — transient DB failures surface in `failed`, not `unresolved`. (#1702)
+- **`entity_enrichment`** — failed lookups get their own warn line instead of joining unresolved. (#1702)
 - **Slack file download** — abort streamed bodies once they exceed the voice-note size cap. (#1600)
 - **Contact merge** — merge writes run in one transaction; a failure no longer strands identities or dedup rulings. (#1695)
 - **Signal voice calls** — duplicate `RINGING_INCOMING` for one call no longer self-rejects it as busy. (#1672)
--- a/docs/specs/11-entity-context-enrichment.md
+++ b/docs/specs/11-entity-context-enrichment.md
@@ -214,7 +214,7 @@ A new skill that assembles the full context payload for one or more entities. Av
 
 Diagnostic buckets (`nodeless`, `failed`, `unresolved`) are emitted before `entities` in the skill result so head-slice output truncation sacrifices bulk entity data rather than the warnings.
 
-When every requested ID lands in `failed` and `entities` is empty, the skill returns `success: false` (not a success-shaped empty result with retry hints) so the agent does not loop against a hard-down database.
+When every requested ID lands in `failed` and `entities`, `unresolved`, and `nodeless` are all empty, the skill returns `success: false` (not a success-shaped empty result with retry hints) so the agent does not loop against a hard-down database. Batches mixing `failed` with `unresolved` or `nodeless` still return `success: true` with all diagnostic buckets.
 
 **Resolution priority for each ID:**
 1. If the ID matches a `contacts.id` with a non-null `kg_node_id` → assemble from that KG node
--- a/docs/specs/11-entity-context-enrichment.md
+++ b/docs/specs/11-entity-context-enrichment.md
@@ -196,16 +196,33 @@ A new skill that assembles the full context payload for one or more entities. Av
   },
   "outputs": {
     "entities": "EntityContext[]",
-    "unresolved": "string[] (IDs that couldn't be found)"
+    "unresolved": "string[] (IDs that matched no contact and no KG node)",
+    "nodeless": "{inputId, contactId, displayName, reason}[]? (contacts that exist but have no stored profile — cannot hold facts; not evidence of absence)",
+    "failed": "{inputId, reason}[]? (lookups that errored — retry may succeed; not evidence of absence)"
   }
 }
 ```
 
+**Result buckets** — `assembleMany` and the skill surface four disjoint outcomes per input ID:
+
+| Bucket | Meaning | Agent action |
+|---|---|---|
+| `entities` | Full `EntityContext` assembled | Use the payload |
+| `unresolved` | ID matched no contact and no KG node | Treat as genuinely unknown |
+| `nodeless` | Contact exists but holds no KG node (#1694) | Report capability gap; do not store facts or infer absence |
+| `failed` | Assembly threw (#1702) | If reason says retryable: retry. If not: report failure. **Not** evidence the contact is unknown |
+
+Diagnostic buckets (`nodeless`, `failed`, `unresolved`) are emitted before `entities` in the skill result so head-slice output truncation sacrifices bulk entity data rather than the warnings.
+
+When every requested ID lands in `failed` and `entities` is empty, the skill returns `success: false` (not a success-shaped empty result with retry hints) so the agent does not loop against a hard-down database.
+
 **Resolution priority for each ID:**
-1. If the ID matches a `contacts.id` → look up `kg_node_id` from the contact, then assemble from the KG node
-2. If the ID matches a `kg_nodes.id` directly → assemble from the KG node
-3. Special values: `"caller"` → resolve to `ctx.caller.contactId`, `"agent"` → resolve to the agent's contactId
-4. If no match → include in the `unresolved` array (not a hard error)
+1. If the ID matches a `contacts.id` with a non-null `kg_node_id` → assemble from that KG node
+2. If the ID matches a `contacts.id` with a null `kg_node_id` → include in `nodeless` (contact exists but cannot hold knowledge)
+3. If the ID matches a `kg_nodes.id` directly → assemble from that KG node
+4. Special values: `"caller"` → resolve to `ctx.caller.contactId`, `"agent"` → resolve to the agent's contactId
+5. If no match → include in the `unresolved` array (not a hard error)
+6. If assembly throws → include in `failed` with a retry classification (not `unresolved`)
 
 ### Assembly Pipeline
 
--- a/skills/entity-context/handler.ts
+++ b/skills/entity-context/handler.ts
@@ -8,11 +8,13 @@
 // entity_enrichment manifest declaration instead — that runs the same
 // assembler without an extra LLM round-trip.
 //
-// Three buckets come back and all three are surfaced: entities that assembled,
-// `unresolved` IDs that matched nothing, and `nodeless` contacts that exist but
-// cannot hold knowledge (#1694 / ADR-040).
+// Four buckets come back and all are surfaced: entities that assembled,
+// `unresolved` IDs that matched nothing, `nodeless` contacts that exist but
+// cannot hold knowledge (#1694 / ADR-040), and `failed` IDs whose lookup errored
+// (#1702). The handler owns all LLM-facing wording for `nodeless` and `failed`.
 
 import type { ToolHandler, ToolContext, ToolResult } from '../../src/skills/types.js';
+import type { AssembleManyFailedEntry, AssembleManyResult } from '../../src/entity-context/assembler.js';
 
 // Shown to the LLM for every contact that resolved but holds no KG node.
 //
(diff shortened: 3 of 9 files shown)


src/entity-context/assembler.ts
```diff
@@ -685,3 +698,23 @@ function isContactRowColumnPopulated(
     case 'birthday':      return row.birthday != null;
   }
 }
+
+/**
+ * Produce a safe, retry-oriented reason for the LLM when assembly throws.
+ * Must not expose table names, SQL state codes, or other DB internals.
+ */
+function formatAssemblyFailureReason(err: unknown): string {
+  const base =
```
josephfung: **Every error is labeled "transient" and told to retry.**

`formatAssemblyFailureReason` special-cases `53300` and `57014`, then falls through to *"failed due to a transient error. Retry the lookup"* for everything else. But the `catch` it feeds also swallows permanent failures:

- `42703` undefined_column (migration drift, which this repo has hit before)
- `42P01` undefined_table
- `42501` insufficient_privilege
- a plain `TypeError` thrown anywhere in the assembly pipeline

None of those get better on retry, and the LLM is now being instructed to keep trying.

That is the same shape as the bug this PR fixes, rotated: #1702 is about a wrong label producing a wrong agent action. This one produces a retry loop instead of a false "no record".

The two claims in the reason string are independent and should be separated:
- *"this is not evidence the entity is unknown"* - always true
- *"retry the lookup"* - only sometimes true

Suggested shape: have the assembler return a classification rather than prose, e.g. `failed: { inputId, retryable: boolean }` (or an error class), and let the handler own the LLM-facing wording. That also resolves the duplicated-string problem flagged in `handler.ts`.

skills/entity-context/handler.ts
```diff
@@ -30,6 +31,12 @@ const NODELESS_REASON =
   + 'relationships, or background context. Having no context here does not mean the '
   + 'contact is unknown to us. Do not try to store facts about them; report the gap instead.';
 
+// Shown to the LLM for every ID whose assembly threw. Wording is load-bearing:
+// a failed lookup is not evidence of absence — the correct action is to retry.
+const FAILED_REASON =
```
josephfung: **Duplicated reason string across two layers.**

`FAILED_REASON` here is a verbatim copy of the `base` string in `formatAssemblyFailureReason` in `src/entity-context/assembler.ts`. Two copies of the same LLM-facing prose in different files will drift.

Note also the asymmetry with `nodeless`: `NODELESS_REASON` lives in the handler because `NodelessContact` carries no reason field, so the presentation layer owns the wording. This PR puts the equivalent prose in the assembler instead. Pick one layer and be consistent.

My preference is the classification approach suggested on `assembler.ts`: the assembler returns `retryable` (it is the only place that has the error), and the handler owns all wording. That collapses both problems into one fix.

skills/entity-context/handler.ts
```diff
@@ -98,6 +113,14 @@ export class EntityContextHandler implements ToolHandler {
                 })),
               }
             : {}),
+          ...(result.failed.length > 0
+            ? {
+                failed: result.failed.map(f => ({
+                  inputId: f.inputId,
+                  reason: f.reason || FAILED_REASON,
```
josephfung: `f.reason || FAILED_REASON` is a fallback that cannot fire - `assembleMany` always populates `reason` for every entry it pushes into `failed`. Untested dead code on a defensive branch. Either drop the fallback, or (better) move the wording here entirely per the comment on `FAILED_REASON` above.

josephfung (commented):
## Review: draft to final

CI is fully green (11/11), the design is right, and every acceptance criterion in #1702 is met. Backfilling `nodeless` into spec 11 (which #1701 missed) was the right call, and the key-ordering test that pins `failed` ahead of `entities` is the correct way to defend a truncation-order comment.

Below is everything that should be addressed before this comes out of draft. Inline comments cover the code; the process items are here because they have no line to anchor to.

### Substantive

**1. Every error is labeled "transient" and told to retry.** See inline on `src/entity-context/assembler.ts`. This is the highest-priority item.

**2. Total failure still returns `success: true`.** If every ID errors, the handler returns `{ success: true, entities: [], unresolved: [], failed: [...] }`. Not a regression (pre-PR it returned success with everything in `unresolved`), but combined with item 1 it needs a decision: the agent's error budget counts errors, not futile successes, so a hard-down DB plus a "retry" hint has nothing to stop the loop. Either return `success: false` when `entities` is empty and `failed` is non-empty, or write down in spec 11 why a partial-shaped result with zero partials is still a success.

**3. Duplicated reason string across two layers.** See inline on `skills/entity-context/handler.ts`.

### Correctness in the non-code files

**4. Duplicate `### Fixed` heading in CHANGELOG.** See inline.

**5. Missing the public-API-surface line.** #1694 recorded `nodeless` under `### Changed` as "new `nodeless` key on `ToolResult.data` (public API surface)". `failed` is the same kind of addition to the same surface but appears only under Fixed. The changelog preamble calls out `tool.json` / `ToolContext` surfaces explicitly, so follow the #1694 precedent and add the matching Changed line.

**6. Spec 11's resolution-priority list was not updated.** The new four-bucket table is good, but eight lines below it step 4 still reads "If no match -> include in the `unresolved` array", and step 1 does not mention the nodeless outcome. The list and the table now disagree about what a lookup can produce. See inline.

### Repo conventions (these are the actual draft-to-final blockers)

**7. Commit attribution.** The single commit is `Signed-off-by: Cursor Agent <cursoragent@cursor.com>` with a `Co-authored-by: Joseph Fung` trailer. Repo rules are: no `Co-Authored-By` trailers, and the DCO sign-off must be the human author. Compare a003df2f (#1701), which is authored and signed off by Joseph with no co-author. Rewrite the commit author and sign-off, and drop the trailer.

**8. Cursor promo footer in the PR body.** The "Open in Web / Open in Cursor" badge block is agent-tool attribution in a PR description, which the no-credits rule names explicitly. Strip it.

**9. PR metadata.** Still draft; no labels (#1702 carries `bug`, `P2`, `memory-context`); no milestone (#1702 is on v0.43 - Hardening). The checklist's "CI passes" is unchecked though all checks are green. The body says "Fixes #1702" in Summary and "Closes #1702" twice more - keep one.

### Test gaps to close before formal review

- **No test that a non-transient error is described differently.** Comes free with the fix for item 1, and it is the case most likely to regress.
- **No single batch exercising all four buckets at once.** Spec 11 now claims four disjoint outcomes; nothing proves disjointness in one call. `tests/unit/entity-context/assembler.test.ts:288` has the mixed-batch nodeless test to extend.
- **The handler's new warn line is untested.** The `entity_enrichment` equivalent is covered; the handler one is not.
- **The mid-batch test hand-rolls a call-counter mock** instead of using the file's `makeSequentialPool`. See inline - style, not correctness.

### Minor

- The new `failed` branch in `src/skills/execution.ts` omits the "no tool.json declares `entity_enrichment`, so this branch is dormant" note its `nodeless` neighbour carries. See inline.
- The handler's warn logs `inputIds`, which can be raw email addresses, while the assembler's `nodeless` warn deliberately logs only `contactId`. Low stakes (the pre-existing assembler error log already includes the id), but it is a new PII surface at the skill layer.

CHANGELOG.md
```diff
@@ -33,6 +33,11 @@ bus event types) are noted explicitly even in the `0.x` range.
 - **`contact_dedup_exclusions`** — dedup exclusions get a table; migration 084 backfills KG facts. (#1625)
 - **ADR-039 (Accepted)** — decisions belong in the relational ledger, learned knowledge in the KG. (#1625)
 
+### Fixed
```
josephfung: **Duplicate `### Fixed` heading.**

`## [Unreleased]` already has a `### Fixed` section further down (currently line 53 on `main`, holding the #1600 / #1695 / #1672 / #1647 / #1663 entries). This inserts a second one between `### Added` and `### Changed`.

Post-merge the section reads: Added / Fixed / Changed / Fixed / Security. Keep a Changelog also orders Fixed after Changed.

Move both new entries into the existing `### Fixed` section and drop this heading.

docs/specs/11-entity-context-enrichment.md
```diff
@@ -196,11 +196,24 @@ A new skill that assembles the full context payload for one or more entities. Av
...
+**Result buckets** — `assembleMany` and the skill surface four disjoint outcomes per input ID:
+
+| Bucket | Meaning | Agent action |
+|---|---|---|
+| `entities` | Full `EntityContext` assembled | Use the payload |
+| `unresolved` | ID matched no contact and no KG node | Treat as genuinely unknown |
+| `nodeless` | Contact exists but holds no KG node (#1694) | Report capability gap; do not store facts or infer absence |
+| `failed` | Assembly threw (connection limit, timeout, `ECONNRESET`, etc.) (#1702) | Retry; **not** evidence the contact is unknown |
+
+Diagnostic buckets (`nodeless`, `failed`, `unresolved`) are emitted before `entities` in the skill result so head-slice output truncation sacrifices bulk entity data rather than the warnings.
```
josephfung: The four-bucket table is good, but the **Resolution priority for each ID** list immediately below it was not updated and now contradicts the table:

- Step 1 ("If the ID matches a `contacts.id` -> look up `kg_node_id` ... then assemble") does not mention that a contact with a NULL `kg_node_id` produces `nodeless`, not an assembled entity.
- Step 4 ("If no match -> include in the `unresolved` array") is the only stated outcome for a non-happy path, with no mention of `failed`.

A reader who stops at the numbered list comes away with the pre-#1694 two-bucket model. Worth two small edits so the list and the table agree.

src/skills/execution.ts
```diff
@@ -1622,6 +1622,12 @@ export class ExecutionLayer {
               'entity_enrichment: contacts have no KG node — skill ran without their context',
             );
           }
+          if (enrichmentResult.failed.length > 0) {
```
josephfung: The `nodeless` branch directly above carries a note explaining that no `tool.json` currently declares `entity_enrichment`, so the branch is dormant but deliberately kept. This new branch has the same property and no such note - a future reader will wonder why it appears untriggered and may delete it as dead code. Worth one line pointing at the neighbouring explanation.

Minor, but also note the log-field naming across the three branches here: `unresolved`, `contactIds`, `inputIds`. Consistent within the new code, inconsistent across the block.

tests/unit/entity-context/assembler.test.ts
```diff
@@ -561,14 +561,55 @@ describe('EntityContextAssembler', () => {
...
+          inputId: 'jenna@example.com',
+          reason: expect.stringMatching(/retry/i),
+        }),
+      ]);
+    });
+
+    it('returns partial entities plus failed when a mid-batch DB error occurs', async () => {
+      const pool = makeMockPool();
+      let callIndex = 0;
+      vi.mocked(pool.query).mockImplementation(() => {
```
josephfung: This hand-rolls a call-counter mock with an inline six-response array, duplicating (and diverging from) the file's existing `makeSequentialPool` helper at the top.

`makeSequentialPool` throws on an unexpected query count specifically so query-shape regressions fail loudly with a clear message. This version instead falls through to the rejection branch, so a change in `assembleOneClassified`'s query count fails the test with a confusing assertion mismatch rather than "add a response to makeSequentialPool".

Style, not correctness - it does still fail either way. Cleaner fix: teach `makeSequentialPool` to accept an error entry in the response queue, and use it here.

Separately, this describe block would be the natural home for the two missing tests:
1. A non-transient error (e.g. `42703`) not being described as retryable, once that classification exists.
2. One batch producing all four buckets at once, proving the disjointness spec 11 now claims.

skills/entity-context/handler.ts
```diff
@@ -30,6 +32,49 @@ const NODELESS_REASON =
...
+  }
+  return `Entity context lookup failed due to a system error. Do not retry; report the failure. ${NOT_UNKNOWN_SUFFIX}`;
+}
+
+function formatTotalFailureError(failed: AssembleManyFailedEntry[]): string {
+  const retryable = failed.some(f => f.retryable);
+  if (retryable) {
+    return `Entity context lookup failed for all requested IDs due to transient errors. Retry; ${NOT_UNKNOWN_SUFFIX}`;
+  }
+  return `Entity context lookup failed for all requested IDs due to a system error. Do not retry; report the failure. ${NOT_UNKNOWN_SUFFIX}`;
```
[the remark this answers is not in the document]
josephfung: Good catch — `formatTotalFailureError` now counts retryable vs non-retryable separately and returns a mixed message when both are present (retry only the transient ones). Added `does not claim all failures were transient when retryability is mixed` in `6892e762`.

skills/entity-context/handler.ts
```diff
@@ -69,10 +114,26 @@ export class EntityContextHandler implements ToolHandler {
...
+          { failedCount: result.failed.length, retryableCount },
+          'entity-context: some lookups failed — see failed bucket in result',
+        );
+      }
+
+      // Total failure: every ID errored and nothing assembled. A success-shaped
+      // result with only a retry hint has nothing to stop the agent retrying into
+      // a hard-down DB (#1702 review).
+      if (result.entities.length === 0 && result.failed.length > 0) {
+        return { success: false, error: formatTotalFailureError(result.failed) };
```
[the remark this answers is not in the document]
josephfung: Agreed — introduced `isTotalFailure()` which requires `failed` non-empty and `entities`, `unresolved`, and `nodeless` all empty before returning `success: false`. Batches mixing failed with unresolved or nodeless stay `success: true` with all buckets surfaced. Tests added in `6892e762`.

josephfung:
@coderabbitai approve

PR #1709, opened 2026-09-02
fix(extract-facts): keep raw subject names out of remaining fact-loop logs

### LINKED ISSUE #1706: extract-facts logs raw `subject` names at ten pre-existing sites, against its own stated rule
Surfaced by CodeRabbit while reviewing PR #1705. That PR fixed the two log lines it introduced; these ten predate it and were left alone for scope.

## Problem

`skills/extract-facts/handler.ts` states the rule itself, at the malformed-fact warn:

```ts
// Log only field types — never the raw values (subject/attribute/value may contain PII).
subjectType: typeof fact?.subject,
```

and there is a test pinning it (`malformed-fact warn logs only structural metadata — not the raw fact object (PII guard)`).

Ten other log calls in the same file ignore that rule and log `subject` (and often `attribute`) raw. `subject` is an entity name the LLM extracted from a conversation transcript — in practice usually a person's name.

`src/logger.ts` redacts `password`, `token`, `secret`, `api_key`, `senderId`, `email`, `senderEmail`, `phoneNumber`. It does not cover `subject`, so these values reach the log transport verbatim. In production that is the container log; in dev it is `curia.log`.

## Sites

All in `skills/extract-facts/handler.ts` (line numbers as of `5c1fa97e`):

| line | call |
|---|---|
| 285 | `warn` — canonical attribute normalization failed |
| 303 | `warn` — findContactByKgNodeId failed |
| 315 | `info` — canonical attribute redirected |
| 326 | `warn` — canonical redirect validation failed |
| 333 | `error` — canonical redirect unexpected error |
| 376 | `info` — fact stored |
| 387 | `warn` — fact not stored (conflict) |
| 394 | `warn` — fact not stored |
| 408 | `error` — programming error in fact loop |
| 414 | `error` — failed to persist fact |

`attribute` travels with `subject` in most of them. `attribute` is a snake_case key (`home_city`, `job_title`) and is not itself identifying, but `subject` + `attribute` + a timestamp is.

## Suggested fix

Same substitution PR #1705 used on the ambiguity path: log the resolved node id (`entityNode.id`) rather than the name. It identifies the same entity, is already logged in several of these calls, and is more useful for diagnosis than the string — an operator can look the node up. Where no node has been resolved yet, `subjectType` / presence flags are enough.

Alternatively, add `subject` to the redact list in `src/logger.ts` — cheaper, but it blunts every legitimate use of the field name elsewhere and leaves the call sites looking correct when they are not. The node-id substitution is the better fix.

## Acceptance criteria

- [ ] None of the ten call sites above logs a raw `subject` value
- [ ] Each still identifies the entity well enough to diagnose the log line (node id where one is resolved)
- [ ] A PII-guard test per affected path, matching the existing malformed-fact and ambiguity-path guards
- [ ] `attribute` is reviewed alongside — kept where it carries no identity, dropped where it pairs with one

## Related

- PR #1705 (#1694 increment 2) — fixed the two lines it added and filed this for the rest

josephfung:
## Summary

`extract-facts` already refused to log raw `subject` on the malformed-fact and ambiguity paths — it is an LLM-extracted entity name, usually a person's name, and pino does not redact it. Ten older fact-loop call sites still logged it. Surfaced by CodeRabbit on PR #1705; that PR fixed the two lines it added and filed this for the rest.

This applies the same substitution: log `entityNodeId` where a node is resolved. `attribute` stays — it is a snake_case key and is not identifying once unpaired from the name.

Review follow-up: `buildCanonicalPatch` no longer interpolates the raw phone into `reason` (stable code `phone_normalization_failed`). Remaining callee/sibling leaks (`resolveOrCreate`, `addAlias`, `memory-store`) are tracked in #1713.

## Changes

- Replace `subject` with `entityNodeId` on the ten remaining fact-loop log sites in `skills/extract-facts/handler.ts`
- Hoist `entityNode` as `KgNode | undefined` so the catch can log the id when resolution completed
- `buildCanonicalPatch` phone fallback reason is a stable code, not the raw number
- Add a PII-guard test per affected path (Error.message folded into the serializer)
- Patch-bump `extract-facts` to 1.3.1

## Related Issues

Closes #1706

## Checklist

- [x] Spec docs updated if behavior changes — N/A (log hygiene only)
- [x] CI passes

### DIFF
--- a/skills/extract-facts/handler.test.ts
+++ b/skills/extract-facts/handler.test.ts
@@ -587,8 +587,8 @@ describe('ExtractFactsHandler', () => {
   // These ten pre-existing call sites still logged it. `subject` is an LLM-extracted
   // entity name (usually a person's name) and the pino redact list does not cover it.
   // Substitute: entityNodeId where a node is resolved; omit it when resolution has
-  // not completed. `attribute` is a snake_case key and is kept — it is not identifying
-  // on its own.
+  // not completed. `attribute` is logged only when it is snake_case — the LLM can
+  // put a name in that field, and pino does not redact it.
   describe('fact-loop logs never carry the raw subject name (PII guard)', () => {
     const PII_SUBJECT = 'Priya Ramanathan';
     const UNNORMALIZABLE_PHONE = '020 7946 0958';
@@ -834,6 +834,29 @@ describe('ExtractFactsHandler', () => {
       const nodes = await entityMemory.findEntities(PII_SUBJECT);
       expect(call![0]).toHaveProperty('entityNodeId', nodes[0]!.id);
       expect(call![0]).toHaveProperty('attribute', 'dietary_preference');
+    });
+
+    it('does not log a non-snake_case attribute (PII guard)', async () => {
+      // The extraction prompt asks for snake_case keys, but the model can ignore
+      // that and put a person's name in `attribute`. pino does not redact it.
+      const entityMemory = makeEntityMemory();
+      vi.spyOn(entityMemory, 'storeFact').mockResolvedValueOnce({
+        stored: false,
+        action: 'conflict',
+        conflict: 'contradicts existing value: London',
+      });
+      const ctx = makeCtx(
+        entityMemory,
+        { text: `${PII_SUBJECT} is vegetarian.`, source: 'test' },
+        makeMockInfraLlm(['yes', factJson(PII_SUBJECT, 'vegetarian')]),
+      );
+      const log = attachSpyLog(ctx);
+
+      await new ExtractFactsHandler().execute(ctx);
+
+      const call = findCall(log.warn, /fact not stored/);
+      expectNoPii(call);
+      expect(call![0]).not.toHaveProperty('attribute');
     });
 
     it('programming-error in fact loop', async () => {
--- a/skills/extract-facts/handler.test.ts
+++ b/skills/extract-facts/handler.test.ts
@@ -586,19 +586,23 @@ describe('ExtractFactsHandler', () => {
   // The malformed-fact and ambiguity paths already omit the raw subject name.
   // These ten pre-existing call sites still logged it. `subject` is an LLM-extracted
   // entity name (usually a person's name) and the pino redact list does not cover it.
-  // Substitute: entityNodeId where a node is resolved; presence flags otherwise.
-  // `attribute` is a snake_case key and is kept — it is not identifying on its own.
+  // Substitute: entityNodeId where a node is resolved; omit it when resolution has
+  // not completed. `attribute` is a snake_case key and is kept — it is not identifying
+  // on its own.
   describe('fact-loop logs never carry the raw subject name (PII guard)', () => {
     const PII_SUBJECT = 'Priya Ramanathan';
+    const UNNORMALIZABLE_PHONE = '020 7946 0958';
 
     function spyLog() {
-      return {
+      const log = {
         warn: vi.fn(),
         info: vi.fn(),
         debug: vi.fn(),
         error: vi.fn(),
         child: vi.fn(),
       };
+      log.child.mockReturnValue(log);
+      return log;
     }
 
     function attachSpyLog(ctx: ToolContext) {
@@ -611,9 +615,24 @@ describe('ExtractFactsHandler', () => {
       return fn.mock.calls.find(c => messageRe.test(String(c[c.length - 1])));
     }
 
-    function expectNoRawSubject(call: unknown[] | undefined) {
+    // JSON.stringify skips Error.message (non-enumerable). Pino does serialize it,
+    // so fold err.message/stack in or the guard is inert on the three `err` sites.
+    function serializeLogPayload(payload: unknown): string {
+      const obj = payload as Record<string, unknown> | null;
+      const err = obj?.err;
+      const errText = err instanceof Error
+        ? `${err.name} ${err.message} ${err.stack ?? ''}`
+        : '';
+      return `${JSON.stringify(payload)}\n${errText}`;
+    }
+
+    function expectNoPii(call: unknown[] | undefined, extraForbidden: string[] = []) {
       expect(call).toBeDefined();
-      expect(JSON.stringify(call![0])).not.toContain(PII_SUBJECT);
+      const serialized = serializeLogPayload(call![0]);
+      expect(serialized).not.toContain(PII_SUBJECT);
+      for (const s of extraForbidden) {
+        expect(serialized).not.toContain(s);
+      }
       expect(call![0]).not.toHaveProperty('subject');
     }
 
@@ -638,8 +657,8 @@ describe('ExtractFactsHandler', () => {
       const entityMemory = makeEntityMemory();
       const ctx = makeCtx(
         entityMemory,
-        { text: `${PII_SUBJECT} phone is not-a-phone-number.`, source: 'test' },
-        makeMockInfraLlm(['yes', factJson('phone', 'not-a-phone-number')]),
+        { text: `${PII_SUBJECT} phone is ${UNNORMALIZABLE_PHONE}.`, source: 'test' },
+        makeMockInfraLlm(['yes', factJson('phone', UNNORMALIZABLE_PHONE)]),
         makeContactService(),
       );
       const log = attachSpyLog(ctx);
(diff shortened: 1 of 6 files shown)


skills/extract-facts/handler.ts
```diff
@@ -284,7 +286,9 @@ ${text}`,
             if (patch !== null) {
               if (patch.fallbackToKg) {
                 ctx.log.warn(
-                  { subject, attribute, reason: patch.reason },
+                  // entityNodeId, not subject: the LLM-extracted name is PII and the
+                  // pino redact list does not cover it. attribute is a snake_case key.
+                  { entityNodeId: entityNode.id, attribute, reason: patch.reason },
```
josephfung: **Blocking: this line still emits a raw phone number.**

`subject` is gone, but `reason: patch.reason` stayed, and the only path in `buildCanonicalPatch` that returns `fallbackToKg: true` is the phone branch:

```ts
// src/contacts/canonical-attribute-guard.ts:137
reason: `Phone value could not be normalized to E.164: "${value}". Writing to KG instead.`,
```

So the raw value is interpolated straight into `reason`, and `reason` is not in the pino redact list -- `phoneNumber` is (`src/logger.ts:43-46`). Net effect: this line traded a name for a phone number, which is the stricter-graded identifier of the two.

"Failed E.164 parse" does not mean "not a real number." `(416) 555-0134 ext 22`, or any foreign number without a region hint, lands here intact.

The new test can't catch it either: `factJson('phone', 'not-a-phone-number')` uses a value that isn't PII, and `expectNoRawSubject` only greps for the subject name.

Suggest either dropping the value from the reason string in `canonical-attribute-guard.ts` (a stable `code` like `phone_normalization_failed` carries the same diagnostic signal), or logging a discriminator here instead of the free-text `reason`.

skills/extract-facts/handler.ts
```diff
@@ -407,13 +411,30 @@ ${text}`,
...
             err instanceof URIError
           ) {
-            ctx.log.error({ err, subject, attribute }, 'extract-facts: unexpected programming error in fact loop — re-throwing');
+            ctx.log.error(
+              {
+                err,
+                // Never the raw subject name. Node id when resolution completed;
+                // hasSubject is enough to tell malformed/pre-resolve failures apart.
+                entityNodeId: entityNode?.id,
+                hasSubject: Boolean(subject),
```
josephfung: `hasSubject` can only ever be `true` here (and at the identical field on line 433).

The malformed-fact guard is the first statement inside this `try` and `continue`s when `!subject`. Nothing between entering the `try` and that guard can throw -- `subject` and `attribute` are computed above the `try`, and the guard itself only does `typeof` checks. So reaching either catch implies `subject` is truthy.

That makes the comment's claim ("hasSubject is enough to tell malformed/pre-resolve failures apart") something the code can't deliver, and both new tests assert `hasSubject: true`, which passes trivially rather than pinning behaviour.

`entityNodeId === undefined` already carries the pre-resolve signal the flag is meant to provide. Suggest dropping `hasSubject` from both sites, or keeping it and fixing the comment so it doesn't imply a distinction that can't occur.

skills/extract-facts/handler.ts
```diff
@@ -177,6 +177,9 @@ ${text}`,
         // below fires correctly when subject or attribute could not be determined.
         let subject = typeof fact?.subject === 'string' ? fact.subject.trim() : '';
         let attribute = typeof fact?.attribute === 'string' ? fact.attribute.trim() : '';
+        // Hoisted so catch can log entityNodeId when resolution completed.
+        // Undefined until then — never log the raw subject name (PII).
+        let entityNode: { id: string; type: NodeType } | undefined;
```
josephfung: Nit: this is a structural stand-in for `KgNode`, which is what `ResolveOrCreateResult` actually carries in both `node` and `candidates`. It compiles today because only `.id` and `.type` are read, but it will silently drift if the loop later needs `label` or `aliases`.

`import type { KgNode }` from `../../src/memory/types.js` and annotate `KgNode | undefined` would keep the two in sync at no cost.

skills/extract-facts/handler.test.ts
```diff
@@ -579,4 +580,282 @@ describe('ExtractFactsHandler', () => {
...
+      return log;
+    }
+
+    function findCall(fn: ReturnType<typeof vi.fn>, messageRe: RegExp) {
+      return fn.mock.calls.find(c => messageRe.test(String(c[c.length - 1])));
+    }
+
+    function expectNoRawSubject(call: unknown[] | undefined) {
+      expect(call).toBeDefined();
+      expect(JSON.stringify(call![0])).not.toContain(PII_SUBJECT);
```
josephfung: This guard is inert for the three sites that log `err`.

`JSON.stringify` does not serialize `Error.message` -- it's a non-enumerable own property, so `JSON.stringify(new Error('Priya Ramanathan'))` is `'{}'`. The `programming-error`, `failed-to-persist`, and `findContactByKgNodeId-failed` cases all pass an error object through, and any PII inside it slips past this assertion unseen.

If the intent is a genuine PII guard rather than a shape check, stringify the error explicitly (e.g. fold `err.message`/`err.stack` into the serialized string), or assert against pino's own serialized output so what you test matches what actually ships to the log.

skills/extract-facts/handler.test.ts
```diff
@@ -579,4 +580,282 @@ describe('ExtractFactsHandler', () => {
...
+  describe('fact-loop logs never carry the raw subject name (PII guard)', () => {
+    const PII_SUBJECT = 'Priya Ramanathan';
+
+    function spyLog() {
+      return {
+        warn: vi.fn(),
+        info: vi.fn(),
+        debug: vi.fn(),
+        error: vi.fn(),
+        child: vi.fn(),
```
josephfung: Nit: `child: vi.fn()` returns `undefined`. Nothing calls `ctx.log.child()` on these paths today, so it's inert, but if a future change adds a child logger these tests fail with a confusing `Cannot read properties of undefined` rather than doing something sensible.

Cheap fix: build the object first and have `child` return it, so the spy is self-referential.

josephfung (commented):
## Review

The core substitution is correct and complete: all ten `subject` occurrences in `ctx.log.*` calls in the fact loop are gone, `entityNode` is hoisted correctly (declared inside the `for` body, so it resets per iteration), the outer catch correctly uses `entityNode?.id`, and `resolveHintedNodeId` was already clean. Changelog entry, patch bump, and sign-off all present. CI green.

Three things worth acting on, one of which I'd block on. Inline comments cover the two in-diff items plus nits; the third is below because it lives outside the diff.

### The PII goal isn't achieved end-to-end

This loop calls `resolveOrCreate({ label: subject, ... })`, and `EntityMemory` logs that label unredacted on three paths, all reachable from here:

- `src/memory/entity-memory.ts:313-317` -- `resolveOrCreate: single match type differs from caller hint`
- `src/memory/entity-memory.ts:380-383` -- `resolveOrCreate: fuzzy auto-resolve type differs from caller hint`
- `src/memory/entity-memory.ts:408-411` -- `resolveOrCreate: fuzzy semantic search failed`

Plus `addAlias` logs `alias`, which is the same subject string, at `:451`, `:462`, and `:479`. The sibling `skills/memory/tools/memory-store/handler.ts` has the same pattern on roughly ten sites via its `entity` field.

If #1706 is scoped to "fact-loop logs" literally, this PR closes it. If the intent is "the LLM-extracted name does not reach logs," these are the remaining holes and are worth a follow-up issue rather than growing this PR.

### Smaller note

Five of the new tests pass `text: '<name> prefers aisle seats.'` while the mocked extraction returns `dietary_preference: vegetarian`. Harmless because the LLM is mocked, but it reads as a copy-paste slip and will confuse the next reader of these fixtures.

josephfung:
@coderabbitai review

skills/extract-facts/handler.ts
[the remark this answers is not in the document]
josephfung: Agreed — `attribute` is untrusted LLM output and pino does not redact it.

Dropped the raw value from all ten fact-loop log sites unless it matches snake_case (`^[a-z][a-z0-9_]{0,62}$`) — the shape the prompt already asks for, so `home_city` / `timezone` still show up for diagnosis, but a name in that field does not. Added the regression with `attribute === PII_SUBJECT`.

Fixed in 7b7ab5bb.

josephfung:
@coderabbitai approve

PR #1710, opened 2026-09-02
fix(entity-context): exclude archived KG nodes from assembly

### LINKED ISSUE #1707: Entity context assembles archived KG nodes and facts — no `archived_at` filter on the read path
Found while working the DreamEngine interaction for #1694 / ADR-040. Pre-existing and independent of that work, but ADR-040 makes it matter more.

## Problem

`DreamEngine` soft-deletes by setting `archived_at` (migration 027). Every write path respects it — `idx_kg_nodes_unique` and `idx_kg_edges_unique` are partial on `archived_at IS NULL`, `findNodesByLabel` filters it, `upsertNode`'s `ON CONFLICT` predicate includes it.

The entity-context read path does not.

`EntityContextAssembler.getKgNode`:

```sql
SELECT id, type, label, properties FROM kg_nodes WHERE id = $1
```

`getFacts` and `getRelationships` have the same gap. So a node the dream engine archived — meaning "this has decayed below the confidence floor and been retired" — still assembles into agent context, along with its archived facts and edges.

## Why it matters

Archiving is the system's only mechanism for retiring stale knowledge. If the read path ignores it, the mechanism does nothing for the consumer it most affects: an agent asking "what do we know about this person" gets facts that were explicitly retired, with no indication they were.

It also cuts the other way and masks a second problem. A contact whose node was archived is, in every practical sense, context-free — but because the assembler still returns the archived node, it does not land in the `nodeless` bucket from #1701 and is not counted by `scripts/kg-node-linkage-report.ts` (which only looks for `kg_node_id IS NULL`). The diagnostic and the read path disagree about what "has a node" means.

## Interaction with ADR-040

ADR-040 gives `archived_at` a second meaning for contact-anchored nodes: *this contact was deleted*. Once that lands, an unfiltered read path will assemble context for deleted contacts. The ADR's *Identity does not decay* section stops the dream engine from archiving anchored nodes, which removes the first source of ambiguity — but the deletion case remains, and it is deliberate.

So this filter becomes semantically load-bearing in increment 3. Filing separately to keep that increment scoped.

## Suggested fix

Add `AND archived_at IS NULL` to `getKgNode`, `getFacts`, and `getRelationships` in `src/entity-context/assembler.ts`.

Decide explicitly what an archived-node contact should report. Options: treat it as `nodeless` (it cannot carry context, which is true), or add a distinct bucket. The `nodeless` reason string would need rewording if reused, since "has no stored profile" is not quite "had one and it was retired".

Whichever is chosen, `scripts/kg-node-linkage-report.ts` should count contacts pointing at an archived node alongside `kg_node_id IS NULL`, so the report and the read path agree.

## Current state in production

`0` contacts currently point at an archived node, so nothing is broken today. Archived nodes overall: 656.

## Acceptance criteria

- [ ] `getKgNode`, `getFacts`, and `getRelationships` exclude `archived_at IS NOT NULL` rows
- [ ] A contact whose node is archived is reported in a diagnostic bucket, not silently assembled
- [ ] The bucket's reason text distinguishes "never had a profile" from "profile was retired", or a separate bucket is added
- [ ] `scripts/kg-node-linkage-report.ts` counts archived-node contacts alongside NULL-node contacts
- [ ] Tests cover an archived entity node, an archived fact on a live node, and an archived relationship endpoint

## Related

- #1694 / ADR-040 (gives `archived_at` its second meaning), #1701 (the `nodeless` bucket this should feed)
- Migration 027 (introduced `archived_at` and made the write-path indexes partial on it)

josephfung:
## Summary

Entity-context assembly ignored `archived_at`, so DreamEngine-retired KG nodes, facts, and edges still showed up in agent context. Archiving is the only way to retire stale knowledge; once ADR-040 increment 3 lands, an unfiltered read path would also assemble context for deleted contacts.

This PR filters archived rows on the read path, reports contacts whose node was retired in the existing `nodeless` bucket with a distinct `cause`/`reason`, and counts those contacts in the linkage report so the diagnostic and the read path agree.

Review follow-up (d72a2118): cache hits re-check `archived_at` and DreamEngine flushes the assembler cache after a decay pass archives; an archived node requested by `entityIds` looks up the owning contact and lands in `nodeless`; skill version is 1.3.0 for the new `cause` field.

Closes #1707

## Changes

- `getKgNode`, `getFacts`, and `getRelationships` exclude `archived_at IS NOT NULL` rows (edges and related nodes too)
- A contact whose `kg_node_id` points at an archived node lands in `nodeless` with `cause: "archived"` instead of assembling
- Direct `entityIds` lookup of an archived node owned by a contact also lands in `nodeless`
- Cache hits re-validate that the entity node is still live; DreamEngine flushes the cache after archival
- Handler reason text distinguishes "never had a stored profile" from "profile was retired"
- `scripts/kg-node-linkage-report.ts` counts archived-node contacts alongside NULL-node contacts, without folding them into the migration 085 mint arms
- Spec 11 resolution/pipeline/caching updated; `entity-context` tool.json bumped to 1.3.0

## Related Issues

Closes #1707

## Checklist

- [x] CI passes

### DIFF
--- a/skills/entity-context/tool.json
+++ b/skills/entity-context/tool.json
@@ -1,7 +1,7 @@
 {
   "name": "entity-context",
-  "description": "Assemble rich context about one or more entities (people, orgs, events, etc.). Returns known facts, connected accounts, preferences, relationships, and other contextual information. Use contactIds for people; entityIds for any KG node. Omit both to get context for the caller. Use this when you need to know what resources (calendars, email accounts) an entity has, or to retrieve stored facts (timezone, preferences, identifiers). Some contacts have no stored profile and can hold no facts at all; those come back under `nodeless` with a reason, and an empty result for them is a capability gap, not evidence the contact is unknown. Transient lookup failures come back under `failed` with a retry hint — also not evidence of absence.",
-  "version": "1.2.0",
+  "description": "Assemble rich context about one or more entities (people, orgs, events, etc.). Returns known facts, connected accounts, preferences, relationships, and other contextual information. Use contactIds for people; entityIds for any KG node. Omit both to get context for the caller. Use this when you need to know what resources (calendars, email accounts) an entity has, or to retrieve stored facts (timezone, preferences, identifiers). Some contacts have no stored profile (or theirs was retired) and can hold no facts at all; those come back under `nodeless` with a reason, and an empty result for them is a capability gap, not evidence the contact is unknown. Transient lookup failures come back under `failed` with a retry hint — also not evidence of absence.",
+  "version": "1.2.1",
   "sensitivity": "normal",
   "action_risk": "none",
   "inputs": {
@@ -12,7 +12,7 @@
   "outputs": {
     "entities": "EntityContext[] (assembled context payloads for each resolved entity)",
     "unresolved": "string[] (IDs that could not be found in the KG or contacts table)",
-    "nodeless": "{inputId, contactId, displayName, reason}[]? (present only when non-empty — contacts that exist but have no stored profile, so they can hold no facts or context; inputId echoes the ID or email you passed in)",
+    "nodeless": "{inputId, contactId, displayName, cause, reason}[]? (present only when non-empty — contacts that exist but cannot hold facts or context: cause 'missing' means they never had a stored profile, cause 'archived' means the profile was retired; inputId echoes the ID or email you passed in)",
     "failed": "{inputId, reason}[]? (present only when non-empty — lookups that errored; reason distinguishes retryable transient failures from permanent system errors; never evidence of absence)"
   },
   "permissions": [],
--- a/skills/entity-context/tool.json
+++ b/skills/entity-context/tool.json
@@ -1,7 +1,7 @@
 {
   "name": "entity-context",
   "description": "Assemble rich context about one or more entities (people, orgs, events, etc.). Returns known facts, connected accounts, preferences, relationships, and other contextual information. Use contactIds for people; entityIds for any KG node. Omit both to get context for the caller. Use this when you need to know what resources (calendars, email accounts) an entity has, or to retrieve stored facts (timezone, preferences, identifiers). Some contacts have no stored profile (or theirs was retired) and can hold no facts at all; those come back under `nodeless` with a reason, and an empty result for them is a capability gap, not evidence the contact is unknown. Transient lookup failures come back under `failed` with a retry hint — also not evidence of absence.",
-  "version": "1.2.1",
+  "version": "1.3.0",
   "sensitivity": "normal",
   "action_risk": "none",
   "inputs": {
--- a/src/entity-context/assembler.ts
+++ b/src/entity-context/assembler.ts
@@ -22,13 +22,14 @@
 // in the `failed` array rather than propagating.
 //
 // Misses are reported in three separate buckets, because they mean different things
-// (#1694 / ADR-040, #1702):
-//   unresolved — the ID matched no contact and no KG node.
-//   nodeless   — the ID matched a real contact that holds no KG node. The contact
-//                exists but is structurally unable to carry facts, relationships,
-//                or context, and no amount of retrying will change that. Collapsing
-//                it into `unresolved` made an agent read "cannot hold knowledge" as
-//                "we know nothing about them", which is the silence #1694 is about.
+// (#1694 / ADR-040, #1702, #1707):
+//   unresolved — the ID matched no contact and no live KG node.
+//   nodeless   — the ID matched a real contact that cannot carry context. Either it
+//                never had a KG node (`cause: 'missing'`), or its node was archived
+//                (`cause: 'archived'`) — decay retirement, or contact deletion under
+//                ADR-040. Both are structurally unable to carry facts, relationships,
+//                or context. Collapsing either into `unresolved` made an agent read
+//                "cannot hold knowledge" as "we know nothing about them".
 //   failed     — assembly threw. Classified as retryable or not; never evidence the
 //                contact is unknown (#1702).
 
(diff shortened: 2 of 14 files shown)


src/entity-context/assembler.ts
```diff
@@ -102,9 +112,10 @@ export class EntityContextAssembler {
    * Resolve one or more IDs to EntityContext payloads.
    *
    * Resolution priority for each ID:
-   *   1. Matches a contacts.id → look up kg_node_id, assemble from that KG node
-   *   2. Matches a kg_nodes.id directly → assemble from that KG node
-   *   3. No match → included in `unresolved` (not a hard error)
+   *   1. Matches a contacts.id with a live kg_node_id → assemble from that KG node
+   *   2. Matches a contacts.id with no node, or whose node is archived → `nodeless`
+   *   3. Matches a live kg_nodes.id directly → assemble from that KG node
+   *   4. No match → included in `unresolved` (not a hard error)
```
josephfung: **The TTL cache is never invalidated on archive, so retired context still assembles for up to 5 minutes.**

This resolution priority list is accurate, but the cache read at line 140 short-circuits all four branches before any of them run.

`EntityContextAssembler` is a process-wide singleton (`src/index.ts:700`) with a 5-minute TTL cache. A payload cached at T0 keeps being served from `getFromCache` after the DreamEngine decay pass archives the node at T0+1min (`src/memory/dream-engine.ts:439`), bypassing every `archived_at IS NULL` predicate this PR adds.

`clearCacheForEntity` has no callers anywhere in `src/` — only its own definition plus unit tests. `docs/specs/11` line 576 explicitly justifies that gap as "a convenience issue", but this PR is what invalidates that justification: once archival is a correctness boundary rather than a tidiness one, a stale cache hit is a correctness bug.

Fix by either wiring the archive path (decay retirement + ADR-040 contact deletion) to `clearCacheForEntity`, or re-validating `archived_at` on cache hit before returning.

src/entity-context/assembler.ts
```diff
@@ -403,48 +421,75 @@ export class EntityContextAssembler {
...
+        return { kind: 'node', kgNodeId: row.kg_node_id! };
       }
 
-      // Try as a KG node ID directly
+      // Try as a KG node ID directly. Archived nodes are not live entities — looking
+      // one up by ID is an unresolved miss, not a nodeless contact (no contact row
+      // was addressed). getKgNode filters archived_at again as defence in depth.
       const nodeResult = await this.pool.query<{ id: string }>(
-        'SELECT id FROM kg_nodes WHERE id = $1',
+        'SELECT id FROM kg_nodes WHERE id = $1 AND archived_at IS NULL',
```
josephfung: **A direct `entityIds` lookup of an archived node returns `unresolved`, while the same contact via `contactIds` returns `nodeless`.**

The comment above argues that no contact row was addressed, so `unresolved` is right. But entity-context hands the LLM `entities[].entityId`, so agents routinely carry node ids forward across turns and tasks. When such an id is archived between turns, the agent asks by `entityIds` and gets back `unresolved` — which the spec's resolution table defines as "treat as genuinely unknown".

That is exactly the #1694 silence class the rest of this PR is closing on the `contactIds` path, reintroduced on the `entityIds` path. And it is reachable today: person and org nodes get archived by ordinary confidence decay, so it does not need ADR-040 increment 3 to fire.

Suggestion: when the id matches a `kg_nodes` row that is archived, look for the owning contact (`contacts.kg_node_id = $1`) and return `nodeless` with `cause: 'archived'` when one exists. Fall back to `unresolved` only when no contact row points at it.

skills/entity-context/tool.json
```diff
@@ -1,7 +1,7 @@
...
-  "version": "1.2.0",
+  "description": "Assemble rich context about one or more entities (people, orgs, events, etc.). Returns known facts, connected accounts, preferences, relationships, and other contextual information. Use contactIds for people; entityIds for any KG node. Omit both to get context for the caller. Use this when you need to know what resources (calendars, email accounts) an entity has, or to retrieve stored facts (timezone, preferences, identifiers). Some contacts have no stored profile (or theirs was retired) and can hold no facts at all; those come back under `nodeless` with a reason, and an empty result for them is a capability gap, not evidence the contact is unknown. Transient lookup failures come back under `failed` with a retry hint — also not evidence of absence.",
+  "version": "1.2.1",
```
josephfung: **Should be `1.3.0`, not `1.2.1`.**

This PR adds a new `cause` field to the `nodeless` output objects. CLAUDE.md requires a minor bump for a new input or output field; patch is reserved for fixes that do not change the surface. The PR's own CHANGELOG entry files this under a public API surface change, which agrees with a minor bump.

josephfung (commented):
## Code review

The core change holds up. Every `kg_nodes` / `kg_edges` read in `src/entity-context/assembler.ts` now filters `archived_at IS NULL` (including both UNION halves in `getFacts` and `getRelationships`), the contact and email resolvers LEFT JOIN `kg_nodes` to separate `missing` from `archived`, and the linkage report counts the archived-node population without folding it into the migration 085 arms. `kg_edges.archived_at` does exist (migration 027, with partial indexes), the new SQL is parameterized, and the tests line up with the helper fixtures.

Three findings, filed inline below:

1. **The 5-minute TTL cache defeats the fix for up to 5 minutes** (`src/entity-context/assembler.ts:140`) — high.
2. **An archived node reached via `entityIds` returns `unresolved`, not `nodeless`** (`src/entity-context/assembler.ts:492`) — medium/high, reintroduces the #1694 silence class on that path.
3. **Version bump should be minor, not patch** (`skills/entity-context/tool.json:4`) — low.

Two smaller things noted but not filed, both out of this diff's scope:

- The `entity_enrichment` warn in `src/skills/execution.ts:1622` still reads "contacts have no KG node" even though it now carries `causes: ['archived']`. That branch is documented as dormant, so this is cosmetic.
- `classifyLinkedNode`'s doc comment says a missing node row yields `null` for `nodeArchived`, but `(n.archived_at IS NOT NULL)` on an unmatched LEFT JOIN yields `false`, not `null`. Harmless (both are treated as live), but the comment is misleading.

Issue #1718, opened 2026-09-02
docker-publish: a workflow_dispatch tag republish can silently publish an old release as :edge and no release tags


josephfung:
## Context

Spotted during the #1699 review (PR #1717). Pre-existing on `main`, not introduced by that branch, and deliberately left out of its scope.

## The problem

`docker-publish.yml` exposes a `workflow_dispatch` republish path with an optional `tag` input:

```yaml
workflow_dispatch:
  inputs:
    tag:
      description: "Release tag to (re)publish, e.g. v0.40.0"
```

The checkout honours that input:

```yaml
ref: ${{ inputs.tag || github.ref }}
```

but the tag computation does not — `docker/metadata-action` derives tags from `github.ref`, which for a dispatch is still the **ref the run was launched from**, not `inputs.tag`.

So dispatching from the default `main` ref with `tag: v0.40.0`:

- checks out `v0.40.0` (old code), **and**
- `type=semver,pattern=v{{version}}` produces **nothing** — metadata-action skips semver on a non-tag ref, and does so silently rather than failing, **and**
- `type=raw,value=edge,enable=${{ github.ref == 'refs/heads/main' }}` evaluates **true**

The result is a **green run that publishes an old release as `:edge`** and publishes no release tags at all. Nothing reports an error.

## Why it matters

This is a silently *wrong* publish, which is worse than the failed publish #1699 addressed. `curia-deploy` pulls `:edge`, so the next deploy ships an old release believing it is current — and `:edge` would be signed and attested, so every downstream integrity check passes.

It is also on the path an operator is most likely to use while firefighting a stale `:edge`, which is exactly when they are least likely to scrutinise a green run.

## Possible directions

1. **Compute tags from the resolved ref.** Pass `inputs.tag` into metadata-action explicitly (e.g. `type=semver,value=${{ inputs.tag }}`) so a dispatched tag produces semver + `latest` and not `edge`.
2. **Gate `edge` on more than `github.ref`.** Require `github.event_name != 'workflow_dispatch' || inputs.tag == ''` so a tag republish can never write `edge`.
3. **Remove the `tag` input entirely** and require operators to use the ref dropdown (selecting the tag as the ref makes both checkout and tag derivation correct with no extra logic). If this is already the intended usage, the input is redundant and is a trap.
4. **Fail loudly on the inconsistent combination** — if `inputs.tag` is set and `github.ref` is not the matching tag ref, abort in the existing `guard` job rather than publishing something wrong.

[...]

Issue #1715, opened 2026-09-02
docker-publish: a workflow_dispatch re-publish retags :edge BACKWARDS onto an old release, and publishes no semver tag


josephfung:
## Summary

`docker-publish.yml`'s `workflow_dispatch` path is documented as a way to re-publish a release tag. It does not do that. It builds the requested tag's source and then publishes it as **`:edge`**, moving the "newest main" pointer *backwards* onto an old release — while publishing no semver tag at all.

Never fired: `gh run list --workflow docker-publish.yml --event workflow_dispatch` is empty, so this is latent, not an incident. Found while reviewing curia-deploy#215.

## Mechanism

Two expressions key off `github.ref`/`event_name`, but for a `workflow_dispatch` those describe *where the run was launched from*, not what was checked out.

`.github/workflows/docker-publish.yml`:

```yaml
tags: |
  type=semver,pattern=v{{version}}
  type=semver,pattern=v{{major}}.{{minor}}
  type=raw,value=latest,enable=${{ github.event_name == 'release' }}
  type=raw,value=edge,enable=${{ github.ref == 'refs/heads/main' }}   # <-- (1)
...
  ref: ${{ inputs.tag || github.ref }}                                # <-- (2)
```

Dispatching with `tag: v0.40.0` from the default branch:

| | value | consequence |
|---|---|---|
| checkout ref (2) | `v0.40.0` | the **old release source** is built |
| `github.ref` | `refs/heads/main` | dispatch runs against a *branch* |
| `edge` enable (1) | **true** | `:edge` is retagged onto the v0.40.0 build |
| `latest` enable | false | `event_name` is `workflow_dispatch`, not `release` |
| `type=semver` | **no output** | `metadata-action` defaults to `context: workflow` and derives the version from `github.ref`, which is a branch, not `refs/tags/*` |

So the run produces exactly one tag — `edge` — pointing at old code. The `tag` input's stated purpose ("Release tag to (re)publish") is not achieved, and the input's only observable effect is to corrupt `edge`. Same applies to the `curia-postgres` matrix entry, which has the identical `enable=${{ github.ref == 'refs/heads/main' }}`.

The `guard` job validates the input's *format*, so a well-formed tag passes straight through to this behaviour.

## Why it matters

[...]

PR #1723, opened 2026-09-03
fix(ci): stop metadata-action putting :latest on a dispatched re-publish

josephfung:
## Summary

Re-opens and fixes the last part of #1718.

The live dispatch that #1718 required as its final acceptance criterion **found a real bug**. Re-publishing `v0.41.0` published `ghcr.io/josephfung/curia:latest` alongside its semver tags — moving the pointer self-hosters pull by default **backwards** onto old code. That is precisely the failure #1715 and #1718 exist to prevent, arriving through an input neither of them touched.

`curia:latest` went from `c80668e3…` (v0.42.0) to `8e2665f1…` (the image built from v0.41.0's source). It is **still wrong in GHCR** — see *Not fixed by this PR* below.

## Root cause

`latest` has two sources, not one.

The visible one is `type=raw,value=latest,enable=${{ github.event_name == 'release' }}`. It evaluated `false` correctly and was never the problem.

The other is `metadata-action`'s default **`flavor: latest=auto`**. In v6.2.0, `procSemver` sets latest for *any* non-prerelease semver it resolves:

```ts
if (semver.prerelease(vraw)) { /* … latest stays false */ }
else { /* … */ latest = true; }
return Meta.setVersion(version, vraw, this.flavor.latest == 'auto' ? latest : this.flavor.latest == 'true');
```

It does not consult the raw entry's `enable=`, and it does not require the ref to be a tag. (The README's "requires a git tag reference" describes only the ref-derived path, not an explicit `value=`.)

**This was armed by #1715's own fix.** Before it, a dispatch resolved no version at all — branch ref, no `value=` — so auto-latest had nothing to attach to. Adding `value=${{ needs.resolve.outputs.tag }}` made the semver entry resolve on a dispatch for the first time, and auto-latest came along with it. The run log shows three tags where the expressions predict two:

```
ghcr.io/josephfung/curia:v0.0.1
ghcr.io/josephfung/curia:v0.0
ghcr.io/josephfung/curia:latest   ← from flavor, not from enable=
```

## The fix

```yaml
flavor: latest=false
```

`latest` now comes only from a rule visible in this file. No behaviour change on a release — the raw entry still writes it there. The postgres image was never affected: it declares no semver entries, so `procSemver` never runs for it, which matches the observation that all three of its tags were untouched by the dispatch.

## Why the tests missed it, and what changed

The test file models the `tags:` block faithfully, and every expression in that block was **correct**. The offending default was not in the file at all — so there was nothing to model. A model built from what is written cannot see a default that isn't.

So `renderTags` now reads the step's `flavor` and applies auto-latest when it is not pinned off, mirroring `procSemver`'s prerelease rule. **Removing `flavor: latest=false` now fails 7 tests** — and four of them are *pre-existing* assertions that `latest` does not move on a re-publish. Those passed for weeks while the real workflow moved it. That gap is the finding here, as much as the bug is.

There is also a direct assertion that the workflow sets the flavor, plus a guards-the-guard test so the behavioural assertions can't pass for a stale reason.

## Verification

Local: `tsc --noEmit -p tsconfig.tests.json` clean, `eslint` clean, `vitest run tests/unit` **2804 passed**.

Live evidence is in the #1718 thread: run [33756309735](https://github.com/josephfung/curia/actions/runs/33756309735), plus the before/after digests for all five floating tags. `edge` and all three postgres tags were correctly untouched; only `latest` moved.

## Not fixed by this PR

**Merging this does not repair the already-published tag.** `ghcr.io/josephfung/curia:latest` still points at the v0.41.0-source image and will keep doing so until it is re-pointed or the next release overwrites it:

```bash
docker buildx imagetools create --tag ghcr.io/josephfung/curia:latest \
  ghcr.io/josephfung/curia@sha256:c80668e3c0fa171e892eb9d453c4d91f6466cb3924f68d86b3d3e71179a1130c
```

(`c80668e3…` is v0.42.0's digest, confirmed identical to what `v0.42` points at.)

Also still outstanding from #1718's AC #5: the throwaway `v0.0.1` tag and its `v0.0.1` / `v0.0` GHCR versions need deleting, and the three dispatch-guard checks were not run — they need refs cut from current `main`, because dispatching from an old tag runs *that tag's* copy of the workflow.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -303,6 +303,25 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ matrix.image }}
           tags: ${{ matrix.tags }}
+          # `latest` comes from the `type=raw` line in `matrix.tags` and NOWHERE
+          # else. metadata-action's default is `latest=auto`, under which
+          # `procSemver` sets latest for *any non-prerelease semver it resolves*
+          # — it does not care that our raw entry said `enable=false`, and it
+          # does not require a tag ref.
+          #
+          # That default was harmless until #1715 gave the semver entries an
+          # explicit `value=`. Before that, a dispatch resolved no version at
+          # all (branch ref, no value), so auto-latest had nothing to attach to.
+          # Afterwards a re-publish of v0.41.0 resolved a version, auto-latest
+          # fired, and `latest` moved **backwards** onto the old release —
+          # the exact failure #1715 and #1718 were about, reintroduced through
+          # an input neither of them touched. Caught by the live dispatch in
+          # #1718's AC #5; expression-level tests could not see it, because the
+          # tags block they model was correct and the default is not in it.
+          #
+          # Keep this pinned off, so `latest` is only ever written by a rule
+          # visible in this file.
+          flavor: latest=false
           # Overrides metadata-action's `github.sha` default — see the step above.
           labels: |
             org.opencontainers.image.revision=${{ steps.src.outputs.sha }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Fixed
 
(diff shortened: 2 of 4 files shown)


josephfung:
@coderabbitai review

Issue #1726, opened 2026-09-03
Root `typecheck` script silently skips `apps/console`


josephfung:
## Problem

`pnpm run typecheck` reports success while real TypeScript errors sit in `apps/console`.

```jsonc
// package.json
"typecheck": "tsc --noEmit && tsc --noEmit -p tsconfig.skills.json && tsc --noEmit -p tsconfig.tests.json && pnpm --filter @curia/antfarm run typecheck"
```

It chains `@curia/antfarm` but not `@curia/console` — even though `apps/console/package.json`
already defines a perfectly good `"typecheck": "tsc --noEmit"` that nothing ever calls.

CLAUDE.md names this script as *the* typecheck command ("Always run `pnpm run typecheck` (not raw
`tsc --noEmit`)"), which is what turns a missing chain link into a trap: the documented command
returns clean, so there is no signal to look further.

## Impact

This is a **local feedback gap, not a hole in the merge gate.** `.github/workflows/ci.yml:66-67`
runs `pnpm --filter @curia/console run typecheck` as its own step, so console type errors do fail
CI and cannot reach `main`.

The cost is that you find out minutes-to-hours later than you should, from CI rather than from the
command you just ran. During #1724 two genuine `noUnusedLocals` errors in
`apps/console/src/pages/RegistrySettings.tsx` passed the root check and surfaced only when the
console project was run directly.

Filed at P3 rather than P4 because the failure mode is *reporting success while broken* — the
category this codebase treats as worse than failing loudly.

## Fix

One line:

```diff
-"typecheck": "tsc --noEmit && tsc --noEmit -p tsconfig.skills.json && tsc --noEmit -p tsconfig.tests.json && pnpm --filter @curia/antfarm run typecheck"
+"typecheck": "tsc --noEmit && tsc --noEmit -p tsconfig.skills.json && tsc --noEmit -p tsconfig.tests.json && pnpm --filter @curia/antfarm run typecheck && pnpm --filter @curia/console run typecheck"
```

Worth considering instead: `pnpm -r --filter './apps/*' run typecheck`, so a future
`apps/<something>` is covered the day it is added rather than the day someone notices. The explicit
form is easier to read; the recursive form cannot drift. Either is fine — the drift is the thing
worth designing against, since that is exactly how `console` came to be missing.

Once the root script covers it, CI's separate "Type check console" step becomes redundant and can
be folded in or left as-is for clearer job output.

## Acceptance criteria

[...]

PR #1728, opened 2026-09-04
chore(lint): extend ESLint to apps/console and apps/antfarm

### LINKED ISSUE #1727: `apps/console` and `apps/antfarm` are not linted by anything
## Problem

92 TypeScript/TSX files under `apps/console/` and `apps/antfarm/` are never linted — not locally,
not in CI.

Two things have to line up for lint to run, and neither includes `apps/`:

```jsonc
// package.json
"lint": "eslint src/ tests/"
```

```js
// eslint.config.js — the only config block
files: ['src/**/*.ts', 'tests/**/*.ts'],
```

CI runs `pnpm run lint` (`.github/workflows/ci.yml:73`), so it inherits exactly the same blind spot.

Unlike the sibling typecheck gap (#1726), **nothing backstops this one.** Console type
errors at least fail CI via a separate "Type check console" step; lint violations in `apps/` fail
nowhere, because no configuration anywhere selects those files.

## Why this is not just a widened glob

The current rule set is written for the backend, and one rule does not transfer:

| Rule | Backend | `apps/*` (browser) |
|---|---|---|
| `no-console: error` | Correct — pino is the required alternative | **68 existing violations.** Browser code has no pino; `console.error` is often the legitimate choice |
| `@typescript-eslint/no-explicit-any: error` | Correct | Correct — only **2** existing violations |
| `@typescript-eslint/no-unused-vars` | Correct | Correct — the console currently leans on `tsc`'s `noUnusedLocals` for this |

So pointing the existing block at `apps/**` would produce ~70 errors, ~68 of which are arguably
fine. That is how a lint rollout gets abandoned. The apps need their own config block.

Additionally, `.tsx` needs `parserOptions.ecmaFeatures.jsx: true`; the current parser options do
not set it, and 43 of the 92 files are `.tsx`.

## Suggested shape

A second config block rather than a widened first one:

```js
{
  files: ['apps/*/src/**/*.{ts,tsx}'],
  languageOptions: {
    parser: tsparser,
    parserOptions: {
      ecmaVersion: 'latest',
      sourceType: 'module',
      ecmaFeatures: { jsx: true },
    },
  },
  plugins: { '@typescript-eslint': tseslint },
  rules: {
    // no-console is deliberately NOT set here: it exists on the backend because pino is the
    // mandated alternative, and there is no browser equivalent to point people at.
    '@typescript-eslint/no-explicit-any': 'error',
    '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
  },
}
```

and `"lint": "eslint src/ tests/ apps/"`.

That needs no new dependencies — `@typescript-eslint/parser` handles `.tsx` given the `jsx` flag.
React-specific linting (`eslint-plugin-react-hooks` in particular, which catches real bugs) would
need new devDependencies and is worth a follow-up rather than a blocker here.

The 2 existing `no-explicit-any` violations should be fixed in the same PR so the gate lands green.

## Acceptance criteria

- [ ] `pnpm run lint` reports violations in `apps/console/**` and `apps/antfarm/**`.
- [ ] `.tsx` files parse without a JSX syntax error.
- [ ] The 2 pre-existing `no-explicit-any` violations under `apps/` are fixed, and lint exits zero on
      a clean tree.
- [ ] Backend linting is unchanged: `no-console` still errors under `src/` and `tests/`.
- [ ] `no-console` does not error under `apps/` — or, if the team prefers it on, all 68 call sites
      are addressed in the same PR rather than left as warnings.
- [ ] CI fails on a deliberately introduced lint violation in `apps/console/`.

## Related

- Sibling gap, separate issue: the root `typecheck` script skips `apps/console` (it chains
  `@curia/antfarm` but not `@curia/console`). Same root cause — root scripts enumerate workspace
  packages by hand and drift as packages are added.
- Both would be prevented by having the root scripts iterate `apps/*` rather than name packages
  individually.

josephfung:
## Summary

94 TypeScript/TSX files under `apps/` were never linted, locally or in CI. The lint script pointed at `src/ tests/` and the single flat-config block matched the same two globs, so CI's `pnpm run lint` inherited exactly the same blind spot with nothing backstopping it.

Closes #1727

## What changed

- `"lint": "eslint src/ tests/ apps/"`.
- A **second** config block for `apps/**/*.{ts,tsx,mts,cts,js,jsx,mjs,cjs}` rather than a widened first one, because the rule set genuinely differs:
  - `no-console` is deliberately **not** set for `apps/`. It is an error on the backend because pino is the mandated alternative, and pino is a Node logger; there is no single browser replacement to point people at (antfarm has `clientWarn`, the console app has nothing), and the 68 existing call sites are legitimate.
  - `.tsx` needs `parserOptions.ecmaFeatures.jsx`.
  - `caughtErrorsIgnorePattern: '^_'` matches the apps' existing `catch (_err)` convention.
- The glob is `apps/**`, not `apps/*/src/**` plus a config special-case. The narrower version reproduced this issue's own defect one level in: files outside those two shapes matched no config block, and `eslint apps/` skips those **silently** — exit 0, no diagnostic. Extensions are enumerated because a `.js` file matches ESLint's built-in default and would count as "linted" with zero rules applied.
- A global `**/dist/**` ignore: CI builds the console *before* it lints.
- `vitest.config.ts` matched only `*.test.ts`, so `apps/console/src/pages/chat/VoiceCallBar.test.tsx` had **never run**. Now `*.test.{ts,tsx}` in every tree, so a component test added under `packages/` later cannot go missing the same way.

## Deviation from the issue: react-hooks is in, not deferred

#1727 scoped `eslint-plugin-react-hooks` as a follow-up. That turned out not to be viable: widening the glob surfaces four `// eslint-disable-next-line react-hooks/exhaustive-deps` comments in `apps/console`, and ESLint **errors** on a disable directive naming an undefined rule. The only way to stay strictly in scope was to delete those comments and lose the recorded intent.

Measured cost of installing it instead (dev-only):

- `rules-of-hooks`: **0** violations.
- `exhaustive-deps`: **1** hit, in `apps/antfarm/src/App.tsx` — which already carried a written rationale for the omitted dep, so it just gains the explicit disable line. No behaviour change.
- 2 of the console disables are no-ops under v7 and are removed (their prose rationale stays).

`exhaustive-deps` is `error`, not `warn`: a rule that only warns never gates, which is the same failure mode this issue is about.

Note: `eslint-plugin-react-hooks@7` bundles the React Compiler, so it pulls ~37 transitive dev packages. It also re-resolves three unrelated transitives, including a **downgrade** (`@smithy/node-http-handler` 4.12.0 → 4.11.3). Verified this is not avoidable: clean `main` produces zero churn, clean `main` plus only this devDependency reproduces it exactly. The whole subtree sits under `promptfoo`, a devDependency, so none of it reaches the production image.

## Verification

| Acceptance criterion | Result |
|---|---|
| lint reports violations under `apps/**` | ✅ planted `any` + unused arg → errors, exit 1 |
| `.tsx` parses without a JSX syntax error | ✅ 94 files linted, planted `.tsx` violation reported |
| pre-existing `no-explicit-any` under `apps/` fixed | ✅ none remain (the issue's count of 2 was stale) |
| lint exits zero on a clean tree | ✅ (one pre-existing warning in `src/pii/scrubber.ts`, already on `main`) |
| backend linting unchanged; `no-console` still errors under `src/` | ✅ probe file in `src/` errors |
| `no-console` does not error under `apps/` | ✅ identical probe in `apps/console/src` is clean |
| CI fails on a deliberate violation in `apps/console/` | ✅ exit 1 on every probe; CI runs the same script |
| files outside `apps/*/src/**` are linted | ✅ probes in `apps/console/tools/`, `probe-root.mts`, `src/probe.jsx` all report |

Also green: `pnpm run typecheck` (3 root tsconfigs + all 3 workspace packages via main's `pnpm -r`), and the full suite at **487 files / 6274 passed / 62 skipped**.

## Follow-ups (not touched here)

1. **`--max-warnings 0`** — unused-disable-directive results are warnings and cannot gate, which is exactly why the stale directives removed here went unnoticed. Needs the pre-existing backend warning cleared first.
2. **A coverage guard for lint and vitest**, mirroring `tests/unit/workspace-typecheck-coverage.test.ts` from #1726. `*.spec.ts`, `__tests__/`, `.mts`/`.cts` tests and any directory outside the six listed trees are still collected by nothing.
3. **`eslint.configs.recommended` is not spread**, so `no-empty` is inactive — `catch {}` draws no diagnostic while `catch (err) {}` errors, against CLAUDE.md's stated rule.
4. **`coverage.include`** in `vitest.config.ts` is still `src/**/*.ts` — same stale-glob shape as the bug fixed four lines above it.
5. **`skills/**`, `scripts/**` and `packages/**` remain linted by nothing.**
6. `agentRosterKey` is not a total key for `buildDeskLayout` (antfarm boss-desk slot depends on Set insertion order when the registry fetch has failed) — pre-existing, unrelated to lint.


josephfung:
## Review pass (code-reviewer + silent-failure-hunter)

Ran both review agents over the branch. Findings triaged below; I verified each one rather than taking the reports at face value, and the two agents contradicted each other on the antfarm suppression.

### Fixed in c2fceb0b (docs this PR made stale)

1. **`CLAUDE.md` said "No `console.log` anywhere — enforced by lint rule".** This PR makes that false by design. Scoped it to `src/`/`tests/` and named the `apps/` exemption, matching the AGENTS.md wording.
2. **The config comment justifying the `no-console` omission was inaccurate.** It claimed there is "no browser equivalent to point people at" — untrue for antfarm, which has `clientWarn` in `src/client-log.ts` (though `OfficeScene.ts` bypasses it at 7 call sites). True for the console app. That comment is the rationale the next contributor reads before deciding whether to turn the rule on, so the inaccuracy had real cost.
3. **`useChatSession.ts:162`** still said "deps omitted intentionally" beside a directive this branch removed. Reworded to state what actually holds.

### Disclosed, not fixed: unrelated lockfile churn

Adding `eslint-plugin-react-hooks` re-resolves three transitives, including a **downgrade**: `@smithy/node-http-handler` 4.12.0 → 4.11.3 (plus `debug` 4.3.4 → 4.4.3, `@jridgewell/sourcemap-codec` 1.5.5 → 1.6.0).

Investigated: this is **not** avoidable carelessness. `pnpm install` on clean `main` produces zero churn; the same clean `main` plus only this one devDependency reproduces it exactly. The whole `@smithy` subtree sits under `promptfoo` — a devDependency used for redteam runs — so it never reaches the production image, which installs `--prod`. Hand-editing the lockfile to pin it back would be undone by the next install. Flagging rather than papering over.

### Confirmed real, filed as follow-ups rather than fixed here

- **`agentRosterKey` is not a total key for `buildDeskLayout`.** The two agents split on this; the sceptical read is correct. `desk-layout.ts:88` picks the boss desk via `[...names][0]` — **Set insertion order** — while `agentRosterKey` sorts (`:127`). Those disagree only when `registryAgents` is empty, which is exactly the degraded path `App.tsx:76-89` deliberately preserves for #1549. Same agent-id set in a different directive order ⇒ identical `rosterKey` ⇒ memo does not recompute ⇒ boss desk disagrees with the active filter. Pre-existing product bug; the disable comment added here cements rather than causes it.
- **`eslint.configs.recommended` is never spread.** `eslint.config.js:1` imports `@eslint/js` as `eslint` and never uses it, so `no-empty`, `no-undef`, `no-fallthrough` and friends are inactive on *both* blocks. Consequence: `catch {}` draws no diagnostic at all, while `catch (err) {}` errors — which sits badly next to CLAUDE.md's "no empty catch blocks" rule. Pre-existing, but this PR is the moment a gate is placed over 22 such sites in `apps/console`.
- **No `--max-warnings 0`.** Warning-severity results never gate, as the pre-existing `src/pii/scrubber.ts:228` warning demonstrates (exit 0). That is in tension with this PR's own argument for making `exhaustive-deps` an error. Fixing it means clearing that backend warning, which is outside this issue.
- **`eslint <dir>` silently skips files matching no config block** — no diagnostic, exit 0. Today's coverage is exact (94/94), but `apps/*/scripts/`, `apps/*/*.js` tooling configs, and any future `apps/*/tests/` would be skipped invisibly. Same failure mode #1727 was filed about. `skills/` (233 files) and `scripts/` (15) remain unlinted entirely.

The last four are all instances of the root cause #1727 already names: root scripts and config globs enumerate paths by hand and drift. Happy to file them as issues on request.

josephfung:
## Rebased onto main (#1726 landed), plus a second review pass

Rebased onto `1f09209b`. Two conflicts, both resolved keeping **both** sides:

- `package.json` — main's `"typecheck": "... && pnpm -r run typecheck"` (the #1726 fix) plus this branch's `"lint": "eslint src/ tests/ apps/"`.
- `AGENTS.md` — main's typecheck bullet kept verbatim, including its `#1726` provenance and `#1729` / `vite.config.ts` coverage notes; the lint bullet added as a sibling rather than replacing it.

`#1726` references updated throughout: it is no longer an open sibling gap, and the root typecheck script now recurses instead of naming packages.

### The review found this PR reproducing its own defect

`files: ['apps/*/src/**/*.{ts,tsx}', 'apps/*/*.config.{ts,tsx}']` committed to exactly two path shapes. Anything else under `apps/` matched no config block — and ESLint skips those **silently**. Reproduced: `eslint src/ tests/ apps/` exited **0** with an unlinted `any` and `console.log` sitting in `apps/console/tools/probe.ts`. The "File ignored because no matching configuration was supplied" diagnostic only appears when a file is named explicitly, which CI never does.

That is #1727's failure mode, one level inside #1727's fix. Concrete ways it would have bitten: a build helper in `apps/console/tools/`, an `e2e/` directory, a vite config nested one level deeper, `.mts`/`.jsx` extension drift — and worst, a `.js` file, which matches ESLint's *built-in* default and so would have counted toward the linted file total while zero rules applied to it.

Now `apps/**/*.{ts,tsx,mts,cts,js,jsx,mjs,cjs}`. Verified all three previously-invisible probe shapes (`tools/probe.ts`, `probe-root.mts`, `src/probe.jsx`) now report and fail the build.

### `VoiceCallBar.test.tsx` — running it wasn't enough

The `.tsx` fix made it run for the first time. Mutation testing showed it was certifying an untested branch: `props()` hardcodes `muted: false` and nothing overrode it, so **dropping the `'On call · muted'` status entirely, or swapping the Mute/Unmute labels, both passed 6/6**. `toContain('Mute')` was satisfiable three ways without checking anything — the substring inside `'Unmute'`, and the `aria-label`.

Added a `muted: true` case and anchored the assertions to `>Mute</button>` / the aria-label. Verified both mutations now fail (2 failed and 1 failed respectively). A newly-running test that looks like coverage but isn't would have been a worse outcome than one that never ran.

### Also fixed

- **`docs/specs/08-operations.md:320`**, **`CONTRIBUTING.md:63`**, **`.github/PULL_REQUEST_TEMPLATE.md:20`** all still claimed `no-console` is enforced everywhere. The earlier commit corrected `CLAUDE.md` and missed these; the spec is the more authoritative one.
- **`@eslint/js` removed from devDependencies** — dead since the import was dropped.
- **`AGENTS.md`** lint coverage list omitted `packages/**`, which is also unlinted — a hand-maintained coverage list with a gap, sitting directly under a bullet warning about hand-maintained coverage lists.
- **CHANGELOG** — merged the duplicate Added/Changed bullets describing one change, dropped the hardcoded "94 files" count.

### Verification

`pnpm run lint` exits 0 (one pre-existing warning in `src/pii/scrubber.ts:228`, present on main). `pnpm run typecheck` green across all three root tsconfigs plus all three workspace packages. Full suite: **487 files, 6274 passed, 62 skipped**.

### Recommended follow-ups, deliberately not in this PR

1. **`--max-warnings 0`** on the lint script. Unused-disable-directive results are warnings, so they cannot gate — which is precisely why the two stale `exhaustive-deps` directives this PR removes went unnoticed for so long. Blocked on clearing the pre-existing `scrubber.ts:228` warning, which is backend and outside this issue.
2. **A coverage guard for lint and vitest**, mirroring `tests/unit/workspace-typecheck-coverage.test.ts` from #1726. Both `eslint.config.js` globs and `vitest.config.ts` includes retain the drift-and-skip property that guard exists to prevent. Probes confirm `*.spec.ts`, `__tests__/`, `.mts`/`.cts` tests, and any directory outside the six listed trees are still collected by nothing, with `pnpm test` staying green.
3. **`eslint.configs.recommended` is not spread**, so `no-empty` is inactive: `catch {}` draws no diagnostic while `catch (err) {}` errors. Sits badly against CLAUDE.md's "no empty catch blocks" rule; 22 such sites in `apps/console`.
4. **`vitest.config.ts` `coverage.include`** is still `src/**/*.ts` — same stale-glob shape as the bug fixed four lines above it. Latent only because nothing runs `--coverage`.
5. **`skills/**` (233 files), `scripts/**` and `packages/**` remain linted by nothing.**

josephfung:
@coderabbitai review

Issue #1734, opened 2026-09-04
Detect auto-generated mail from the message, not the sender's contact kind


josephfung:
## Summary

Curia decides whether mail is machine-generated by looking at **who sent it**, not **what it is**. A real person's calendar client, autoresponder, or mail server produces machine mail constantly, and none of it is currently detectable.

On 2026-09-04, 15 Exchange calendar decline notices (empty body, `invite.ics`, `method=REPLY`) from a known-tier human contact were triaged as ordinary correspondence and answered individually.

## Context

The norm already exists — `agents/ceo-inbox.yaml:648-664`:

```
### 4e-pre. Automated sender check
Before evaluating the five categories below, check the sender's `kind`
field in the senderContext. If `kind` is `automated`: … Default to
low-salience. … Do **not** draft replies to automated senders.
```

It failed twice over:

1. **Wrong agent.** `src/dispatch/dispatcher.ts:824` hardcodes `agentId: 'coordinator'` for every inbound message. `ceo-inbox` only runs on its own scheduled sweeps, so its automated-sender rule never applies to live inbound mail. `coordinator` has no equivalent rule.
2. **Wrong subject.** The rule keys on the contact's `kind`. The sender's record correctly reads `kind=person`. Their Exchange client sent the mail.

Meanwhile the signal was already computed and discarded. `isCalendarInviteAttachment()` (`src/channels/email/nylas-client.ts:619`) sets `attachments[].isCalendarInvite = true`, and grepping the tree shows it is written at `nylas-client.ts:585` and **read nowhere**. The `method=REPLY` parameter on the content type is likewise available and unused.

The affected category is large and ordinary: RSVP replies, out-of-office, read receipts, delivery/bounce notices, subscription confirmations.

## Proposed direction (not prescriptive)

- Derive an auto-generated flag from the **message**: `Auto-Submitted` header (RFC 3834), `Precedence: bulk/auto_reply`, `X-Auto-Response-Suppress`, empty body plus a `text/calendar` part with `method=REPLY`, and the existing `isCalendarInvite` signal.
- Surface it on the inbound event so dispatch and agents can both see it, and let it suppress replies by default regardless of sender `kind`.
- Keep the existing actionable-signal carve-outs from the `ceo-inbox` rule (bounce, payment failure, security alert, hard deadline) so genuinely important machine mail still escalates.
[...]

PR #1736, opened 2026-09-04
fix(dispatch): honour NO_REPLY so agents can decline to send

### LINKED ISSUE #1732: Agent cannot decline to reply — dispatcher relays its closing text unconditionally, and the content-block handler forces a rewrite
## Summary

The dispatcher turns every `agent.response` into an outbound reply, and the content-filter block handler can only produce a reply. Together these make "send nothing" unrepresentable — an agent that concludes *nothing needs to go out* has no way to say so, because its conclusion **is** the payload.

On 2026-09-04 this sent 14 unwanted emails to a `known`-tier contact in 20 minutes, including the agent's own internal monologue, verbatim:

> "This is an automated calendar decline notification — no reply needed from me. I'll just archive it."
> — `outbound.message` → `<known-tier contact>`, 07:55:03 ET

## Context

**F2 — the relay is unconditional.** `handleAgentResponse` (`src/dispatch/dispatcher.ts:996`) publishes an `outbound.message` for every agent response that has a routing entry. The only suppressor is the reply-lock (`routing.humanReplySent`), which fires solely when `email-reply` / `email-send` already succeeded. There is no path where the agent finishes a turn and nothing is sent.

**F3 — the block handler backfires.** When the outbound judge blocks a relayed reply, `buildContentBlockRewriteTask` (`src/dispatch/content-block-relay.ts:44`) schedules a rewrite whose prompt admits exactly one outcome:

```
[OUTBOUND CONTENT FILTER — REWRITE REQUIRED]

Your previous reply was blocked before delivery. Rewrite it to satisfy the
findings below, applying your normal audience and voice rules. Return ONLY
the corrected reply text — no preamble about the block.
```

An `llm-judge-audience-leak` block means *this should never have been addressed to this person*. To the agent that is indistinguishable from *fix the wording*. In the incident all 10 blocks produced polished, recipient-appropriate acknowledgements that then passed the judge and were delivered.

**The guardrail manufactured the emails it existed to prevent.** Every delivered "Thanks for the heads-up" exists because a block fired first.

- A structured no-reply outcome the relay honours — a sentinel the agent can return, or an explicit reply/no-reply decision in the response envelope — so `handleAgentResponse` can legitimately send nothing.
- Extend the rewrite prompt with the matching escape: *return corrected text, **or** `NO_REPLY` if this should not have been addressed to this recipient at all.*
- Emit an audit event when a turn ends in no-reply, so silence is observable rather than indistinguishable from a dropped message.

Either half is independently useful: fixing F3 alone stops the reply storm; fixing F2 alone stops the monologue leak. Doing both is what makes silence a first-class outcome.

## Acceptance criteria

- [ ] An agent can end a turn without any `outbound.message` being published, via an explicit and documented mechanism.
- [ ] The content-block rewrite prompt offers an abandon-send option alongside rewrite, and the dispatcher honours it without falling through to the salvage draft.
- [ ] A blocked `llm-judge-audience-leak` reply that the agent abandons produces no delivered message and no further retry.
- [ ] Choosing no-reply is recorded in the audit log (distinct from `outbound.suppressed_duplicate`).
- [ ] Unit tests cover: no-reply on a normal turn, no-reply from the rewrite path, and the existing reply-lock behaviour unchanged.
- [ ] Existing behaviour is preserved when the agent does return reply text.

## Evidence

Prod `audit_log`, 2026-09-04 (thread `email:1a06c376fb0bf152` received 5 acknowledgements in 7 minutes). Block IDs `block_f037ac81`, `block_29b0c38f`, `block_fe9cb69b`, `block_01bd5f7e` (+6).

josephfung:
## Summary

The dispatcher turned every `agent.response` into an `outbound.message`, and the content-block rewrite prompt could only produce a reply. An agent that concluded *nothing should go out* had no way to say so — its conclusion **was** the payload. On 2026-09-04 that sent 14 unwanted emails, including internal monologue, and every audience-leak block was polished into a recipient-appropriate acknowledgement that then delivered.

This PR makes silence a first-class outcome: the agent returns `NO_REPLY`, the dispatcher publishes `outbound.no_reply` instead of sending, and the rewrite path offers the same escape so abandon-send skips the salvage draft.

## Changes

- Detect the `NO_REPLY` sentinel in `handleAgentResponse` and skip `outbound.message`
- Emit `outbound.no_reply` (reasons `agent_declined` / `content_block_abandoned`) for the audit log — distinct from `outbound.suppressed_duplicate`
- Extend the content-block rewrite prompt with an abandon-send option; audience-leak blocks prefer `NO_REPLY`
- Document the mechanism in the coordinator prompt and specs 00/02/05/06/10/15
- Unit tests: normal-turn no-reply, rewrite-path abandon, reply-lock unchanged, ordinary reply text still sends

## Related Issues

Closes #1732

## Checklist

- [x] CI passes


josephfung (commented):
## Review: security + architecture

The core move is right, and it is the right *size* of move. Making silence a first-class outcome at the point where the relay decides to publish, rather than trying to teach the content filter to recognise "this shouldn't be a reply", targets the actual root cause of the 2026-09-04 storm. Wiring `outbound.no_reply` through the bus permission allowlist, the field extractor, and the spec set (rather than just early-returning) is the difference between a fix and a silent drop, and I'm glad it was done that way. The `content_block_abandoned` escape on the rewrite path directly disarms the mechanism that manufactured the emails. CI is green across all 11 checks.

What follows is mostly about the *failure modes of the new mechanism*, not the mechanism itself. Two of them I'd want addressed before merge.

### Blocking-ish

1. **Silence toward the principal is enforced only by the prompt.** (`dispatcher.ts:1001`) This PR introduces a new fail-silent behaviour and defends it with exactly the enforcement model whose failure this issue documents. Detail inline.
2. **A near-miss sentinel is delivered as the email body.** (`no-reply.ts:31`) The fallback for imprecise detection is "send the control token to the external recipient", which is the same class of embarrassment as the incident. Detail inline.

### Worth considering

3. In-band sentinel: three subscribers read `agent.response`, only one understands the token (`events.ts:242`).
4. Abandoning on the rewrite path leaves no draft, unlike the retries-exhausted path. Probably correct, but it should be a stated decision rather than a consequence of the early return (`dispatcher.ts:1024`).
5. The sentinel is now a control token sharing a channel with untrusted inbound text, and `InboundScanner` doesn't know it exists (`coordinator.yaml:490`).
6. An empty response is still delivered as an empty email (`no-reply.test.ts:22`).
7. One test asserts absence against a fixture that can silently no-op (`dispatcher-no-reply.test.ts:194`).

### Things I checked and found fine

- No `CHECK` constraint on `audit_log.action` (it is plain `TEXT`, migration 078), so adding `suppress` to the vocabulary needs no migration and cannot break prod inserts.
- The `content_block_abandoned` vs `agent_declined` inference is sound: `handleOutboundBlocked` sets `contentBlockRetryAttempt: nextAttempt` (>= 1) on the rewrite routing entry at `dispatcher.ts:1112`, so `> 0` is a correct discriminator and the first block is attempt 1.
- The early return is correctly placed before `relayOutbound.set`, so an abandoned turn leaves no dangling block-retry context.
- Reply-lock ordering is preserved and tested.
- `event.parentEventId!` is safe here: the routing lookup at the top of `handleAgentResponse` already returned when `parentEventId` was absent.
- The rewrite prompt only says "Prefer NO_REPLY" for audience-leak findings, not for every block. Good restraint, and tested both ways.

src/dispatch/dispatcher.ts
```diff
@@ -993,6 +995,35 @@ export class Dispatcher {
       return;
     }
 
+    // Explicit no-reply (#1732): the agent declined to send. Honour it before publishing
+    // outbound.message so silence is representable — including on the content-block rewrite
+    // path, where returning NO_REPLY abandons delivery without a salvage draft.
+    if (isNoReplyContent(event.payload.content)) {
```
josephfung: **No code-level guard against declining to reply to the principal.**

The only thing stopping the coordinator from returning `NO_REPLY` to the CEO is one sentence at the end of `coordinator.yaml:490` ("Do not use `NO_REPLY` when talking to the CEO").

That is the same enforcement model whose failure produced #1732. The `Do not draft replies to automated senders` rule lives at `agents/ceo-inbox.yaml:664` and has never fired once against live inbound mail, because `dispatcher.ts:826` hardcodes every inbound task to `coordinator`. A prompt-only safety rule in this codebase already has a track record of being structurally unreachable, and this PR adds a *new* behaviour whose failure is invisible by construction.

The failure mode: the CEO asks something over email or Signal, the coordinator returns `NO_REPLY` (misclassifying its own turn, or nudged there by the new rewrite-path pressure to prefer silence), and there is no user-visible signal at all. Not a bounce, not an error, not a delay. Only an `outbound.no_reply` row that nobody is watching. Compare the reply-lock directly above: it can only suppress a *duplicate*, so its worst case is one missing echo of something already sent. This one's worst case is total.

The signal is already computed and cheap to carry. `stampOriginator()` at `dispatcher.ts:812` yields `originator.isPrincipalOriginated` and the off-bag `liveTurn`, both derived from the contact resolver and explicitly un-forgeable from channel metadata. Adding a flag to the `taskRouting` entry (line 852) alongside `humanReplySent` costs one field.

What to do when it fires is a genuine design question and I don't think there's an obvious answer. Options as I see them: refuse the decline and fall through to sending (noisy but never silent); honour it but publish an `outbound.notification` so the CEO learns the turn ended deliberately; or honour it and log at `warn`. My preference is the second, since "the CEO said thanks and Curia correctly said nothing" is a real case worth supporting, and the notification is what distinguishes intentional silence from a hung agent. But any of the three beats prompt-only.

Related: consider whether the same reasoning applies per-channel. On a voice call, silence is a much worse outcome than in email, and `routing.channelId` is right there.

src/dispatch/no-reply.ts
```diff
@@ -0,0 +1,32 @@
...
+export function isNoReplyContent(content: string): boolean {
+  let s = content.trim();
+  if (s.length >= 2) {
+    const first = s[0]!;
+    const last = s[s.length - 1]!;
+    if ((first === '`' || first === '"' || first === "'") && first === last) {
+      s = s.slice(1, -1).trim();
+    }
+  }
+  return s.toUpperCase() === NO_REPLY_SENTINEL;
```
josephfung: **When detection misses, the fallback is to email the control token to an external contact.**

Matching is exact after `trim()` plus one symmetric quote layer. Everything else falls through to `createOutboundMessage` with the model's raw text as the body. So the near-misses all deliver:

- `` ```NO_REPLY``` `` (a markdown fence, one of the most common shapes a model wraps a literal token in) strips one backtick and compares `` ``NO_REPLY`` `` -> no match -> the recipient gets the fenced token.
- `NO_REPLY.` -> no match -> delivered.
- `NO_REPLY\n\n(archiving, no action needed)` -> no match -> delivered.

The PR's own test at `tests/unit/dispatch/no-reply.test.ts:20` codifies this: `'NO_REPLY — I will archive this.'` is asserted `false`, meaning the specified behaviour is to email that string to the external contact. That is the incident again, with a different string. Internal control text arriving in a spouse's inbox is exactly what #1732 is about, and the response that produces it is one where the agent's *intent to stay silent is unambiguous*.

The asymmetry matters: a false positive costs one unsent reply that the audit log records and the block notification surfaces. A false negative sends `NO_REPLY` to a real person. Those are not the same price, so the tie-break should not be "fall through to send".

A reasonable shape: if the trimmed content *starts with* the sentinel on a word boundary but isn't an exact match, treat it as an ambiguous decline. Suppress the send, log at `warn` with the full content so it's diagnosable, and (email only) route it through `publishContentBlockSalvage` so the text lands as a draft for human review instead of in someone's inbox. Fail toward silence plus a draft, never toward delivering the token.

Belt-and-braces alternative, if you'd rather not widen the matcher: a guard on the send path that refuses to publish any `outbound.message` whose body contains the sentinel as a standalone token. That one is cheap, has no false-positive cost worth worrying about, and would have caught every variant above.

src/bus/events.ts
```diff
@@ -234,6 +234,20 @@ interface OutboundSuppressedDuplicatePayload {
   reason: 'human_reply_already_sent';
 }
 
+// OutboundNoReplyPayload — emitted by the dispatch layer when an agent.response would have
+// produced an outbound.message, but the agent explicitly declined to reply (entire content
+// is the NO_REPLY sentinel). Distinct from outbound.suppressed_duplicate: nothing was sent
+// on this turn, whereas the reply-lock fires after a human-facing skill already succeeded.
+// See #1732.
+interface OutboundNoReplyPayload {
```
josephfung: **In-band signalling: three subscribers read `agent.response`, only one understands the token.**

`agent.response` is consumed at `dispatcher.ts:209`, `scheduler.ts:166`, and `agents/resumable-continuation-subscriber.ts:36`. The sentinel is interpreted in exactly one of them, so today a scheduled job whose agent declines records the literal string `NO_REPLY` as its run summary (`scheduler.ts:172` takes `content.slice(0, 500)` and `completeJobRun` writes it via `COALESCE`), and the conversation checkpoint stores it as the agent's utterance in working memory. Neither is harmful, but both are the same bug shape as the delivery case above: a control token leaking into a content field because it was never lifted out of one.

The payload already has the precedent for the alternative. `isError` is a structured boolean on `AgentResponsePayload` rather than a magic string in `content`, and every consumer can act on it without agreeing on a parser. A `suppressDelivery` boolean set once by the agent runtime when it detects the sentinel would give this the same shape: consumers read a flag, `content` stops carrying control semantics, and the string match exists in exactly one place instead of implicitly in three.

I don't think that needs to block the PR. The sentinel is the pragmatic choice and the coordinator already returns bare reply text, so there's no clean out-of-band channel to reach for without touching the runtime. But if it stays in-band, the minimum I'd want is that the dispatcher blanks `content` on the no-reply path before anything else can read it, so the token can't reach a summary field or a memory turn. Worth a line in spec 02 either way, since "the sentinel is only interpreted by dispatch" is currently a fact you can only learn by grepping the subscribers.

src/dispatch/dispatcher.ts
```diff
@@ -993,6 +995,35 @@ export class Dispatcher {
...
+        routingTaskId: event.parentEventId!,
+        agentId: event.payload.agentId,
+        conversationId: routing.conversationId,
+        channelId: routing.channelId,
+        reason,
+        parentEventId: event.id,
+      });
+      await this.bus.publish('dispatch', noReply);
+      this.scheduleCheckpoint(routing.conversationId, event.payload.agentId, routing.channelId);
+      return;
```
josephfung: **Abandoning on the rewrite path leaves nothing behind; retries-exhausted leaves a draft.**

This early return skips `publishContentBlockSalvage`, so a `content_block_abandoned` turn preserves the blocked text only inside the `outbound.blocked` audit payload. The sibling path at `dispatcher.ts:1128` saves it as a real draft in the mailbox.

I think this is probably the right call: the whole point of the abandon option is that the message should not exist, and a draft is a standing invitation to send it. Spec 15 states the behaviour clearly. But right now it reads as a consequence of where the `return` landed rather than a decision that was weighed, so I'd like it recorded as the latter.

The case that makes me want it explicit is a wrong abandon. If the agent misjudges and declines something that genuinely needed a reply, the retries-exhausted path leaves a recoverable draft and this path leaves nothing a human would ever stumble across. Joseph does still get the `outbound.blocked` notification at block time, so it isn't invisible, but the notification points at a message that no longer exists anywhere he can act on.

A question rather than a request: is the block notification considered sufficient closure for the abandoned case, or should `content_block_abandoned` carry the abandoned content into the audit payload so it's recoverable without a log dig? Whichever way, worth a sentence in spec 15 saying the salvage skip is deliberate and why.

agents/coordinator.yaml
```diff
@@ -482,6 +487,7 @@ system_prompt: |
 
   #### Sending and replying
   - **Replying to a direct inbound email**: just compose your reply text and return it as your response. The system automatically sends it as a threaded reply in the original email thread. Do NOT call `email-reply` or `email-send` for this — calling those skills creates a new thread.
+  - **When nothing should go out**: return exactly `NO_REPLY` as your entire response. The system will not send an outbound message. Use this for automated notifications, archive-only FYIs, and any turn whose correct outcome is silence. Never narrate "no reply needed" — that narration is delivered as the email. Do not use `NO_REPLY` when talking to the CEO; they expect a response.
```
josephfung: **The sentinel is now a control token sharing a channel with untrusted inbound text.**

Inbound email bodies land in the coordinator's task content, and the system prompt now teaches the model that a specific literal token means "send nothing". An external sender can write `NO_REPLY`, or "reply with exactly NO_REPLY", into an email body.

I want to be clear about how much this is and isn't worth: it is *not* a serious escalation. Injection could already steer reply content, which is strictly worse than suppressing a reply, and the blast radius here is bounded to the attacker silencing Curia's reply to themselves. CEO notifications go out via `outbound.notification`, a separate path this cannot touch, so it can't be used to hide activity from Joseph. Low severity, and I wouldn't hold the PR for it.

The cheap mitigation is that `InboundScanner`'s `DEFAULT_INJECTION_PATTERNS` (`src/dispatch/inbound-scanner.ts:33`) has no entry for the token, so an inbound containing it scores zero injection risk and produces no finding. Adding one pattern means the token showing up in inbound text lands in `messageTrustScore` and the audit findings. Worth noting the scanner only scores, it doesn't block, so this buys visibility rather than enforcement, which is the right trade for a risk this size.

Separately on this prompt block: `NO_REPLY` is documented only in `coordinator.yaml`. That's correct for now (it's the only agent whose responses the dispatcher routes for live inbound), but it does mean the mechanism is coordinator-only in practice while `handleAgentResponse` will honour it from any agent. When #1734 lands and auto-generated detection gets somewhere to send its verdict, that asymmetry is worth revisiting rather than inheriting.

tests/unit/dispatch/no-reply.test.ts
```diff
@@ -0,0 +1,25 @@
...
+  it('matches a single layer of quotes or backticks', () => {
+    expect(isNoReplyContent('`NO_REPLY`')).toBe(true);
+    expect(isNoReplyContent('"NO_REPLY"')).toBe(true);
+    expect(isNoReplyContent("'NO_REPLY'")).toBe(true);
+  });
+
+  it('rejects narration that merely mentions the sentinel', () => {
+    expect(isNoReplyContent('NO_REPLY — I will archive this.')).toBe(false);
+    expect(isNoReplyContent('This is an automated notification — no reply needed.')).toBe(false);
+    expect(isNoReplyContent('')).toBe(false);
```
josephfung: **Empty content is the sibling case and is still delivered.**

`isNoReplyContent('')` returning `false` is correct as written, and I'd keep the assertion. But it does mean a model that declines by returning nothing at all publishes an `outbound.message` with an empty body, which the gateway will happily deliver as a blank email.

That's pre-existing rather than introduced here. I'm raising it because this PR is the natural home for it: an empty response isn't a reply either, and the whole premise is that "nothing to say" needed a representation. Suppressing whitespace-only content down the same path (with its own reason, something like `empty_response`, so it stays distinguishable from a deliberate decline in the audit log) would close the last way a turn can send a non-message.

tests/unit/dispatch/dispatcher-no-reply.test.ts
```diff
@@ -0,0 +1,196 @@
...
+    seedRouting(dispatcher, 'task-nr-cleanup');
+    await fireAgentResponse(subscribeHandlers, {
+      taskEventId: 'task-nr-cleanup',
+      content: NO_REPLY_SENTINEL,
+    });
+
+    const taskRouting = (dispatcher as unknown as {
+      taskRouting: Map<string, unknown>;
+    }).taskRouting;
+    expect(taskRouting.has('task-nr-cleanup')).toBe(false);
```
josephfung: **This assertion passes whether or not the fixture worked.**

`seedRouting` writes to `dispatcher.taskRouting` through an `as unknown as` cast (line 60-67). If that private field is renamed or reshaped, the cast keeps compiling and the seed silently writes to nothing. Most tests in this file survive that, because they assert a no-reply event *is* published and would fail loudly. This one asserts `taskRouting.has(...)` is `false`, which is exactly what an empty map returns, so it would keep passing while testing nothing.

The rest of the file's assertions are also mostly absence-shaped (`toHaveLength(0)` on outbound messages), which is the pattern that tolerates a broken fixture most easily. Worth being deliberate about which assertions are load-bearing.

`dispatcher.test.ts` uses the public `registerExternalTaskRouting` seam for the same job, and the end-to-end rewrite test added at `dispatcher.test.ts:2196` already exercises `contentBlockRetryAttempt` through the real block flow rather than by hand. That's the stronger test of the two. If `contentBlockRetryAttempt` is the only reason for the cast, it may be worth either exposing it on the public seam or dropping the hand-seeded rewrite case here in favour of the end-to-end one.

josephfung:
@Coderabbitai review

docs/specs/15-outbound-safety.md
```diff
@@ -69,6 +69,42 @@ The gateway emits an `outbound.delivered` event after every successful send. Thi
...
+
+The content-filter rewrite prompt (#1355) offers `NO_REPLY` alongside rewrite. Returning it
+abandons delivery: no send, no salvage draft, no further retry. An `llm-judge-audience-leak`
+block that the agent abandons therefore cannot be polished into a recipient-appropriate
+acknowledgement and delivered — the failure mode of the 2026-09-04 reply storm.
+
+Skipping salvage on an exact rewrite abandon is deliberate, not an accident of the early
+return. A draft is a standing invitation to send the blocked text; the point of abandon is
+that the message should not exist. The blocked body is copied onto `outbound.no_reply`
+(`abandonedContent`) so it is recoverable from the audit row. The `outbound.blocked` CEO
```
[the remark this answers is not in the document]
josephfung: Agreed — “never included, on either path” contradicted both `outbound.blocked.content` (stores `redactedBody`) and `outbound.no_reply.abandonedContent` on exact rewrite abandon.

Spec 15 now states the audit contract explicitly:

- `outbound.blocked.content` stores the PII-redacted body
- `outbound.notification` omits the body
- exact rewrite abandon stores `abandonedContent` on `outbound.no_reply`

The same wording replaced the Stage 1/2 “never included” sentence.

Fix: `ee9de302`.

josephfung:
@Coderabbitai approve

Issue #1059, opened 2026-06-19
Gate C fails open when an external originator has no tier stamped


josephfung:
## Context

While reviewing the Gate C escalation-judge work in #1057 (closes #950), both an independent security review and CodeRabbit flagged the same fail-open in `getInitiatingTier()` (`src/contacts/principal.ts`). It was deliberately deferred out of #1057 to avoid a behavior change mid-migration; this issue tracks closing it.

Related discussion: https://github.com/josephfung/curia/pull/1057#discussion_r3442071995

## Problem

`getInitiatingTier(metadata)` returns `null` for three distinct cases, and Gate C is skipped (fail-open) whenever it returns `null`:

1. **system**-originated tasks — genuinely no external contact → safe to skip.
2. **agent**-originated tasks — genuinely no external contact → safe to skip.
3. **external** originator that carries no `tier` field (a pre-#950 task in flight during a deploy, or a dispatcher stamping defect) → **a consequential external action bypasses tier enforcement entirely**, including the escalation judge.

Cases 1–2 are correct. Case 3 is a fail-open: an externally-initiated, consequential action (e.g. `email-send` / `signal-send`) auto-executes with no tier gate. Now that #1057 makes Gate C the authoritative third-party-facing enforcement point, a `null`-tier fall-through defeats all of it.

The current behavior is also **silent** — the skip logs nothing, so a stamping defect would be invisible.

## Proposed approach

Distinguish "structurally not externally-initiated" from "external contact with a missing tier":

- `systemRole === 'system' | 'agent'` → skip Gate C (unchanged, safe).
- External originator (any other `systemRole`) with `tier == null` → **fail closed** (escalate/block), not skip.
- Emit a `warn` log on the fail-closed path (skill name + `action_risk`) so an unstamped-tier external task is observable rather than silent.

Note the migration consideration: confirm the dispatcher now always stamps `tier` for external originators before flipping this, so legitimate in-flight tasks don't all start escalating. If a backward-compat window is still needed, gate the fail-closed behavior so only genuinely external-with-missing-tier tasks escalate.

## Acceptance criteria

- [ ] An externally-originated task (`systemRole` not `system`/`agent`) with a missing/`null` `tier` and a non-`none` `action_risk` **fails closed** (Gate C escalates; the skill handler does not execute).
- [ ] `system`-originated and `agent`-originated tasks still skip Gate C (no regression).
[...]

Issue #972, opened 2026-06-13
secret.captured event + agent resume after secret capture (follow-up to #971)


josephfung:
## Summary

Follow-up to #971. Today secret capture is fire-and-forget: the agent mints a link, the user fills it, and nothing tells the agent. This issue adds a `secret.captured` event and a thin **resume** path so an agent that asked for a secret is brought back into the conversation when it arrives, and can continue what it was blocked on.

Target UX:

```
User:  please log into my Aeroplan account and check my balance
Curia: happy to. I'll need your username and password — two secure links:
       <link 1: username>  <link 2: password>
[user fills link 1]
Curia: got your username, still waiting on the password.
[user fills link 2]
Curia: got the password. checking your balance...
Curia: your Aeroplan balance is 42,150 points.
```

The secret **value** never flows through the LLM, the event, or the resume task — only secret names/labels and routing metadata do (consistent with #971's structural guarantee). The agent reads actual values via `ctx.secret()` only at the moment it performs the action.

**Design note:** there is no group/coordination layer. Each capture link is independent. When a capture arrives, the agent is re-entered and reasons about completeness from its own conversation history — it asked for "username and password," it now sees "username captured," so it knows it's still waiting on the password. We don't duplicate that bookkeeping in a DB table; the LLM already has the context.

## Depends on
- #971 (secret-capture skills, token table, capture endpoint) — this builds directly on that token store.
- #973 (Consume) — the consumer skill that reads `user.*` secrets and performs the action. Out of scope here; this issue only delivers the event + resume plumbing.

## Approach

Model on the existing **approval-then-resume** lifecycle (`src/autonomy/approval-trigger.ts`, `skills/approve-action/handler.ts`) and the **scheduler's synthetic-task** pattern (`src/scheduler/scheduler.ts`). State is parked in the DB; the agent is re-entered by publishing a new `agent.task` that threads `conversationId` / `parentEventId` — not by holding a live process.

### 1. Store routing context on the token (builds on #971)
No new table. Extend #971's `secret_capture_tokens` with the origin context needed to route a resume back to the right agent: `conversation_id`, `channel_id`, `agent_id`, `task_event_id`, `originator`, and an optional `resume_intent TEXT` (natural-language description of what to resume, e.g. the original user ask).

Origin context is captured at mint time from `SkillContext` (`conversationId`, `agentId`, `channelId`, `taskEventId`, `caller`, `taskMetadata.originator` — all already available per `src/skills/types.ts`).

[...]

Issue #847, opened 2026-06-02
dispatcher: suppress duplicate outbound.message when human-facing reply already sent


josephfung:
## Background

On Jun 1, 2026 Joseph received **three replies in 73 seconds** on a single email thread (Monthly Reconciliation — May 2026 Statements). Full incident reconstruction: `~/.claude/plans/help-me-debug-a-majestic-tower.md`.

The three sends were:

| Time (UTC) | Sender                | Path                          |
|------------|-----------------------|-------------------------------|
| 20:32:18   | T2125-expense-tracker | direct `email-reply` skill    |
| 20:33:22   | coordinator           | direct `email-reply` skill    |
| 20:33:31   | coordinator           | `agent.response` → dispatcher |

## Root cause (this issue)

`src/dispatch/dispatcher.ts:838` `handleAgentResponse` publishes `outbound.message` for every `agent.response` whose task has a routing entry. The coordinator's wrap-up `agent.response` therefore always produces an extra outbound — even when the coordinator already called `email-reply` directly during the run. No "reply already sent on this conversation" flag is checked.

The (a)+(b) case — T2125 and coordinator both calling `email-reply` — is left as a follow-up (specialist email policy). This issue covers only the dispatcher reply-lock that fixes case (c).

## Scope

1. Add `humanReplySent: boolean` to the routing entry in `taskRouting` (`dispatcher.ts:817`).
2. Subscribe to `skill.result` events; when an `email-reply` or `email-send` succeeds with a recipient that matches `routing.senderId`, set `humanReplySent = true` on the routing entry for the task chain root (walk `parentEventId` / `originator` ancestry to find the routed root).
3. In `handleAgentResponse`: if `routing.humanReplySent === true`, **skip** the `bus.publish('dispatch', outbound)`. Emit an `outbound.suppressed_duplicate` audit event with `{ routingTaskId, agentId, conversationId, reason: 'human_reply_already_sent' }`.
4. The `agent.response` payload still goes to `audit_log` — nothing is lost, only the duplicate send.
5. Do not modify the `email-reply` skill handler. Ownership of dedup lives in the dispatcher.

## Acceptance criteria

- [ ] When a routed agent.task calls `email-reply` successfully, the subsequent `agent.response` on the same task does NOT produce an `outbound.message`.
- [ ] Exactly one `outbound.suppressed_duplicate` audit event is recorded in that case, with the structured payload above.
[...]

PR #1737, opened 2026-09-04
fix(dispatch): enforce Gate C on the auto-reply relay path

### LINKED ISSUE #1733: Dispatcher relay bypasses Gate C — blocking a reply skill diverts the send onto an ungated path
## Summary

Gate C (the contact-tier action gate) lives in the skill execution layer. The dispatcher's auto-reply relay does not go through skill execution, so it carries **no tier check at all** — and blocking the gated path is precisely what diverts the send onto the ungated one.

Observed 2026-09-04 on the Easter Monday thread:

```
07:40:40  tool.result   email-reply   BLOCKED  (Gate C → approval d3686451)
07:40:59  outbound.message    source_id=email
07:41:02  outbound.delivered  source_id=dispatch  → <known-tier contact>
```

The escalation was honoured for 22 seconds. The CEO dismissed approval `d3686451` at 08:01:59 — 21 minutes after the message it was gating had already been delivered. After that first block the agent stopped calling `email-reply` entirely and let the relay carry the remaining 14 replies.

## Context

- `handleAgentResponse` (`src/dispatch/dispatcher.ts:996`) publishes `outbound.message` directly to the bus.
- The message still passes `OutboundGateway`, but the gateway enforces only the **autonomy score** (Step 0) and the **content filter / judge**. It never consults the initiating contact's tier. Grep for `applyActionPolicy` in `src/skills/outbound-gateway.ts` returns nothing.
- Gate C's `resolveGateCDecision` / `applyActionPolicy` (`src/skills/execution.ts`, `src/autonomy/escalation-policy.ts`) are reachable only via skill invocation.
- Net effect: any external contact whose message reaches an agent gets a reply regardless of tier, and the Gate C escalation on `email-reply` is decorative.

This is the wrong direction for a guardrail to fail in — denying the audited path should not silently open an unaudited one.

- Apply the Gate C tier decision to the relay path before publishing `outbound.message`, treating a relayed reply as the equivalent of `email-reply` (`action_risk: medium`, reply-to-sender vs third-party-facing distinction intact).
- Alternatively, route the relay through the same execution-layer entry point the reply skills use, so there is one gated send path rather than two.
- Whichever shape: an escalation on the skill must not be satisfiable by falling through to the relay. If the relay escalates, it should create/join the same pending approval rather than issuing a second one.

Interacts with the no-reply work (see the F2/F3 issue) and with #1530 (`OutboundAdapter` interface / collapsing gateway `send()` dispatch), which may be the natural place to land a single choke point.

## Acceptance criteria

- [ ] A dispatcher-relayed reply to a non-principal contact is subject to the same tier decision as an equivalent `email-reply` call.
- [ ] When Gate C escalates a reply skill, the relay for that same turn cannot deliver the message instead.
- [ ] An escalated relay does not create a duplicate approval request for an approval already pending on the same turn.
- [ ] Principal-tier and principal-sole-recipient behaviour is unchanged (no new friction on CEO-facing replies) — ADR-034/ADR-035 carve-outs still hold.
- [ ] System notifications and infrastructure alerts to the principal are unaffected.
- [ ] The gate decision on the relay path is auditable (`authorization.decision` emitted with the relay as source).
- [ ] Tests cover: known-tier third-party relay escalates, principal relay allows, and a skill-blocked turn does not deliver via relay.

## Evidence

Prod `audit_log`, 2026-09-04 07:39–08:02 ET. 24 relayed replies attempted to a `known`-tier contact, 14 delivered, zero Gate C decisions recorded on the relay path.

josephfung:
## Summary

Closes #1733.

The dispatcher auto-reply relay (`handleAgentResponse` → `outbound.message`) had no Gate C check. Blocking `email-reply` on a known-tier contact diverted the send onto that ungated path — the 2026-09-04 incident pattern.

## Changes

- Apply Gate C to every dispatcher-relayed reply (`action_risk: medium`), consulting the EscalationJudge on the known × reversible-external ambiguous cell (same as `email-reply`)
- New `dispatcher-relay` skill so approve-action can re-publish the withheld `outbound.message`
- Enrolled in `config/registry-defaults.yaml` so reconcile/load actually registers it (approve path otherwise still 404s)
- Suppress the relay when a reply skill on the same turn already Gate C–escalated (matched via `authorization.decision` `taskEventId` / `conversationId` — no error-string matching)
- Emit `authorization.decision` with `action: dispatcher-relay`, `sourceLayer: dispatch`, and `conversationId`
- Join only reply-class pending approvals on the same turn (`dedupePendingSkillsOnTask`); exclude expired rows
- Reject oversized relay bodies before creating an approval (handler limit retained as defence in depth)
- Secret-capture resume stamps the same normalized originator on task metadata and routing; `parseOriginator` preserves absent `tier` (does not coerce to `null`)
- HTTP escalate settles `EventRouter` waiters with a terminal pending-approval status (no 120s hang)
- `registerExternalTaskRouting` requires `originator`
- Specs 06 / 09 / 15 + CHANGELOG updated for the relay Gate C path

## Related Issues

Closes #1733

## Checklist

- [ ] CI passes (pending on `328a99b6`)

## Test plan

- [x] Unit / integration coverage for Gate C relay, approve path, registry enrollment, resume originator
- [x] CodeRabbit follow-ups: oversized body, expired dedupe, normalized resume originator
- [x] Fixed CI: `specialist-resume-flow` originator equality when `tier` is absent
- [x] Skipped: rewriting conversation-memory/vertical-slice onto real ContactResolver (out of #1733 scope)

<div><a href="https://cursor.com/agents/bc-01a06d77-f674-7e1c-b448-cb5709296e4b?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-01a06d77-f674-7e1c-b448-cb5709296e4b&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


src/dispatch/dispatcher.ts
```diff
@@ -1074,6 +1230,104 @@ export class Dispatcher {
...
+    senderId: string;
+    content: string;
+    tier: import('../contacts/types.js').ContactTier | 'unresolved';
+  }): Promise<void> {
+    if (!this.approvalTrigger) return;
+    try {
+      const result = await this.approvalTrigger.request({
+        taskId: args.taskEventId,
+        conversationId: args.conversationId,
+        toolName: RELAY_GATE_C_ACTION,
```
josephfung: **The approval this creates cannot be fulfilled, and approving it destroys the message.**

`toolName: RELAY_GATE_C_ACTION` writes `dispatcher-relay` into `autonomy_action_log.tool_name`. Follow what `approve-action` does with that row (`skills/autonomy/tools/approve-action/handler.ts:50-72`):

1. `resolveRow(row.id, 'approved', 'ceo')` — the row leaves `pending_approval`. This is a one-way transition and it happens *before* any execution.
2. `ctx.executionLayer.invoke(row.toolName, row.payload, …)` with `row.toolName === 'dispatcher-relay'`.
3. There is no such skill, so `execution.ts:725` returns `Tool 'dispatcher-relay' not found in registry`.
4. A child `failure` row is written and the handler returns.

So the CEO gets "Approval needed — Curia wanted to relay a reply", approves it, sees an error, and the reply is gone. It cannot be retried: the row is `approved`, so it no longer matches any pending query, and the dispatcher's in-memory `taskRouting` entry was deleted at the top of `handleAgentResponse` long before. `deny` and `dismiss` at least end honestly; `approve` is the one verb that promises delivery and can't deliver.

The content isn't lost in the database — it's sitting in the approval's `input.body` — so the missing piece is a resume path, not the data. Some shape has to exist for the approved relay to actually publish the withheld `outbound.message`. Whether that's a real registered skill, a dispatcher-side subscriber on `human.decision`, or reusing the existing skill (record the row as `email-reply` with the resolved recipient so the normal re-execution path works) is a design call I don't want to make for you, but the current shape can't be the answer.

This is what makes finding 2 severe rather than merely noisy. On its own, "known-tier replies now require approval" is a policy change people could live with. Combined with this, it means routine correspondence with a known contact silently stops working.

src/dispatch/relay-gate-c.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+    return { kind: 'skip', reason: 'system_or_agent' };
+  }
+
+  const actionClass = mapActionRiskToConsequenceClass(RELAY_ACTION_RISK);
+  const decisionIfReplyToSender = applyActionPolicy(initiatingTier, actionClass, false, false);
+  const decisionIfThirdParty = applyActionPolicy(initiatingTier, actionClass, true, false);
+
+  // Ambiguous cell (known × reversible-external): fail closed without a judge.
+  if (decisionIfReplyToSender !== decisionIfThirdParty) {
```
josephfung: **This resolves the ambiguous cell by always escalating, where the execution layer resolves it with the judge.**

Lines 77-81 are a faithful copy of `resolveTierGateDecision` (`src/skills/execution.ts:602-606`) right up to the branch. The execution layer's next step is `execution.ts:623`: consult the `EscalationJudge` to determine the third-party axis, then re-apply the policy. Here that becomes an unconditional `escalate`.

The judge is on by default in production (`config/default.yaml:274-278`, `enabled: true`, haiku, 5s timeout). So for the same content and the same contact, `email-reply` asks the judge and frequently allows, while the relay always escalates. The first acceptance criterion on #1733 is "subject to the **same** tier decision as an equivalent `email-reply` call" — this is the same inconsistency the issue is about, pointing the other way. The module docstring's "without re-running the EscalationJudge on this hot path" is the whole substance of the divergence, and 5s on a path that is already an async LLM turn doesn't read as a hot path to me.

The scale matters. `applyActionPolicy`'s own policy note says "known — reply to sender is fine" and "a plain reply to the initiating sender is governed by the DISCLOSURE gate (what the reply says), not this action gate, to avoid double-counting the same message." `known` is the tier ordinary business contacts land in, and `agents/coordinator.yaml` explicitly instructs the agent to return reply text for the relay rather than call `email-reply` for direct inbound. So the escalating path is the default path for the modal tier.

**That said, I don't think hardcoding `isThirdPartyFacing: false` is right either, and the PR is correct not to.** `outbound.message` carries `recipientId = routing.senderId`, which reads as reply-to-sender, but the email adapter reply-alls: `src/channels/email/email-adapter.ts:735-742` builds `ccAddresses` from the thread message's `to[]` and `cc[]` and attaches them to the send. On a multi-party thread a relayed reply genuinely does reach third parties. Good catch, and it's the reason the naive fix is wrong.

Which points at the actual problem: **the gate is being applied at a layer that cannot see the recipient set.** The dispatcher knows only the sender; the CC list doesn't resolve until the adapter. So failing closed here fails closed on 1:1 threads too, where there is no third party by construction. Three directions I'd consider, roughly in order of how much I like them:

- Decide the axis where the recipients are known — the adapter or the gateway — which is also where #1530 is taking the send path anyway. One gated choke point was the issue's second proposed direction for exactly this reason.
- Consult the judge on the relay path the way the execution layer does, accepting the latency.
- Have the dispatcher carry an "axis unresolved" marker forward rather than collapsing it to `escalate`, and let the layer that knows the CC set finish the decision.

Worth noting the current behaviour isn't only over-strict: it's over-strict on the safe case (1:1) for the same reason it can't distinguish the unsafe one, so it isn't buying much safety for the friction it adds.

src/dispatch/relay-gate-c.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/**
+ * True when a failed tool.result error string indicates a Gate C (tier) block
+ * rather than a score gate or handler failure.
+ */
+export function isGateCBlockError(error: unknown): boolean {
+  if (typeof error !== 'string') return false;
+  return (
+    error.includes("initiating contact's tier") ||
```
josephfung: **A security control keyed on the wording of an error message.**

These four `includes()` checks are matching human-readable prose produced elsewhere in the codebase. If someone rewords a Gate C error in `execution.ts` — clearer copy, i18n, a refactor that wraps the message — `isGateCBlockError` returns `false`, `replySkillGateCBlocked` is never set, and the relay delivers a message whose skill-path equivalent was just denied. That is #1733 verbatim, reintroduced silently.

Nothing would catch it. The unit test at `dispatcher-relay-gate-c.test.ts` asserts against strings the test file authors itself, so it can't detect drift from the real emitter. A test that imports the actual error construction, or better a structured discriminator (an error code on the tool result, or a typed `blockedBy: 'gate_c'` field), turns this from a coupling that rots quietly into one the compiler or a test can hold.

Worth asking whether the string matcher needs to exist at all. You already subscribe to `authorization.decision`, which is structured and unambiguous. The comment at `dispatcher.ts:1054` says the two cover different join keys — `taskEventId` for the direct-coordinator path, `conversationId` for delegated specialists. If the authz event carried the routing key (or the dispatcher could map `conversationId` → routing entries the same way `handleToolResult` does), the structured path could cover both and the prose matching could go away entirely. That seems worth checking before hardening the string list.

Smaller note on the same mechanism: `handleToolResult` marks *every* routing entry matching the conversation. For concurrent turns in one conversation that's broader than intended, and `replySkillGateCBlocked` is never cleared once set.

src/autonomy/action-log-repo.ts
```diff
@@ -192,6 +192,24 @@ export class ActionLogRepo {
     return mapRow(result.rows[0]);
   }
 
+  /**
+   * Find any pending_approval row for a task, regardless of skill/payload (#1733).
+   * Used by the dispatcher relay so a Gate C escalate joins an approval already
+   * pending from a reply skill on the same turn instead of creating a second one.
+   */
+  async findAnyPendingByTask(taskId: string): Promise<ActionLogRow | null> {
```
josephfung: **This joins any pending approval on the task, including ones for entirely unrelated actions.**

The query filters on `task_id` and `outcome = 'pending_approval'` only — no skill or payload predicate — and `ORDER BY created_at ASC LIMIT 1` picks the *oldest* pending row. So the relay's escalate attaches itself to whatever escalated first on that turn, whatever it was.

That isn't hypothetical for this issue. In the 2026-09-04 incident, `calendar-update-event` approvals were pending on the same turns as the blocked replies — that's the other half of what Joseph was seeing that morning. Under this code the relay escalate would find the calendar approval, return `{ created: false, reason: 'duplicate' }`, and `dispatcher.ts` would log "relay escalate joined existing pending approval" with the calendar row's short ref. Approving it re-runs `calendar-update-event`. The reply is never sent, nothing on either row records that a second and different action was folded in, and the log line says the opposite of what happened.

The acceptance criterion this implements is narrower than the flag: "an escalated relay does not create a duplicate approval request **for an approval already pending on the same turn**" was about the reply the skill path had already requested approval for, not about any approval whatsoever. Scoping the predicate to reply-class tools (`REPLY_SKILLS_GATE_C` plus the relay action) would match the intent. If a genuine join does happen, it'd also be worth recording it on the existing row, so approving that row is *known* to cover both actions rather than the coverage being implied by a log line.

Separately: `dedupeAnyPendingOnTask` runs an extra query before the existing `findPendingByTaskAndSkill` check rather than replacing it, so the relay path now does two round trips where one would do. Minor, but the ordering means the broad check always wins over the precise one.

tests/integration/http-api.test.ts
```diff
@@ -420,23 +431,19 @@ describe('Trust scoring — unknown sender via HTTP', () => {
   it('agent.task carries messageTrustScore for unknown HTTP sender', async () => {
     const tasks: AgentTaskEvent[] = [];
 
-    // Subscribe before injecting the request so we don't miss a fast delivery.
-    // bus.subscribe() returns void (no unsubscribe) — the subscription is scoped
-    // to this describe block's bus instance which is discarded after afterAll.
+    // Subscribe before publishing so we don't miss a fast delivery.
+    // Publish inbound.message directly on the bus — POST /api/messages waits for
+    // outbound.message, which Gate C withholds for unknown-tier senders (#1733).
```
josephfung: **This comment is documenting a regression, and the test change routes around it rather than resolving it.**

`POST /api/messages` registers a waiter and blocks on it: `eventRouter.waitForResponse(conversationId, RESPONSE_TIMEOUT_MS)` with `RESPONSE_TIMEOUT_MS = 120000` (`src/channels/http/routes/messages.ts:25,54`). The waiter is resolved by `outbound.message`. Gate C now withholds that event for any non-principal sender, and nothing else resolves the waiter — so the caller hangs for two full minutes and then times out, with no indication that an approval is pending.

Rewriting the test to publish `inbound.message` directly on the bus makes the suite green, but it also removes the only coverage of the endpoint's behaviour for an unknown sender, which is precisely the behaviour that changed. The test now exercises trust scoring while stepping around the part of the path this PR altered.

The question worth settling explicitly: in production, do HTTP senders resolve as the principal? If the console authenticates and its `sender_id` maps to the principal contact, this is a non-issue for real traffic and the fix is a comment saying so plus a test at the boundary. If it doesn't — or if any other API consumer exists — then a second surface hangs for 120s on every message. Either way the endpoint arguably shouldn't hang: an escalated relay is a known terminal state for that request, and returning "withheld, approval pending" would be more honest than a timeout. Same reasoning as the CLI/vertical-slice harnesses, which were also switched to principal resolvers rather than exercising the new branch.

josephfung (commented):
## Review: security + architecture

The suppression half of this is well built. `replySkillGateCBlocked` closes the exact hole in the incident (a Gate C escalate on `email-reply` diverting onto the relay), and covering it from *both* `tool.result` and `authorization.decision` is the right belt-and-braces, since the two carry different join keys. Threading `originator` through `RelayOutboundContext` so the content-block rewrite retry stays gated is a detail that would have been easy to miss. Failing closed when `publishRelayGateCDecision` rejects is correct: not being able to prove the decision was audited is the right reason to withhold. And `decideRelayGateC` reuses `getInitiatingTier` / `isExternalOriginatorMissingTier` rather than reimplementing the tier semantics, so the fail-closed carve-out from #1059 stays in one place.

The gating half has a problem I don't think can ship as-is, and it compounds with a second one.

### Blocking

**1. Approving a `dispatcher-relay` escalation cannot work, and destroys the reply.** (`dispatcher.ts:1299`) `approve-action` transitions the row to `approved` and then calls `executionLayer.invoke(row.toolName, …)`. There is no `dispatcher-relay` skill, so that returns `Tool 'dispatcher-relay' not found in registry` and the row is no longer `pending_approval`. Every escalation this PR creates is a dead end that consumes the withheld message. Detail inline.

**2. Every known-tier relay escalates, and known is the modal tier.** (`relay-gate-c.ts:81`) The ambiguous cell is resolved by unconditional escalate rather than by the judge the execution layer consults, so `email-reply` and the relay now return opposite decisions for identical content. `coordinator.yaml` tells the agent to prefer the relay for direct inbound replies, so this makes approval the default for ordinary correspondence. Detail inline, including why I don't think the axis should just be hardcoded to `false` either.

Together these mean: routine reply to a `known` contact → approval request → approve → error → reply gone.

### Worth addressing

3. `isGateCBlockError` is substring matching on error prose, so rewording a message in `execution.ts` silently reopens #1733 (`relay-gate-c.ts:106`).
4. `findAnyPendingByTask` has no skill filter, so a relay escalate can join an unrelated pending approval — including the `calendar-update-event` approvals that were live on these exact turns (`action-log-repo.ts:200`).
5. Non-principal HTTP callers now block for the full 120s `RESPONSE_TIMEOUT_MS` (`http-api.test.ts:436`).
6. Six existing suites were switched to principal resolvers or no originator, so the dispatcher's default test posture is now gate-off (`dispatcher-reply-lock.test.ts:61`).
7. Absent originator means "skip the gate", and one production registration site wasn't updated (`relay-gate-c.ts:57`).
8. No spec updates, and two specs now state the wrong thing (`CHANGELOG.md:74`).

### Checked and fine

- `REPLY_SKILLS_GATE_C` is a strict superset of the old `email-reply` / `email-send` check, and the success-path narrowing back to those two is preserved, so the #847 reply-lock behaviour is unchanged.
- The Gate C block sits after the reply-lock and the #1732 no-reply check, which is the right order: an explicit decline shouldn't manufacture an approval request.
- `handleAuthorizationDecision` correctly filters to `gate_c` + `escalate` + a reply skill before touching routing, so unrelated authz traffic can't set the flag.
- The rewrite path re-seeds `replySkillGateCBlocked: false`, which is right — the relay is suppressed before publishing, so no block can originate from a flagged turn.
- Principal and trusted allow deterministically; the #1301 principal-sole-recipient carve-out is correctly documented as not applying, since the relay recipient is the initiator.

tests/unit/dispatch/dispatcher-reply-lock.test.ts
```diff
@@ -46,12 +46,20 @@ function seedRouting(
...
+      senderId: string;
+      humanReplySent: boolean;
+      replySkillGateCBlocked: boolean;
+    }>;
   }).taskRouting.set(taskEventId, {
     channelId: opts.channelId ?? 'email',
     conversationId: opts.conversationId ?? 'email:thread-abc',
     senderId: opts.senderId ?? 'sender@example.com',
     humanReplySent: opts.humanReplySent ?? false,
+    // originator omitted → Gate C skips (#1733); isolates humanReplySent suppression.
```
josephfung: **Reasonable in isolation; the aggregate is that the dispatcher's default test posture is now gate-off.**

Omitting the originator to isolate `humanReplySent` is a defensible call for this file, and the comment is honest about it. But counting across the PR: `vertical-slice`, `conversation-memory`, and two `http-api` blocks gained principal resolvers; three blocks in `dispatcher.test.ts` gained principal resolvers; this file skips the gate; and one `http-api` test stopped going through the endpoint entirely. After this change, essentially every pre-existing dispatcher test runs with Gate C either disabled or unconditionally allowing.

The consequence is that the relay's non-principal behaviour exists only inside `dispatcher-relay-gate-c.test.ts`, where routing is hand-seeded through an `as unknown as` cast on the private `taskRouting` map. So the gate is tested in isolation and never in composition — nothing covers Gate C interacting with the content-block rewrite retry (which now propagates `originator` through `RelayOutboundContext`, a path with no test), with the #1732 no-reply check ordering, or with the reply-lock.

At least one end-to-end suite running a `known`-tier sender through the real inbound path would cover the interactions and would also have surfaced findings 1 and 2 as behaviour rather than as code reading: a full-pipeline test with a known contact ends in an approval request that can't be approved.

src/dispatch/relay-gate-c.ts
```diff
@@ -0,0 +1,111 @@
...
+ * third-party-facing would change the outcome (known × reversible-external),
+ * fail closed — escalate — matching email-reply with no judge wired. That is
+ * the cell named by #1733's "known-tier third-party relay escalates" criterion
+ * and closes the ungated auto-reply hole for known contacts. Unambiguous cells
+ * (principal / trusted allow; unknown / blocked escalate) decide deterministically.
+ * Principal-sole-recipient carve-out (#1301) does not apply: the relay recipient
+ * *is* the initiator.
+ */
+export function decideRelayGateC(originator: TaskOriginator | undefined): RelayGateCOutcome {
+  if (!originator) {
```
josephfung: **"Absent originator" now means "no gate", so the set of routing-registration sites is the size of the remaining hole.**

The rule itself is right and consistent with the execution layer — `isExternalOriginatorMissingTier`'s docstring is explicit that a totally absent originator is "structurally internal (e.g. the checkpoint processor) and is intentionally skipped, not escalated." No argument there.

But the two layers source it differently, and that's what makes this worth an inventory. At the execution layer the originator comes from the task metadata, which `stampOriginator` sets on every inbound task, so "absent" really does mean internal. At the relay it comes from the `taskRouting` entry, so it's present only if the code that registered the routing passed it. The inbound path does (`dispatcher.ts:908`). `registerExternalTaskRouting` takes it as optional, and its one production caller — the secret-capture resume wiring at `src/index.ts:2913` — wasn't updated, so that path relays ungated.

That particular path is CEO-facing and probably benign, which is exactly why it's worth being deliberate rather than lucky. Making `originator` required on `registerExternalTaskRouting` would turn "did every caller pass it?" into a compile-time question instead of an audit. Failing that, a `warn` when the gate skips on a non-system channel would at least make the residual surface visible in prod logs.

Minor naming point: `reason: 'no_external_originator'` asserts something the function can't know. It hasn't established there was no external originator, only that none was recorded on the routing entry. `originator_absent` would describe what actually happened and would read less reassuringly in a log, which is appropriate here.

CHANGELOG.md
```diff
@@ -71,6 +71,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Fixed
 
+- **Dispatcher relay Gate C** — auto-replies no longer bypass the contact-tier gate after a reply skill escalates. (#1733)
```
josephfung: **Two specs now describe the old architecture, and the checklist item is unchecked.**

The PR moves Gate C into a second layer and introduces a new `authorization.decision` action value, and both places that document the gate's location are now wrong:

- `docs/specs/06-audit-and-security.md:48` — "Records an allow/deny/escalate authorization outcome — at Gate-1/authz (dispatch) and Gate C (**execution**)". Gate C now also fires from dispatch, with `action: 'dispatcher-relay'`, which appears in no spec table.
- `docs/specs/09-contacts-and-identity.md:628` — "the execution layer emits for Gate C".

`docs/specs/15-outbound-safety.md` enumerates the relay's exceptions (reply-lock, no-reply) and doesn't mention a tier gate, so it's now incomplete too. For comparison, #1732 updated six spec files for a smaller behavioural change; this one is the more architecturally significant of the two and updates none.

The changelog line itself also undersells the change: "no longer bypass the contact-tier gate **after a reply skill escalates**" describes only the suppression half. The relay now runs its own tier decision on every turn regardless of whether a skill was involved, which is the part users will actually notice.

The unchecked "CI passes" box in the description is stale — all 12 checks are green as of `4f9b9cc`.

src/dispatch/dispatcher.ts
josephfung: Not fixed in a deployed sense — the skill is real but never gets registered, so `invoke` still returns the same error.

The skill/handler/payload shape is right: `humanApproved: true` bypasses the `allowed_callers` check at `execution.ts:842`, so `allowed_callers: ["system"]` is not a problem, and `buildRelayApprovalInput` matches what the handler reads. But enablement was missed.

Boot only registers tools the registry says are enabled:

- `reconcileOne` (`src/registry/reconcile.ts:50`) inserts an enabled row **only** for names listed in `config/registry-defaults.yaml`. Non-core items are left uninstalled — that's the deliberate security property in the file's header (an uploaded skill must not be able to self-enable).
- `src/index.ts:1132-1134` builds `enabledToolNames` from those rows and passes it to `loadToolsFromDirectory`.
- `src/skills/loader.ts:197-200` skips any discovery not in that set: `logger.info({ skill: disc.name }, 'Skill not enabled in registry; skipping')`.

`dispatcher-relay` is not in `config/registry-defaults.yaml`, and this branch changes nothing under `config/` or `src/db/migrations/`. So `this.registry.get('dispatcher-relay')` is undefined and `approve-action` gets `Tool 'dispatcher-relay' not found in registry` — the same string, the same dead end, the same destroyed reply, just one layer further out.

This also hits the existing prod deployment, not only fresh installs. The one-shot backfill referenced in the defaults header enrolled what was on disk at the time; a newly added tool has no row, so it stays uninstalled.

The tests can't see it: `skills/dispatcher-relay/handler.test.ts` constructs the handler directly, and the dispatcher tests stub `approvalTrigger`. Nothing exercises the registry lookup, which is exactly where the failure is.

Adding `dispatcher-relay` to the `tools:` list is presumably the fix. Worth a moment's thought on whether it belongs there at all, though: that list is described as the set of tools agents get, and this one is infrastructure that only `approve-action` should ever reach. It'll show up in tool listings and prompt surfaces alongside `email-send`. `allowed_callers: ["system"]` keeps agents from *calling* it, so this may be fine, but it's worth being deliberate rather than incidental.

A test that goes through the real registry — enrol from defaults, then `invoke('dispatcher-relay')` — would close the gap the current tests leave open.

src/dispatch/relay-gate-c.ts
josephfung: Verified — this is the right fix.

`decideRelayGateC` is now async and consults `EscalationJudge.classifyAction` on the ambiguous cell (`relay-gate-c.ts:112-140`), clamps via the same policy re-application as the execution layer, and fails closed on `isThirdPartyFacing === undefined` and when no judge is wired (`:148`). `escalationJudge` is threaded through `DispatcherConfig` and wired at `src/index.ts:2887`, so the production path gets the judge that `config/default.yaml:274` already enables.

The part I like most is that the action description explicitly tells the judge the email adapter may reply-all and CCs thread participants. That's the piece of context the judge would otherwise be missing, and it's what makes the answer meaningful rather than a guess about `recipientId`. It also means a 1:1 thread with a known contact can now come back `allow`, which was the friction I was worried about, while a multi-party thread still escalates.

Good that `isThirdPartyFacing: false` wasn't hardcoded. Tests cover judge-allow, judge-escalate, and no-judge-fail-closed separately, so the three branches won't silently collapse into one.

Resolving.

src/dispatch/relay-gate-c.ts
josephfung: Verified — `isGateCBlockError` is gone (no references anywhere in `src` or `tests`), and the structured route covers both join keys as suggested.

`authorization.decision` now carries optional `conversationId`, and `handleAuthorizationDecision` (`dispatcher.ts:1039-1078`) matches on `taskEventId` first, then sweeps by `conversationId` for the delegated-specialist case. The prose matching is fully replaced rather than supplemented, which is what makes it a real fix — no residual string list to rot.

Two details worth noting, neither blocking:

- The `action === RELAY_GATE_C_ACTION` early return at `:1044` is a nice catch. Without it the relay's own escalate would mark its own routing entry, which would be harmless today but confusing to debug later.
- My smaller note about `replySkillGateCBlocked` never being cleared still stands, but it's bounded now that I've looked properly: routing entries are per-task and deleted at the top of `handleAgentResponse`, so a stale flag can't outlive the turn. The `if (routing.replySkillGateCBlocked) continue;` guard also keeps the sweep from re-logging. Fine as is.

Resolving.

src/autonomy/action-log-repo.ts
josephfung: Verified. `findAnyPendingByTask` is removed entirely and `findPendingByTaskAndSkills` filters on `skill_name = ANY($2::text[])` (`action-log-repo.ts:200-215`). I checked the column against migration 031 — `skill_name TEXT NOT NULL` — so the predicate is on the real column, not a plausible-looking wrong one.

With the relay passing the reply-class set, a pending `calendar-update-event` on the same turn no longer gets joined, which was the concrete incident-shaped failure. Including `dispatcher-relay` itself in the dedupe set is right too — that's what collapses repeated relay escalates on one turn.

The secondary suggestion (recording the join on the existing row so approving it is *known* to cover both actions) wasn't taken. I don't think it needs to be now: with the predicate scoped to reply-class skills, a join means two reply-shaped requests for the same turn, where approving either delivers the reply. The ambiguity I was worried about came from joining across action types, and that's gone.

Resolving.

tests/integration/http-api.test.ts
josephfung: Verified, and it settles the question I asked rather than sidestepping it.

`publishRelayGateCPendingStatus` (`dispatcher.ts:1334-1359`) returns early unless `channelId === 'http'`, so email and Signal escalate silently as before and only the request/response surface gets a terminal event. The content is the fixed `RELAY_GATE_C_HTTP_PENDING_MESSAGE` constant, so this doesn't reintroduce an ungated content path — it's a status, not a relay. The trust-scoring test going back through `POST /api/messages` and asserting the status body restores the coverage the earlier version dropped.

Covering the skill-blocked-suppress and audit-fail-closed paths too was the right call. Those are the other two ways the waiter could have been left hanging, and it would have been easy to fix only the branch I named.

Resolving.

tests/unit/dispatch/dispatcher-reply-lock.test.ts
josephfung: Verified. `dispatcher-relay-gate-c.test.ts` now ends with `end-to-end: known-tier inbound through handleInbound escalates the relay`, which drives the real inbound path instead of hand-seeding the private `taskRouting` map. That's the composition coverage that was missing, and it's the test that would have caught the approve-path dead end as behaviour.

The suite also picked up the judge allow / judge escalate / no-judge branches and the HTTP waiter settlement, so the three-way decision and the channel-specific terminal event are each pinned rather than implied.

One thing that is still true and worth carrying forward rather than fixing here: Gate C interacting with the content-block rewrite retry has no test, even though `originator` now propagates through `RelayOutboundContext` specifically to keep that path gated. It's a narrow gap and the propagation is clearly deliberate, so I'm not going to hold the thread open for it, but it's the one relay interaction still covered only by reading the code.

Resolving.

CHANGELOG.md
josephfung: Verified. Seven spec files updated, and the two I named now say the right thing:

- `06-audit-and-security.md:48` — "Gate C (execution for skills, **and dispatch for the auto-reply relay** with `action: 'dispatcher-relay'`)", plus a note on the new optional `conversationId`.
- `09-contacts-and-identity.md:628` — "the execution layer emits for Gate C on skills, and the dispatch layer also emits Gate C for the auto-reply relay".

Spec 15 gained the relay gate alongside the existing reply-lock / no-reply exceptions, and spec 10 picks up the extraction row. The CHANGELOG line now reads "every auto-reply is tier-gated; approve re-delivers via `dispatcher-relay`", which describes the per-turn gate rather than only the post-skill suppression.

The one caveat: that changelog line and the spec text both describe the approve path as working, and per the thread on `dispatcher.ts` it doesn't yet in a deployed build. Not a docs problem — the docs will be accurate once the registry enrolment lands — but worth not forgetting the ordering if this ships in pieces.

Resolving the docs thread.

src/dispatch/relay-gate-c.ts
josephfung: The change is right, but it landed red — leaving this open until CI is green.

The substance is all there: `originator` is now required on `registerExternalTaskRouting` (so "did every caller pass it?" is a compile-time question), the secret-capture resume parses the round-tripped originator and falls back to a fail-closed unresolved external originator with a warn, and the skip reason is renamed `originator_absent`. That's exactly the shape I was arguing for, including the fallback being fail-closed rather than skip.

What's broken is a stale assertion in the file the change touches. `src/secrets/secret-capture-resume-subscriber.test.ts:97` still asserts:

```ts
expect(routingCalls[0]!.routing).toEqual({ channelId: 'email', conversationId: 'conv-1', senderId: 'ceo' });
```

`toEqual` is exact, and the routing object now carries `originator`, so CI fails with `expected { channelId: 'email', …(3) } to deeply equal { channelId: 'email', …(2) }` — 1 failed / 6683 passed on run 33906177731. The test is asserting the old shape of the thing this thread asked you to change, so it needs the originator added to the expectation (asserting the fail-closed unresolved originator specifically would be better than `expect.anything()`, since that fallback is the security-relevant half).

Worth flagging that the PR body still has the "CI passes" box checked.

josephfung:
## Re-review of `6b1ac70b`

Six of eight resolved. Two still open, one of which is the original blocker in a new form.

**Resolved:** judge on the relay path (with the reply-all context in the action description, which is what makes the verdict meaningful), `isGateCBlockError` removed in favour of `conversationId` on `authorization.decision`, dedupe scoped to reply-class skills via the real `skill_name` column, HTTP waiter settled with a channel-scoped constant, end-to-end known-tier test through `handleInbound`, and seven spec files corrected.

The judge fix in particular is the shape I was hoping for rather than the shortcut — telling the judge that the adapter may reply-all is the difference between a real third-party determination and a guess about `recipientId`.

**Still open:**

**1. `dispatcher-relay` is never registered, so `approve-action` still hits `Tool 'dispatcher-relay' not found in registry`.** The skill and its payload are correct, but `src/skills/loader.ts:197` only loads tools whose registry row is enabled, and `reconcileOne` only creates rows for names in `config/registry-defaults.yaml`. This branch touches nothing under `config/` or `src/db/migrations/`, and `dispatcher-relay` is not in that list. Fresh installs and the existing prod deployment are both affected. The unit tests construct the handler directly and the dispatcher tests stub `approvalTrigger`, so nothing exercises the registry lookup where the failure actually is. Same error string, same destroyed reply, one layer further out.

**2. CI is red.** `src/secrets/secret-capture-resume-subscriber.test.ts:97` uses `toEqual` against the pre-change routing shape, and the (correct) decision to require `originator` on `registerExternalTaskRouting` added a key to it. 1 failed / 6683 passed on run 33906177731. The "CI passes" box in the PR body is checked.

Detail on both in the open threads.

src/dispatch/dispatcher.ts
josephfung: Verified — the approve path now actually reaches the handler.

`dispatcher-relay` is in `config/registry-defaults.yaml:118` under `tools:`, so `reconcileOne` will insert an enabled row on any deployment that has no row for it yet, and `loadToolsFromDirectory` will register the handler. I traced it end to end and also confirmed `humanApproved: true` skips Gate C itself (`execution.ts:891` wraps the whole gate block), so an approved relay publishes rather than re-escalating into a second approval. That loop was the other way this could have gone wrong.

The new `tests/unit/skills/dispatcher-relay-registry.test.ts` is the right test. It reads the real `config/registry-defaults.yaml` and the real `skills/` directory rather than fixtures, so removing the entry breaks the build, and the second case goes through `ExecutionLayer.invoke` and asserts the `outbound.message` actually lands. That's the registry lookup the earlier tests couldn't see.

On my caveat about whether an infrastructure tool belongs in that list: `extract-relationships` already ships with `allowed_callers: ["system"]` and is enrolled the same way, so this follows an established pattern rather than inventing one. Withdrawn. I've left a separate note on the registry test about testing the caller gate's enforcement rather than just its manifest value.

Resolving.

src/dispatch/relay-gate-c.ts
josephfung: Verified — CI is green on `6ea8f3bf` (all 12 checks), and the assertion now includes the originator rather than being loosened to `expect.anything()`.

One observation about the shape it pins, which I've raised as a separate comment on line 102 rather than reopening this: the fixture omits `tier`, so the expectation is a `systemRole: 'principal'` originator with `tier: null`, which `decideRelayGateC` escalates. Production is fine — `build-capture-origin.ts:33` stores the full stamped originator bag, tier included — so this is about what the test demonstrates, not about the code.

Resolving this thread.

src/secrets/secret-capture-resume-subscriber.test.ts
```diff
@@ -94,7 +94,13 @@ describe('SecretCaptureResumeSubscriber', () => {
     expect(routingCalls).toHaveLength(1);
     const publishedTaskId = (published[0]!.event as AgentTaskEvent).id;
     expect(routingCalls[0]!.taskEventId).toBe(publishedTaskId);
-    expect(routingCalls[0]!.routing).toEqual({ channelId: 'email', conversationId: 'conv-1', senderId: 'ceo' });
+    expect(routingCalls[0]!.routing).toEqual({
+      channelId: 'email',
+      conversationId: 'conv-1',
+      senderId: 'ceo',
+      // Gate C on the relay requires originator (#1733) — round-tripped from the capture token.
+      originator: { ...ORIGINATOR, tier: null },
```
josephfung: **This pins the one originator shape that would put a CEO-facing reply behind an approval.**

The assertion is correct for the fixture, and the production path is fine, so this is about what the test demonstrates rather than a defect.

`ORIGINATOR` at line 32 is `{ contactId: 'ceo', systemRole: 'principal', channel: 'email', initiatedAt: 't' }` — no `tier`. `parseOriginator` normalizes that to `tier: null`, which is what this expectation encodes. Run that through the new gate: `getInitiatingTier` only skips for `systemRole` `'system'` or `'agent'`, so a `'principal'` originator falls through to `originator.tier ?? null` → `null`; `isExternalOriginatorMissingTier` then returns true, and `decideRelayGateC` escalates. So the shape this test pins is a CEO secret-capture resume whose reply is withheld pending approval — the exact thing #1733's fourth acceptance criterion says must not change ("no new friction on CEO-facing replies").

Real traffic doesn't hit that, because `build-capture-origin.ts:33` stores the whole stamped `taskMetadata.originator` bag and `stampOriginator` populates `tier`. But the fixture is now the only place in the suite that describes what a resume originator looks like, and it describes one that fails the criterion. Adding `tier: 'principal'` to the fixture would make it representative, and a companion assertion that `decideRelayGateC` allows it would turn the criterion into something the tests actually check rather than something we've each verified by reading `stampOriginator`.

One related thing worth knowing rather than fixing: capture tokens whose bag fails `parseOriginator`'s narrowing — the pre-#972 tokens its own comment mentions — fall back to `unresolvedExternalOriginator` with `tier: null`, so a CEO redeeming an in-flight capture link across this deploy gets their resume reply escalated. Narrow (one-time, short-lived links) and fail-closed is the right default, so I'd just note it rather than special-case it. If it ever does bite, resolving the contact from `senderId` instead of hardcoding `contactId: 'unresolved'` is the obvious out.

tests/unit/skills/dispatcher-relay-registry.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    const count = await loadToolsFromDirectory(
+      discoveries,
+      registry,
+      logger,
+      new Set(['dispatcher-relay']),
+    );
+    expect(count).toBe(1);
+    expect(registry.get('dispatcher-relay')).toBeDefined();
+    expect(registry.get('dispatcher-relay')!.manifest.allowed_callers).toEqual(['system']);
```
josephfung: **This asserts the manifest value, not the enforcement — worth one negative case given what this tool can do.**

The rest of the file is exactly the coverage that was missing, so this is a small addition rather than a criticism of the approach.

`dispatcher-relay` publishes an `outbound.message` with caller-supplied `channelId`, `to`, and `body` straight onto the bus, and `humanApproved: true` skips both the caller gate (`execution.ts:842`) and Gate C (`:891`). That makes it a send-anything-to-anyone primitive whose only guard is that nothing except `approve-action` can reach it. Checking `manifest.allowed_callers` equals `['system']` confirms the manifest was authored correctly; it doesn't confirm the gate refuses a caller. A manifest typo or a change to the caller-gate condition would leave this assertion green.

A second case — coordinator invoking `dispatcher-relay` *without* `humanApproved` and expecting a refusal — is a couple of lines and pins the property that actually matters. The enforcement is generic and tested elsewhere, so this is belt-and-braces, but the blast radius here is larger than for the other `allowed_callers` tools and the test file already has the registry wired up.

Not worth changing, but noting since I checked it: the `allowed_callers` filter in tool search defaults the caller to `'system'` when `options.agentId` is unset (`execution.ts:1453`), which would make a `["system"]` tool visible. I traced the agent path — `runtime.ts:1430-1432` always populates `agentId` in `invokeOptions` — so it isn't reachable from an agent turn, and the pattern predates this PR (`extract-relationships` has the same manifest). Fine as is.

josephfung (commented):
## Re-review of `6ea8f3bf`

Both remaining findings are fixed and CI is green across all 12 checks. All eight threads are now resolved.

The enrolment fix is complete rather than minimal: the entry in `config/registry-defaults.yaml:118` carries a comment explaining why an infrastructure tool is on that list, and the new test reads the real defaults file and the real `skills/` directory instead of fixtures, so it fails if the entry is ever dropped. I also checked the thing that could have made the approve path fail a third way — `humanApproved: true` skips Gate C entirely (`execution.ts:891` wraps the whole gate block), so an approved relay publishes instead of re-escalating into a fresh approval. No loop.

Nothing blocking left. Two advisory notes below, both about test coverage rather than behaviour — neither needs to hold the PR.

For whoever picks this up next: the deferred item worth remembering is that Gate C interacting with the content-block rewrite retry still has no test, even though `originator` propagates through `RelayOutboundContext` specifically to keep that path gated.

josephfung:
@Coderabbitai review

src/dispatch/dispatcher.ts
josephfung: Fixed in `6b1ac70b` + `f234bb3e`.

Added a real `skills/dispatcher-relay/` skill (`action_risk: medium`, `allowed_callers: ["system"]`) whose handler re-publishes the withheld `outbound.message` via `bus.publish('dispatch', …)`. Approval rows now store `buildRelayApprovalInput({ channelId, to, body, conversationId, accountId? })` so `approve-action`'s `invoke(row.toolName, row.payload, { humanApproved: true })` path actually delivers. Also added a VERB_RULES entry so CEO notification copy reads "Relay withheld reply".

Enrollment: `dispatcher-relay` is in `config/registry-defaults.yaml` under Autonomy (infrastructure-only — not pinned to any agent). Reconcile enrolls an enabled row on first boot when no row exists. `tests/unit/skills/dispatcher-relay-registry.test.ts` loads via `loadToolsFromDirectory` and asserts `ExecutionLayer.invoke(..., { humanApproved: true })` publishes `outbound.message`.

src/dispatch/relay-gate-c.ts
josephfung: Fixed in `6b1ac70b` + `6e3c770c`.

`registerExternalTaskRouting` now requires `originator` (compile-time). Secret-capture resume parses the round-tripped originator onto the routing entry; if absent, registers a fail-closed unresolved external originator and warns. Skip reason renamed to `originator_absent`. The resume routing spy assertion now expects `originator` rather than being loosened.

src/dispatch/relay-gate-c.ts
josephfung: Fixed in `6b1ac70b`.

Removed `isGateCBlockError` entirely. `authorization.decision` now carries optional `conversationId` (execution `publishGateCDecision` + relay publish). `handleAuthorizationDecision` marks `replySkillGateCBlocked` by `taskEventId` and/or `conversationId`, covering delegated specialists without prose matching.

CHANGELOG.md
josephfung: Fixed in `6b1ac70b`.

Updated `docs/specs/06-audit-and-security.md`, `09-contacts-and-identity.md`, and `15-outbound-safety.md` for dispatch-layer Gate C / `dispatcher-relay`. CHANGELOG now describes the full per-turn relay gate + fulfillable approve path.

tests/unit/dispatch/dispatcher-reply-lock.test.ts
josephfung: Fixed in `6b1ac70b`.

Added an end-to-end unit that drives a known-tier contact through real `handleInbound` → `agent.response` (not hand-seeded `taskRouting`), asserting escalate + approval payload. Also covered judge allow/escalate branches and HTTP pending-status settlement.

tests/integration/http-api.test.ts
josephfung: Fixed in `6b1ac70b`.

On HTTP escalate (and skill-blocked suppress / audit fail-closed), the dispatcher publishes a terminal `outbound.message` with `RELAY_GATE_C_HTTP_PENDING_MESSAGE` so `EventRouter.waitForResponse` settles immediately instead of hanging 120s. Email/Signal escalate stays silent. Restored the trust-scoring test to go through `POST /api/messages` and assert that status body.

src/autonomy/action-log-repo.ts
josephfung: Fixed in `6b1ac70b`.

Replaced `findAnyPendingByTask` with `findPendingByTaskAndSkills(taskId, skillNames)` (`skill_name = ANY(...)`). The relay passes `dedupePendingSkillsOnTask: [...REPLY_SKILLS_GATE_C]` (includes `dispatcher-relay` + reply skills), so a pending `calendar-update-event` on the same turn is no longer joined. Later: expired rows are excluded via `expires_at > now()` (`1a26197a`).

src/dispatch/relay-gate-c.ts
josephfung: Fixed in `6b1ac70b`.

`decideRelayGateC` is now async and mirrors `ExecutionLayer.resolveTierGateDecision`: when reply-to-sender vs third-party disagree (known × reversible-external), it consults `EscalationJudge` with a description that notes the email adapter may reply-all. Fail closed if the judge is absent/disabled/undetermined. Did **not** hardcode `isThirdPartyFacing: false`. `escalationJudge` is wired through `DispatcherConfig` / `src/index.ts`.

src/secrets/secret-capture-resume-subscriber.test.ts
josephfung: Fixed in `771b67b3` (+ `328a99b6` for absent-tier preservation).

Updated the fixture to `tier: 'principal'` so it matches what `stampOriginator` / `build-capture-origin` actually store. The routing assertion expects that shape, and a companion test runs `decideRelayGateC` on the registered originator and asserts `allow` / `tier_permits_external_send`.

Separately, `parseOriginator` no longer coerces a missing `tier` to `null` — absent stays absent so round-trips match fixtures without tier (`specialist-resume-flow`).

tests/unit/skills/dispatcher-relay-registry.test.ts
josephfung: Agreed — added in `771b67b3` (`rejects non-system callers without humanApproved`): coordinator-shaped `agentId` invoke without `humanApproved` expects refusal. The happy path still covers `humanApproved: true` → `outbound.message`.

skills/dispatcher-relay/handler.ts
```diff
@@ -0,0 +1,76 @@
...
+      return { success: false, error: 'Missing required input: to (string)' };
+    }
+    if (!body || typeof body !== 'string') {
+      return { success: false, error: 'Missing required input: body (string)' };
+    }
+    if (!conversationId || typeof conversationId !== 'string') {
+      return { success: false, error: 'Missing required input: conversationId (string)' };
+    }
+    if (body.length > MAX_BODY_LENGTH) {
+      return { success: false, error: `body must be ${MAX_BODY_LENGTH} characters or fewer` };
```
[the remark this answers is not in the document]
josephfung: Fixed in `1a26197a`.

Shared `RELAY_BODY_MAX_LENGTH` (50_000) between `relay-gate-c` and the handler. `requestRelayGateCApproval` now refuses to create a pending row when the body exceeds that limit (withhold + warn), so approve cannot create an un-fulfillable approval. Handler check retained. Tests cover max and max+1 on both the dispatcher path and the handler.

src/autonomy/action-log-repo.ts
```diff
@@ -192,6 +192,29 @@ export class ActionLogRepo {
...
+    taskId: string,
+    skillNames: readonly string[],
+  ): Promise<ActionLogRow | null> {
+    if (skillNames.length === 0) return null;
+    const result = await this.pool.query(
+      `SELECT * FROM autonomy_action_log
+       WHERE task_id = $1
+         AND outcome = 'pending_approval'
+         AND skill_name = ANY($2::text[])
+       ORDER BY created_at ASC
```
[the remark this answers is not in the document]
josephfung: Fixed in `1a26197a`.

`findPendingByTaskAndSkills` (and `findPendingByTaskAndSkill`, used as the second dedupe check) now require `expires_at > now()`, matching the other pending-approval queries in this repo.

src/secrets/secret-capture-resume-subscriber.ts
```diff
@@ -180,7 +239,20 @@ export class SecretCaptureResumeSubscriber {
     });
 
     try {
-      this.registerRouting?.(task.id, { channelId, conversationId, senderId });
+      const routingOriginator =
+        parseOriginator(originator) ?? unresolvedExternalOriginator(channelId);
```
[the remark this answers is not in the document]
josephfung: Fixed in `1a26197a`.

`routingOriginator` is computed once before `createAgentTask` and written to both `metadata.originator` and dispatcher routing. Absent/malformed bags use `unresolvedExternalOriginator`. Tests assert that those cases escalate under `decideRelayGateC`.

tests/integration/conversation-memory.test.ts
```diff
@@ -41,7 +42,12 @@ describe('Multi-turn conversation with working memory', () => {
     });
     coordinator.register();
 
-    const dispatcher = new Dispatcher({ bus, logger });
+    // Principal resolver so Gate C on the relay (#1733) allows CLI-style delivery.
+    const dispatcher = new Dispatcher({
+      bus,
+      logger,
+      contactResolver: makePrincipalContactResolver(),
+    });
```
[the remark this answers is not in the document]
josephfung: Skipping for this PR.

`makePrincipalContactResolver` was an intentional, scoped harness change so Gate C allows CLI/HTTP happy-path delivery without rewriting those suites into full contact seeding. My earlier review accepted that posture for the principal surfaces and asked for composition coverage via the known-tier `handleInbound` unit test (which we added) rather than converting every integration fixture.

Rewiring `conversation-memory` / `vertical-slice` onto Docker `curia_test` + real `ContactResolver` with seed/cleanup is a larger fixture redesign outside #1733's blast radius. Happy to track it separately if desired.

josephfung:
@Coderabbitai approve

PR #1738, opened 2026-09-05
fix(email): detect auto-generated mail from the message, not contact kind

### LINKED ISSUE #1734: Detect auto-generated mail from the message, not the sender's contact kind

josephfung:
## Summary

Closes #1734. Curia was deciding whether mail is machine-generated from the sender contact's `kind`, so Exchange calendar RSVP declines from a `kind=person` contact were treated as ordinary correspondence and answered individually.

This PR classifies auto-generated mail from **message** signals (RFC 3834 `Auto-Submitted`, `Precedence`, explicit calendar `method=REPLY`/`CANCEL`, OOO subject, etc.), surfaces it on inbound metadata, and suppresses replies at the dispatch layer for live inbound (coordinator) — not only ceo-inbox sweeps.

## Review follow-ups (addressed)

**Round 1**
- Dropped dispatch-level body-keyword carve-outs (marketing "expires" / phishing vocabulary no longer select escalate)
- Always create `agent.task` for non-principal auto-gen; agent judges escalation; relay stays suppressed
- Principal-originated forwards that look auto-generated bypass suppress
- Ambiguous calendar (no MIME `method=`) no longer defaults to `calendar_reply`
- `outbound.no_reply` / `auto_generated` now always references a real `agent.task` + real `agentId`
- Added negative tests for marketing bulk+"expires", phishing security vocabulary, method-less `.ics`, and principal OOO forwards

**Round 2** (`ac405d0c`)
- Coordinator prompt aligned with preamble (no stale "carve-outs" wording; marketing urgency + verified-sender guidance)
- `emailMeta` parsed once and reused after `stampOriginator`
- Observability: `classifyNoReply` runs before auto-gen suppress — preamble compliance → `agent_declined`; narrated reply → `auto_generated` + `abandonedContent`
- Cost note acknowledged: correctness over token savings; signal-level narrowing (`precedence_bulk`) left as a follow-up if audit volume is high

**Round 3 — CodeRabbit** (`bbefdd6e`)
- CHANGELOG Unreleased entry shortened to ≤15 words after the em dash
- Calendar `method=` honoured only with calendar MIME / `isCalendarInvite` / `.ics` evidence (PDF `method=REPLY` no longer classifies)
- Defense in depth: task metadata stamps `autoGeneratedSuppress` + `autoGeneratedSenderId`; `email-reply` / `email-send` refuse that recipient while principal escalation to other addresses stays open

## Changes

- `classifyAutoGeneratedMail` in `src/channels/email/auto-generated-mail.ts` (consumes `isCalendarInvite` + explicit `method=` on calendar MIME only)
- Stamp `isAutoGenerated` / signals on `convertNylasMessage` metadata
- Dispatch: silence preamble + relay suppress for non-principal auto-gen; principal bypass; preamble vs backstop audit split; skill-path sender block via task metadata
- Update coordinator + ceo-inbox prompts; extend specs and CHANGELOG

## Related Issues

Closes #1734

## Checklist

- [x] CI passes

<div><a href="https://cursor.com/agents/bc-01a072ef-df7f-7553-9461-4c3814d5606f?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-01a072ef-df7f-7553-9461-4c3814d5606f&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


src/channels/email/auto-generated-mail.ts
```diff
@@ -0,0 +1,231 @@
...
+const BOUNCE_RE =
+  /\b(?:undeliverable|delivery status notification|mail delivery(?:\s+subsystem)?\s+failed|returned mail|delivery failure|permanent failure|mailbox (?:full|unavailable)|user unknown|delivery has failed|could not be delivered)\b/i;
+
+const PAYMENT_RE =
+  /\b(?:payment (?:failed|failure|declined)|failed payment|invoice overdue|past due|billing (?:issue|problem|failed)|card (?:declined|expired)|payment was (?:not|un)successful)\b/i;
+
+const SECURITY_RE =
+  /\b(?:security alert|suspicious (?:login|sign[- ]?in|activity)|unauthorized access|password reset|account (?:locked|suspended|compromised|disabled)|fraud alert|verify your (?:identity|account)|unusual sign[- ]?in)\b/i;
+
+const DEADLINE_RE =
```
josephfung: **This routes ordinary bulk mail into CEO escalation — the inverse of what the issue asked for.**

`DEADLINE_RE` matches bare `\bexpires?\b` (the `(?:\s+(?:today|tomorrow|on\b|in\s+\d+))?` group is optional), plus `deadline`, `final notice`, and `last chance`. `collectCarveOuts` runs it over `subject + bodyText + snippet`.

I pulled the four regexes out and ran them against realistic subject lines. Every one matched:

| subject | carve-out |
|---|---|
| Your 20% off code expires Sunday | `hard_deadline` |
| Your trial expires in 3 days | `hard_deadline` |
| Early bird pricing deadline is Friday | `hard_deadline` |
| Last chance to save on spring styles | `hard_deadline` |
| Final notice: your profile is incomplete | `hard_deadline` |
| Your subscription expires soon — renew now | `hard_deadline` |

Now chain it. Marketing and list mail sets `Precedence: bulk` or `Precedence: list` almost universally, which `collectHeaderSignals` turns into `precedence_bulk` → `isAutoGenerated: true`. Then any of the above sets `actionable: true`. Then `dispatcher.ts:891` takes the actionable branch: the coordinator runs, with a preamble that says *"Escalate the actionable signal to the principal via their preferred channel or the bullpen."*

So the highest-volume category of auto-generated mail — the one this feature exists to quiet — becomes the category most likely to reach the CEO on Signal. Before this PR that mail got an unwanted reply; after it, it gets an unwanted interruption. The failure moved rather than closed.

The contrast with your own ceo-inbox prompt is the tell. §4e-pre lists "Hard deadline or expiry embedded in the content" as a carve-out and hands it to the LLM with the whole message in view, which is the right instrument for a judgment like that. At dispatch the same decision is a substring match, and the dispatch one is the one that decides whether the coordinator runs at all.

Two directions that both seem defensible:

- **Don't decide actionability at dispatch.** Suppress the *reply* on every auto-generated message (that part the signals genuinely support), and let the agent judge whether anything warrants escalation. Costs a turn per auto-gen message, which is real but bounded and buys you the nuance.
- **Keep a dispatch-level carve-out but restrict it to signals that are unambiguous in a header context.** `Auto-Submitted: auto-replied` plus `Content-Type: multipart/report; report-type=delivery-status` is a genuine DSN. "Expires" in a body is not evidence of anything. `BOUNCE_RE` is the one of the four that mostly survives this test, because its vocabulary is genuinely machine-emitted; the other three are ordinary English.

If the regexes stay in any form, `hard_deadline` is the one I'd drop outright — it has no unambiguous form.

src/channels/email/auto-generated-mail.ts
```diff
@@ -0,0 +1,231 @@
...
+  return OOO_SUBJECT.test(subject.trim()) ? ['ooo_subject'] : [];
+}
+
+const BOUNCE_RE =
+  /\b(?:undeliverable|delivery status notification|mail delivery(?:\s+subsystem)?\s+failed|returned mail|delivery failure|permanent failure|mailbox (?:full|unavailable)|user unknown|delivery has failed|could not be delivered)\b/i;
+
+const PAYMENT_RE =
+  /\b(?:payment (?:failed|failure|declined)|failed payment|invoice overdue|past due|billing (?:issue|problem|failed)|card (?:declined|expired)|payment was (?:not|un)successful)\b/i;
+
+const SECURITY_RE =
```
josephfung: **The security carve-out is written in phishing vocabulary, so an untrusted sender can select the escalate path.**

`SECURITY_RE` matches `security alert`, `verify your (identity|account)`, `password reset`, `unusual sign-in`, `account (locked|suspended|compromised)`. That is close to a word list for the genre of mail most likely to be hostile. I checked both directions and they behave identically:

```
Security alert: unusual sign-in on your account. Verify your identity.  → security_alert
Password reset requested for your account                              → security_alert
```

The consequence is a content-controlled routing primitive. Any external sender who wants the coordinator to run on their message, with a preamble instructing it to escalate to the principal, includes those words. The injection scanner runs first (`dispatcher.ts:719`) but only produces a trust score — it doesn't gate this branch, and none of its patterns cover this.

I want to be fair about the ceiling: the coordinator still applies judgment, Gate C still governs anything outbound, and the escalation reaches the CEO as a notification rather than an action. So this is "attacker can reliably get a message in front of the CEO with an urgent framing", not "attacker can make Curia do something". But that's the exact primitive a phisher wants, and the carve-out hands it over on a substring.

If a dispatch-level security carve-out is worth keeping at all, `senderVerified` is already computed in the converter from `Authentication-Results` and would be the natural gate — a genuine security alert from a real provider passes SPF/DKIM/DMARC; a spoofed one usually doesn't. Requiring `senderVerified` before honouring `security_alert` costs almost nothing and removes the easy version of this.

src/dispatch/dispatcher.ts
```diff
@@ -863,6 +870,53 @@ export class Dispatcher {
 
         taskContent = buildCcPreamble(emailMeta, payload.accountId, nylasMessageId) + taskContent;
       }
+
+      // Auto-generated mail (#1734): message-derived, independent of contact kind.
+      // Non-actionable → suppress at dispatch (no agent.task, no outbound reply).
+      // Actionable carve-out → still run coordinator with a silence+escalate preamble;
+      // the relay is still suppressed so a narrated closing cannot email the sender.
+      if (emailMeta.isAutoGenerated) {
```
josephfung: **No principal carve-out — a forwarded auto-reply from the CEO is silently dropped.**

This block runs before `stampOriginator`, so there is no tier or principal check anywhere on the suppression path. Any message that trips a signal returns early with no `agent.task`, no reply, and nothing but an `outbound.no_reply` row.

`OOO_SUBJECT` explicitly strips `re:` / `fw:` / `fwd:` prefixes, so forwarded auto-replies match. I checked:

```
Fwd: Automatic reply: Out of office     → true
Fw: Out of Office: Jane Doe             → true
Re: Fwd: Automatic reply: I am away     → true
```

So: Joseph gets an OOO from someone he needs, forwards it to Curia with "can you try reaching them another way", and Curia never sees it. That is an ordinary EA request, and the failure is completely invisible from his side — no error, no bounce, no delay, just nothing. The same is true of forwarding a bounce notice to ask Curia to find another address, which is arguably the single most likely thing a person forwards to an assistant.

This is the shape we already worked through on #1732, where the fix was a principal carve-out plus a notify. The argument carries over unchanged, and the ingredients are already here: `stampOriginator` sits about fifty lines below and yields `isPrincipalOriginated`, and moving this block after it (or hoisting the originator stamp above it) is mechanical.

What to do when the principal *does* send auto-generated-looking mail is the interesting half. Dispatching normally seems right to me — if the CEO forwards you a bounce, the bounce is the subject of the request, not noise to filter. The signals are still on the metadata, so the agent can see what it is without dispatch having decided for it.

src/channels/email/auto-generated-mail.ts
```diff
@@ -0,0 +1,231 @@
...
+    } else if (method === 'CANCEL') {
+      signals.push('calendar_cancel');
+    } else if (bodyEmpty && (method === undefined || method === 'REQUEST')) {
+      // Exchange decline notices often ship empty body + invite.ics with
+      // method=REPLY already caught above; when method is absent but the body
+      // is empty, treat as machine RSVP/notice rather than a formal REQUEST.
+      // Explicit method=REQUEST with a non-empty body is a real invite — not
+      // auto-generated here (bodyEmpty is false).
+      if (method === undefined) {
+        signals.push('calendar_reply');
```
josephfung: **The ambiguous calendar case defaults to suppression, which drops real invitations.**

The branch reads: attachment looks like a calendar item, body is empty, no `method=` in the Content-Type → label it `calendar_reply` → `isAutoGenerated` with no carve-out → `dispatcher.ts:880` returns early, no `agent.task` at all.

"Looks like a calendar item" is broad. `isInvite` above is satisfied by `isCalendarInvite`, which `nylas-client.ts:619` sets for any `text/calendar`, `application/ics`, `text/x-vcalendar`, or `*.ics` filename — and independently by the `.ics` filename check here. Plenty of senders attach an `.ics` with no MIME `method` parameter (Calendly, Eventbrite, ticketing systems, anything hand-rolling an ICS), and invitations with empty bodies are entirely normal — an Outlook invite with no note in it has exactly this shape.

So a genuine meeting invitation gets silently dropped. `auto-generated-mail.test.ts:51` pins this as intended, which is what makes me want to raise it rather than assume it's an oversight: the case is understood, and the tie-break went toward suppression.

I'd flip the tie-break. The two errors aren't symmetric — guessing "reply" on a real invite loses a meeting with no trace; guessing "request" on a real RSVP costs one agent turn that ends in `NO_REPLY` anyway now that #1732 has landed. The comment above the branch reasons carefully about which is more *likely*, but likelihood isn't the right axis when one outcome is silent and irreversible.

Better still, the ambiguity is resolvable rather than a coin flip: `METHOD` is a required property of the iCalendar object for any scheduling message, so it's in the `.ics` body even when the MIME parameter is missing. That needs the attachment fetched, which this pure module can't do — but `ceo-inbox-download-attachment` already exists, and the classifier could return "calendar, method unknown" and let a layer that can fetch decide, instead of guessing here.

src/bus/events.ts
```diff
@@ -248,7 +248,12 @@ interface OutboundSuppressedDuplicatePayload {
 // on this turn, whereas the reply-lock fires after a human-facing skill already succeeded.
 // See #1732.
 interface OutboundNoReplyPayload {
-  routingTaskId: string;   // agent.task event ID whose routing entry was skipped
+  /**
+   * agent.task event ID whose routing entry was skipped — or, when
+   * `reason === 'auto_generated'` and no task was created, the inbound.message
+   * event id (#1734).
+   */
+  routingTaskId: string;
```
josephfung: **The field's meaning is now conditional on `reason`, and `agentId` records an agent that never ran.**

Updating the JSDoc honestly is better than leaving it stale, so this is a design comment rather than a correction. But `routingTaskId` now holds either an `agent.task` id or an `inbound.message` id depending on a sibling field, and nothing in the type says so. An audit query that joins `routingTaskId` → `agent.task` — the obvious thing to write given the name — silently returns nothing for every `auto_generated` row rather than failing.

The `agentId: 'coordinator'` at `dispatcher.ts:895` bothers me more. No agent ran; the whole point of this branch is that dispatch decided before any agent was involved. The codebase has an explicit position on this — `approve-action`'s comment about not publishing "a fake audit row with placeholder IDs, as a counterfeit audit trail is worse than a missing one" (ADR-017). A row saying the coordinator declined to reply, when the coordinator was never invoked, is that. It also makes the audit trail actively misleading for exactly the incident-reconstruction work that produced this issue: someone querying "what did the coordinator do that morning" gets turns it never took.

Two shapes that would say what happened:

- Make `agentId` optional on the payload and omit it here, with `routingTaskId` likewise optional — absent fields read as "no agent, no task", which is true.
- Or a distinct event (`inbound.suppressed`, say), since this really is a different thing: the other four reasons all describe an agent turn that produced no send, and this one describes a message that never became a turn.

The second is more work but is the honest model. Either way the extraction row in spec 10 will want a look — `suppress` on a conversation still fits, but `initiator` shouldn't be an agent.

tests/unit/channels/email/auto-generated-mail.test.ts
```diff
@@ -0,0 +1,192 @@
...
+      bodyText: 'Your license expires tomorrow.',
+      snippet: 'Your license expires tomorrow',
+      headers: [{ name: 'Precedence', value: 'bulk' }],
+      attachments: [],
+    });
+    expect(deadline.isAutoGenerated).toBe(true);
+    expect(deadline.carveOuts).toContain('hard_deadline');
+  });
+
+  it('does not flag a normal human email from the same contact', () => {
```
josephfung: **Every carve-out test asserts a true positive; nothing pins a false one.**

The suite is well-organised and the classification tests are genuinely good — `ignores Auto-Submitted: no` and the `method=REQUEST` with a body case are the sort of thing that's easy to skip. My concern is what isn't here.

This test is the only negative case, and it's the weakest possible one: a message with no signals at all, so `isAutoGenerated` is false and the carve-outs never come into play. The carve-out tests all feed text that genuinely is a bounce, a payment failure, a security alert, or a deadline, so all four regexes are only ever exercised on inputs they should match. Nothing establishes what they *shouldn't*.

Three cases that would put the contentious behaviour under test rather than under discussion:

1. `Precedence: bulk` newsletter whose body says "your trial expires in 3 days" → currently `actionable: true`, which I don't think is intended. Writing this test forces the question.
2. A principal-originated message with an OOO subject → currently suppressed with no `agent.task` (see the `dispatcher.ts` thread).
3. An `.ics` attachment with an empty body and no `method=` → currently suppressed as `calendar_reply`; `:51` pins the current behaviour, so a companion asserting a *real* invite of that shape survives would show the tension.

The dispatcher suite has the same shape — `dispatches a normal human email from the same contact` is the only negative, and it's a message with no signals. Neither suite currently has a test that would fail if the carve-out regexes were far too broad, which is why the breadth was easy to miss.

josephfung (commented):
## Review (draft): message-derived auto-generated detection

The core reframe is right and it lands the part of #1734 that mattered: classification is genuinely derived from the message, it finally consumes `isCalendarInvite` (computed in `nylas-client.ts` since it was added and read by literally nothing until now), and it runs at dispatch so it applies to live inbound instead of only ceo-inbox sweeps. That was the structural half of F4 and it's done properly.

A few things I checked that could have undermined it, and don't:

- **Headers are actually available.** `email-adapter.ts:398` already requests `fields: 'include_headers'` for `senderVerified`, so the RFC 3834 signals are live rather than aspirational. Worth knowing that the whole header half of this feature silently depends on that flag staying there.
- **Absent headers fail toward dispatch**, not toward suppression. That's the opposite of `parseSenderVerified`'s fail-closed stance and it's correct here — a missed classification costs one unnecessary agent turn, a false one loses a message.
- **Ordering is safe.** Suppression sits after contact resolution (`:399`), rate limiting (`:691`), and the injection scan (`:719`), so a suppressed message still gets scanned and scored.
- Keeping the carve-out judgment in ceo-inbox as LLM-judged prose is the right call for that agent.

My concerns are all about the **carve-out layer** rather than the classification layer. Classification asks "is this machine-generated", which message signals genuinely answer. Actionability asks "does this matter enough to interrupt the CEO", which they don't — and that's the decision that got implemented as four keyword regexes over the full body.

### Blocking-ish

1. **The carve-out regexes route ordinary bulk mail into CEO escalation** (`auto-generated-mail.ts:157`). I ran them; 10/10 realistic marketing subjects matched. Detail inline.
2. **`security_alert` lets an untrusted sender pick the escalate path** (`auto-generated-mail.ts:154`). Phishing vocabulary is the carve-out vocabulary.
3. **No principal carve-out** (`dispatcher.ts:878`) — a forwarded auto-reply from the CEO is silently dropped with no `agent.task`. Same shape as the #1732 finding, and the signal is computed one block later.
4. **Ambiguous calendar attachments default to suppression** (`auto-generated-mail.ts:133`), which drops real invites.

### Smaller

5. `routingTaskId` sometimes holds an inbound id and `agentId` is fabricated (`events.ts:256`).
6. Carve-out tests are all positive cases (`auto-generated-mail.test.ts:147`).

Since this is still draft: 1 and 4 are the two I'd want resolved before it leaves draft, because both are silent-loss or noise-amplification paths that are hard to notice in production. 3 is cheap. 2 and 5 are judgement calls.

src/channels/email/auto-generated-mail.ts
josephfung: Verified — the regexes are gone (`BOUNCE_RE`/`PAYMENT_RE`/`SECURITY_RE`/`DEADLINE_RE` no longer exist), `actionable` and `carveOuts` are off the classification type, and `AutoGeneratedCarveOut` is deleted. The only surviving references anywhere are the two `not.toHaveProperty('carveOuts')` assertions, which is a good way to pin the removal.

The part I like is where the knowledge went rather than that it was deleted. `buildAutoGeneratedPreamble` now carries the nuance the regexes couldn't: it names the actionable categories *and* explicitly says "marketing 'expires'/'last chance' copy is not actionable" and "security alert **from a verified sender**". That's the same judgment, given to the instrument that can actually make it, with the specific false positives from this thread written into the guidance. Better than either the regex version or a bare "use your judgment".

`does not escalate marketing bulk mail just because the body says expires` locks in the case I raised.

Resolving. One consequence worth a decision rather than a change — noted in a fresh comment: every auto-generated message now costs a coordinator turn.

src/channels/email/auto-generated-mail.ts
josephfung: Verified — `SECURITY_RE` is gone, so phishing vocabulary no longer selects any path at dispatch.

Your framing is right: a sender who only wants a coordinator turn already had one before this feature existed, so removing the carve-out returns that surface to its baseline rather than leaving it widened. The `does not invent a security carve-out from phishing vocabulary alone` test pins it.

Worth noting the `senderVerified` idea survived in a better place — the preamble now says "security alert **from a verified sender**", so the agent gets the qualifier as judgment input rather than dispatch enforcing it on a substring. That's the right home for it.

Resolving.

src/dispatch/dispatcher.ts
josephfung: Verified. `stampOriginator` now runs above the auto-gen block and the guard is `!liveTurn`, and `does not suppress principal-originated forwards that look auto-generated` asserts the `Fwd: Out of Office` case gets a normal task with no preamble and no `outbound.no_reply`.

I checked that `liveTurn` is the right discriminator rather than a convenient one: `stamp-originator.ts` derives it as `originator.systemRole === 'principal'`, sourced from the resolved contact's `systemRole` — the structural binding, not the unreliable `role` field. So this uses the same signal as the elevated-skill gate, which is the correct precedent.

One inherited caveat, not a defect in this PR: the carve-out is only as good as contact resolution. If the principal mails from an address bound to a lookalike contact rather than the principal contact row, `systemRole` isn't `'principal'`, `liveTurn` is false, and the forward is suppressed silently. That's the #1514 identity-binding failure mode showing up in a new place rather than something introduced here — worth knowing about, not worth guarding twice.

Resolving.

src/channels/email/auto-generated-mail.ts
josephfung: Verified — `collectCalendarSignals` now only fires on an explicit `method=REPLY` or `method=CANCEL`, and `does NOT treat empty-body calendar attach without method as auto-generated` pins the flip.

This keeps the incident case covered, which was the thing I wanted to confirm before agreeing to narrow it: the 2026-09-04 Exchange declines carried an explicit `method=REPLY`, so they still classify. What changed is only the ambiguous shape, and that now fails toward a wasted turn instead of a lost meeting.

Agreed on deferring the ICS-body `METHOD` read. It needs a layer that can fetch the attachment, and the pure classifier is the wrong place for it. The residual gap is narrow and now fails safe: an RSVP whose Content-Type omits `method=` gets a normal turn, and #1732 gives the agent `NO_REPLY` as the out. Worth a tracked follow-up rather than scope here — happy to leave it as your call whether that gets filed now or waits for evidence it happens.

Resolving.

src/bus/events.ts
josephfung: Verified, and this is the better outcome than either fix I suggested. Both of my proposals worked around a fabricated row; removing the short-circuit means there is no fabricated row to model.

`routingTaskId` is back to an unconditional "agent.task event ID whose routing entry was skipped", `agentId` is documented as "Agent that produced the response whose delivery was suppressed", and `auto_generated` now sits alongside the other four reasons as a genuine agent-turn outcome. No conditional field meaning, no counterfeit coordinator row, and spec 10's extraction contract holds unchanged.

Resolving. One observability consequence of the new ordering is worth a look — separate comment.

tests/unit/channels/email/auto-generated-mail.test.ts
josephfung: Verified — all three negative cases landed, plus the principal one:

- `does not escalate marketing bulk mail just because the body says expires`
- `does not invent a security carve-out from phishing vocabulary alone`
- `does NOT treat empty-body calendar attach without method as auto-generated`
- `does not suppress principal-originated forwards that look auto-generated` (dispatcher suite)

The `not.toHaveProperty('actionable' | 'carveOuts')` assertions are a nice addition I didn't ask for — they'd fail if the carve-out concept came back through a different door, which is the kind of thing that otherwise creeps back in six months later.

The suite now has real negative coverage on every branch that was contentious. Resolving.

agents/coordinator.yaml
```diff
@@ -73,6 +73,11 @@ system_prompt: |
     message was not meant for this recipient), return exactly `NO_REPLY` as your
     entire response and nothing else. The system will not send. Do not narrate
     that decision in the response — narration becomes the outbound message.
+    Live inbound email may also arrive with an `[AUTO-GENERATED MAIL …]` preamble
+    (message-derived — independent of the sender's contact kind). Treat that as
+    authoritative: do not reply to the sender; escalate actionable carve-outs
```
josephfung: **Stale: this still describes the carve-out mechanism that was just removed.**

The prompt tells the coordinator to "escalate actionable carve-outs (bounce, payment failure, security alert, hard deadline)", but the preamble it actually receives no longer contains a carve-out list — `buildAutoGeneratedPreamble` now takes only `signals`, and dispatch stopped classifying actionability at all. So the agent is told to look for a field that isn't there.

More usefully, it's missing what the preamble *gained*. The new text carries "marketing 'expires'/'last chance' copy is not actionable" and "security alert from a verified sender" — the two qualifiers that exist precisely because the regex version got them wrong. The system prompt is where an agent forms its default posture, so having the anti-false-positive guidance only in the per-message preamble leaves it doing more work than it needs to.

Suggest matching the two: drop "carve-outs" as a term (it no longer names anything in the system), and mirror the preamble's qualifiers. Something like *"escalate to the principal only when the content is genuinely actionable — a real bounce, payment failure, a security alert from a verified sender, or a hard deadline. Marketing urgency copy is not actionable."*

No version bump needed for this — `0.18.5` already covers the branch.

src/dispatch/dispatcher.ts
```diff
@@ -888,6 +894,31 @@ export class Dispatcher {
...
 
+    // Auto-generated mail (#1734): message-derived, independent of contact kind.
+    // Always create agent.task so the coordinator can escalate when warranted;
+    // suppress the relay so a narrated closing cannot email the machine sender.
+    // Keyword body carve-outs do NOT select a privileged path at dispatch — that
+    // was routing marketing/"expires" and phishing vocabulary into CEO escalate.
+    // Principal-originated mail (forwards included) bypasses suppress entirely.
+    if (
+      payload.channelId === 'email' &&
+      parseEmailMetadata(payload.metadata).isAutoGenerated &&
```
josephfung: **`parseEmailMetadata` now runs three times per inbound email, and the "parse once" comment is no longer true.**

Line 825 parses it inside the email block, then this condition parses it again, then line 908 parses it a third time inside the branch it just guarded. The comment at line 820 — "Email metadata: parse once here and pass to the preamble builders below, replacing the repeated `payload.metadata as Record<string, unknown>` casts" — was written to prevent exactly this, and now sits directly above code that does it.

It's cheap (object property reads and two `Array.isArray` filters), so this is tidiness rather than performance. But the comment is the kind that quietly becomes a lie and then stops being trusted.

The reason it happened is real: `emailMeta` is scoped inside the `if (payload.channelId === 'email')` block at 824, and the auto-gen check had to move below `stampOriginator` to get `liveTurn`, which put it outside that scope. Hoisting the `const emailMeta` declaration above the channel check (or to a `let` assigned inside it) would let all three sites share one parse and keep the comment honest.

src/dispatch/dispatcher.ts
```diff
@@ -1116,6 +1148,33 @@ export class Dispatcher {
       return;
     }
 
+    // Auto-generated inbound (#1734): never relay a reply to the sender. The agent
+    // may still escalate via bullpen/Signal; only the dispatcher relay is suppressed.
+    if (routing.autoGeneratedSuppress) {
```
josephfung: **This makes it impossible to tell whether the preamble is working.**

The design now rests entirely on a prompt: the agent is told to return `NO_REPLY` for auto-generated mail, and the relay suppression is the backstop for when it doesn't. That's the right shape. But this branch returns before `classifyNoReply` runs, so every auto-generated turn records `reason: 'auto_generated'` regardless of what the agent actually produced. An agent that correctly returned `NO_REPLY` and an agent that narrated a full reply which the dispatcher then had to catch are indistinguishable in the audit log.

That's the one metric worth having here. "How often does the coordinator ignore the preamble" tells you whether the prompt is carrying its weight or whether the backstop is doing all the work — and if the answer is "the backstop catches most of them", the preamble wording needs another pass. Right now that question can't be answered from the audit log at all, and it's the question I'd expect to ask a month after this ships.

Two cheap ways to keep it, either fine:

- Run `classifyNoReply` first and record the more specific outcome — `agent_declined` when the agent complied, `auto_generated` only when it didn't. Suppression behaviour is unchanged either way; only the label differs.
- Or keep `auto_generated` as the reason and set `abandonedContent` when the content wasn't a decline. The field already exists on the payload for exactly this (`ambiguous_decline` uses it), it's already used for recoverability elsewhere, and it turns "the agent ignored the preamble" into something you can read rather than infer.

The second also preserves the text the agent wanted to send, which is currently discarded with no record. That's safe by design, but a silently dropped reply with no trace is the failure shape this whole issue series has been about.

src/channels/email/auto-generated-mail.ts
```diff
@@ -0,0 +1,186 @@
...
+  return OOO_SUBJECT.test(subject.trim()) ? ['ooo_subject'] : [];
+}
+
+/**
+ * Classify whether an inbound email is machine-generated from message signals.
+ *
+ * Independent of contact `kind`. Does not judge actionability — that stays
+ * with the agent (or ceo-inbox LLM carve-outs), not dispatch regexes.
+ */
+export function classifyAutoGeneratedMail(input: AutoGeneratedMailInput): AutoGeneratedMailClassification {
```
josephfung: **Cost note, not a defect — every auto-generated message now costs a coordinator turn.**

Raising this because I argued for the design that causes it, so it's fair to state the bill. The short-circuit path is gone, so `Precedence: bulk` newsletters, list mail, OOO replies, and RSVP notices all get an LLM turn now. Before this PR they got a turn *and* an unwanted reply, so it isn't a regression against `main` — but it does drop the token saving the earlier version of this PR advertised.

A concrete sense of scale from our own incident: 15 declines arrived in about 20 minutes on one thread family. Under this design that's 15 coordinator turns, each of which should end in `NO_REPLY`. A busy inbox with newsletters and mailing lists will run more.

I still think this is the right trade — the alternative is dispatch guessing at actionability, which is what we just removed — and correctness over cost is the right default. But it's a genuine decision rather than an obvious one, and the data to make it exists: `outbound.no_reply` rows with `reason: 'auto_generated'` in the audit log will give an exact per-day count once this has run for a week. Worth checking that number before assuming it's fine.

If it turns out to be large, the cheapest lever is probably not reintroducing dispatch judgment but narrowing which signals warrant a turn at all. `precedence_bulk` on its own is the highest-volume and lowest-value signal — a mailing-list digest almost never contains something to escalate — whereas `auto_submitted` and `calendar_reply` are lower volume and likelier to matter. That would be a signal-level policy rather than a content-level guess, which is the distinction that made the carve-outs wrong. Not suggesting it now; noting where the lever is if the number comes back high.

josephfung (commented):
## Re-review of `b7e82c3b`

All six threads resolved. CI green. This is a better design than what I proposed on two of the six, so worth saying where.

The carve-out removal is the right call and it was executed as a real deletion rather than a narrowing — the regexes, the `actionable` flag, the `AutoGeneratedCarveOut` type, and the short-circuit dispatch path are all gone. Two knock-on wins came free: `OutboundNoReplyPayload` no longer needs conditional field semantics (my suggestions there were both workarounds for a fabricated row that now doesn't exist), and the carve-out knowledge moved into the preamble, where it can say things a regex can't — "marketing 'expires'/'last chance' copy is not actionable", "security alert **from a verified sender**". That's the nuance the regexes were trying and failing to encode, given to the instrument that can evaluate it.

I checked the two things that could have quietly undone the fix and both hold: `liveTurn` derives from `systemRole === 'principal'` (the structural binding, same signal as the elevated-skill gate — not the unreliable `role` field), and narrowing calendar detection to explicit `method=REPLY`/`CANCEL` still covers the incident, since those Exchange declines carried the method parameter.

Nothing blocking. Four follow-ons below: one stale prompt, one trivial cleanup, one observability gap I'd genuinely want before this runs in prod, and one cost question that is a decision rather than a defect.

josephfung:
@Coderabbitai review

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Added
 
+- **Auto-generated mail detection** — inbound email is classified from message signals (not contact `kind`); replies are suppressed at dispatch while the agent still judges escalation. (#1734)
```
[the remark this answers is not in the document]
josephfung: Fixed in `bbefdd6e` — Unreleased entry is now 11 words after the em dash:

> classifies inbound email from message signals and suppresses replies at dispatch.

src/channels/email/auto-generated-mail.ts
```diff
@@ -0,0 +1,186 @@
...
+  attachments: AutoGeneratedMailInput['attachments'],
+): AutoGeneratedSignal[] {
+  const signals: AutoGeneratedSignal[] = [];
+
+  for (const att of attachments) {
+    // isCalendarInvite is consumed so the flag has a live reader (#1734 AC),
+    // but method= is what distinguishes REPLY/CANCEL from REQUEST/unknown.
+    const looksCalendar =
+      att.isCalendarInvite === true ||
+      parseCalendarMethod(att.contentType) !== undefined ||
```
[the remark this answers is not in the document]
josephfung: Fixed in `bbefdd6e`. `looksCalendar` now requires calendar MIME (`text/calendar` / `application/ics` / `text/x-vcalendar`), `isCalendarInvite`, or a `.ics` filename before `method=` is consulted — so `application/pdf; method=REPLY` no longer classifies as `calendar_reply`. Covered by a negative unit test.

src/dispatch/dispatcher.ts
```diff
@@ -1116,6 +1145,54 @@ export class Dispatcher {
...
     }
 
+    // Auto-generated inbound (#1734): never relay a reply to the sender. The agent
+    // may still escalate via bullpen/Signal; only the dispatcher relay is suppressed.
+    //
+    // Observability: classify the agent response first so audit can tell whether the
+    // preamble worked (agent returned NO_REPLY → agent_declined / empty / ambiguous)
+    // or the backstop caught narration (auto_generated + abandonedContent). Without
+    // this split every auto-gen turn looks identical in the log.
+    if (routing.autoGeneratedSuppress) {
```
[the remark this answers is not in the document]
josephfung: Agreed — relay suppress alone left a skill-path hole. Fixed in `bbefdd6e`:

- Dispatch stamps trusted `autoGeneratedSuppress` + `autoGeneratedSenderId` on task metadata (channel-supplied values are scrubbed).
- `email-reply` / `email-send` call `autoGeneratedSenderBlockReason` before `OutboundGateway.send()` and refuse when the recipient is the machine sender.
- Email (or Signal/SMS) to a *different* recipient — principal escalation — stays open.

Tests cover both the block and the principal-escalation allow path.

josephfung:
@coderabbitai review

Issue #274, opened 2026-04-10
Spec 03 - #5: feat: built-in skill-registry skill — agent-invocable discovery via SkillRegistry.search()


josephfung:
## Overview

`SkillRegistry.search()` already exists in `src/skills/registry.ts`. This issue wraps it in a built-in skill and wires `allow_discovery: true` so the coordinator can search for capabilities it wasn't explicitly pinned to.

## What to build

1. `skills/skill-registry/skill.json` + `handler.ts` — thin wrapper that calls `registry.search({ query })` and returns matching skill names and descriptions
2. In the agent runtime tool-list builder: if agent config has `allow_discovery: true`, append `skill-registry` to the tool definitions regardless of `pinned_skills`

The coordinator already has `allow_discovery: true` in `agents/coordinator.yaml` — this just makes it functional.

## What's already there

- `SkillRegistry.search()` — substring match against name and description
- `allow_discovery` field — parsed by the agent loader, but currently unused at runtime

## Impact

The coordinator gains the ability to discover skills it wasn't pinned to — useful as the skill library grows (GDrive tools, MCP-discovered tools, etc.).

## Notes

- The `skill-registry` skill needs `infrastructure: true` so the handler can access the registry instance via `ctx`
- Or: pass the registry search results at context-build time (avoids needing infrastructure flag)
- Approval/safety gate for first-use elevated skills (also spec 03) is a separate concern — not in scope here

Issue #271, opened 2026-04-10
Spec 03 - #2: feat: Google Drive integration via MCP server


josephfung:
## Overview

Wire the official `@modelcontextprotocol/server-gdrive` MCP server to give Curia read/write access to Google Drive — enabling persistent documents like expense trackers, job application trackers, and other long-lived structured content.

## Depends on

- #270 (MCP client layer)

## What to do

1. Configure `@modelcontextprotocol/server-gdrive` in `config/skills.yaml`
2. Wire credentials (service account or OAuth) so the MCP server has Drive access
3. Share a designated root folder with the service account — Curia operates within that folder only
4. Document setup in a runbook (credential generation, folder sharing, config values)
5. Add the GDrive tools to the coordinator's `pinned_skills` or `allow_discovery` flow

## Tools exposed (via MCP server)

The official server exposes search, read, and create operations on Docs and Sheets. Exact tool names depend on the server version — confirm at integration time.

## Credential approach

Prefer service account over full OAuth to avoid building an OAuth redirect flow. Joseph shares the target Drive folder with the service account. Service account key stored as `GOOGLE_SERVICE_ACCOUNT_KEY` in environment.

## Notes

- No skill handler code to write — this is purely config and credential wiring
- Root folder constraint is enforced by which folders are shared with the service account, not by code
- If the official MCP server doesn't cover all needed operations (e.g. updating a specific Sheet range), supplement with a local skill

PR #1751, opened 2026-09-09
fix(docker): drop pnpm metadata cache and store from the runtime image

### LINKED ISSUE #1750: Trivy reports 12 CRITICAL secret findings from pnpm's metadata cache shipped in the runtime image
The published image ships pnpm's registry **metadata cache** (`/root/.cache/pnpm/v11/metadata-full/*.jsonl`) and its content-addressable **store** (`/root/.local/share/pnpm`). Neither is used at runtime — `CMD` invokes `./node_modules/.bin/tsx` directly and no install ever runs inside the container.

The metadata cache stores one JSONL per package holding every published version's manifest, README included. Trivy's secret scanner reads those READMEs and reports the example tokens in them as leaked credentials. `@octokit/auth-token`'s `createTokenAuth("ghp_…")` documentation snippet alone accounts for **twelve CRITICAL** findings:

| Alerts | Rule |
|---|---|
| #275–#282 | `github-pat` |
| #271–#274 | `github-app-token` |

These are false positives, but they are the only CRITICAL entries in the Security tab and they bury real findings.

Note the cache covers **devDependencies too**, not just what ships: the runtime stage's `pnpm add -w --save-prod --prod tsx` re-resolves the entire workspace, so metadata for the full dependency graph lands on disk.

The runtime stage already strips the global npm/npx and the corepack cache for exactly this reason ([Dockerfile](https://github.com/josephfung/curia/blob/main/Dockerfile)); the pnpm caches were simply missed.

### Acceptance criteria

- [ ] The runtime stage removes `/root/.cache/pnpm` and `/root/.local/share/pnpm` alongside the existing corepack-cache cleanup.
- [ ] The cleanup is fail-closed: the build asserts the runtime entry point still executes after the removal, so a future pnpm/base-image layout change cannot silently no-op the cleanup or leave a broken `tsx` behind a green build.
- [ ] A Trivy image scan of the rebuilt image reports **zero** `github-pat` / `github-app-token` findings.
- [ ] Code-scanning alerts #271–#282 close on the next scan of `main`.
- [ ] The container still boots and passes its `HEALTHCHECK` (`tsx dist/index.js` resolves — node_modules hardlinks survive the store removal).

josephfung:
Closes #1750.

The runtime image shipped pnpm's registry **metadata cache** (`/root/.cache/pnpm/v11/metadata-full/*.jsonl`) and its content-addressable **store** (`/root/.local/share/pnpm`). Neither is reachable at runtime — `CMD` invokes `./node_modules/.bin/tsx` directly and no install runs inside the container.

The metadata cache keeps one JSONL per package holding every published version's manifest, README included. Trivy's secret scanner reads those READMEs and reports the example tokens in them as leaked credentials. `@octokit/auth-token`'s `createTokenAuth("ghp_…")` documentation snippet alone produced **twelve CRITICAL** findings — the only CRITICAL entries in the Security tab:

| Alerts | Rule |
|---|---|
| #275–#282 | `github-pat` |
| #271–#274 | `github-app-token` |

Note the cache covers **devDependencies too**, not just what ships: the runtime stage's `pnpm add -w --save-prod --prod tsx` re-resolves the whole workspace, so metadata for the full dependency graph lands on disk.

The stage already strips the global npm/npx and the corepack cache for the same reason; the pnpm caches were simply missed.

## Why removing the store is safe

`node_modules/.pnpm` entries are **hardlinks into** the store. Unlinking the store paths decrements the link count but leaves those inodes referenced by `node_modules`, so the installed tree is untouched.

## Fail-closed check

`rm -rf` exits 0 whether or not it removed anything. On its own this cleanup would silently become a no-op if a future pnpm or base-image change relocated the store — and a store layout that *copied* instead of hardlinking would leave a broken `tsx` behind a green build. Executing `tsx --version` after the removal turns both into a hard build failure. This mirrors the existing fail-closed assertions around the npm/npx removal a few lines above.

## Verification

- [ ] `docker build` succeeds (the new `tsx --version` assertion runs inside it).
- [ ] Trivy image scan reports zero `github-pat` / `github-app-token` findings.
- [ ] Alerts #271–#282 close on the next scan of `main`.

Trivy's image job doesn't run on pull requests (only on schedule, dispatch, or a push to `main` touching the Dockerfile), so the scan is dispatched against this branch to verify before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Security
 
+- **Docker image** — drop pnpm's metadata cache and store from the runtime image; they aren't used at runtime. (#1750)
 - **Docker image** — remove unused global npm so bundled brace-expansion/tar CVEs clear. (#1568)
 
 ## [0.42.0] — 2026-07-26 — "KITT"
--- a/Dockerfile
+++ b/Dockerfile
@@ -129,16 +129,19 @@ COPY package.json pnpm-lock.yaml pnpm-workspace.yaml ./
 COPY --from=build /app/packages/shared-types ./packages/shared-types
 # Same transient-network guard as the build stage (#1699).
 COPY docker/pnpm-retry.sh /usr/local/bin/pnpm-retry
-RUN pnpm-retry pnpm install --frozen-lockfile --prod
 
+# Install prod deps, add the runtime tsx, and delete every build-only artifact —
+# all in ONE layer. The single-layer part is load-bearing, not stylistic: see the
+# "Why one RUN" note below.
+#
 # tsx is needed at runtime: skill handlers are .ts files loaded via dynamic
 # import(), and they use ESM .js extension mapping (e.g., import from './foo.js'
 # resolving to foo.ts). Node's --experimental-strip-types doesn't handle this;
 # tsx does, and it's already used for dev (pnpm dev).
 #
-# Three flags are load-bearing (verified against pnpm 11.7.0, the pinned pkg mgr).
-# Since packages/shared-types made this a real workspace, this dir is now a workspace
-# ROOT and a bare `pnpm add tsx` fails a chain of guards:
+# Three flags on the `pnpm add` are load-bearing (verified against pnpm 11.7.0, the
+# pinned pkg mgr). Since packages/shared-types made this a real workspace, this dir is
+# now a workspace ROOT and a bare `pnpm add tsx` fails a chain of guards:
 #   -w: without it, a `pnpm add` at a workspace root is rejected with
 #     ERR_PNPM_ADDING_TO_ROOT. tsx is a genuine root-level runtime dep, so the root is
 #     the correct target (the shared-types member copied above lets it re-resolve).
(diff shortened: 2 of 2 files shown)


josephfung:
**Verified in CI.** The first approach — a dedicated later layer that ran `rm -rf` on the caches — did *not* work, even though the removal demonstrably executed (`tsx v4.23.13` printed from the fail-closed assertion in the build log). A dispatched Trivy image scan on that commit still returned all 14 findings.

Trivy's secret scanner reports matches from **any layer in the image**, not just the whiteout-applied final filesystem. That is the correct stance — the content is still extractable from the image tarball — so a later-layer `rm` can never clear a secret finding. Folding the install and the cleanup into a single `RUN` means the cache is never committed to any layer at all.

Two dispatched scans of `trivy.yml` against this branch:

| Commit | Approach | `trivy-image` results |
|---|---|---|
| d0a2ac0e | cleanup in a separate later layer | **14** (12 secret + 2 `qs`) |
| f0deaec1 | install + cleanup in one layer | **2** |

The two remaining results are the `qs@6.15.3` CVEs, which are #1753's job:

```
CVE-2026-82417  app/node_modules/.pnpm/qs@6.15.3/node_modules/qs/package.json
CVE-2026-82562  app/node_modules/.pnpm/qs@6.15.3/node_modules/qs/package.json
```

All twelve `github-pat` / `github-app-token` findings are gone, and the build's `tsx --version` assertion passed — so `node_modules` survives the store removal, as expected from the hardlink semantics.

Issue #1699, opened 2026-09-01
docker-publish: transient npm socket drop during emulated arm64 pnpm install fails the whole publish, silently staling :edge


josephfung:
## Context

`docker-publish` failed on the merge of PR #1696 to `main` ([run 33556099876](https://github.com/josephfung/curia/actions/runs/33556099876), sha `01f696a3`). The cause was a transient npm registry socket drop, not anything in the commit — it touched no `package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, `.npmrc`, or `Dockerfile`.

Failing step: `Build and push`, on the **`linux/arm64`** leg of the `curia` image.

```
#32 [linux/arm64 build 8/17] RUN pnpm install --frozen-lockfile
TypeError: terminated
    at Fetch.onAborted (node:internal/deps/undici/undici:13842:53)
  [cause]: SocketError: other side closed
    code: 'UND_ERR_SOCKET',
    socket: {
      remoteAddress: '104.16.4.34',   # registry.npmjs.org (Cloudflare)
      remotePort: 443,
      bytesRead: 2423403
    }
```

The registry was measurably degraded at that moment — the concurrent amd64 leg logged
`[WARN] Request took 12653ms: https://registry.npmjs.org/@azure%2Fidentity` in the same window.

## Why it fails the whole publish

- `.github/workflows/docker-publish.yml` builds `platforms: linux/amd64,linux/arm64` with `push: true`.
- The arm64 leg runs under QEMU emulation on an amd64 runner, so `pnpm install` of ~1012 packages is slow — the longest possible exposure window to a registry blip.
- There is **no fetch-retry configuration anywhere**: no `.npmrc` in the repo, and no `fetch-retries` / `fetch-retry-maxtimeout` / `network-concurrency` settings in `pnpm-workspace.yaml` or the `Dockerfile`. A single dropped socket surfaces as an unhandled `error` event on a Readable, which kills Node outright rather than being retried.
- buildx pushes only after **all** platforms build, so one flaky leg aborts the push entirely.

## Impact

`ghcr.io/josephfung/curia:edge` silently stays at the previous commit. Because `deploy.sh` pulls `:edge`, a deploy after a failed publish ships older code — and its migrations — while appearing to succeed. The staleness is not surfaced anywhere except the red workflow run.

`curia-postgres` is a separate matrix leg and is unaffected.

## Proposed fix

Make the install resilient rather than relying on a clean network for the full emulated build:

[...]

PR #1757, opened 2026-09-09
fix(docker): keep pnpm's store in the runtime image

### LINKED ISSUE #1756: Removing pnpm's store from the runtime image broke the deploy: downstream build re-downloads the whole dependency graph
#1750 removed pnpm's metadata cache **and** its content-addressable store from the runtime image. Removing the store was a mistake: `curia-deploy`'s thin downstream image depends on it, and dropping it degraded the deploy build into a full re-fetch of the dependency graph.

`deploy/compose/Dockerfile.curia` builds `FROM ghcr.io/josephfung/curia:edge`, switches to root, and runs `pnpm add`. It carries an explicit comment about the coupling:

> the core image built its node_modules as root with the default `HOME=/root`, so the store actually lives at `/root/.local/share/pnpm/store/v11`

That is exactly the path #1750 deleted. The downstream build therefore starts with an empty store and a cold metadata cache, and `pnpm add` forces a strict full-workspace re-resolution (the same behaviour that makes `-w --prod --save-prod` necessary in the core image). Both signatures show up in the deploy log:

```
Progress: resolved 378, reused 0, downloaded 0, added 0
[WARN] Request took 1530710ms: https://registry.npmjs.org/@opencode-ai%2Fsdk
[WARN] GET https://registry.npmjs.org/drizzle-orm error (23). Will retry in 10 seconds.
[WARN] GET https://registry.npmjs.org/drizzle-orm error (UND_ERR_SOCKET). Will retry in 10 seconds.
[WARN] GET https://registry.npmjs.org/@opencode-ai%2Fsdk error (ECONNRESET). Will retry in 10 seconds.
```

- **`reused 0`** — the empty store; with a populated one pnpm reports a non-zero reuse count.
- **`@opencode-ai/sdk`** is a `promptfoo` transitive, i.e. a **devDependency** — the downstream is pulling packuments for the entire workspace graph over the network.

The `ECONNRESET` / `UND_ERR_SOCKET` failures are ordinary registry flakiness and were not introduced here. What changed is exposure: a build that touched the registry lightly now pulls the whole graph, so it is far likelier to land on a bad connection. This is permanent, not a cold-start cost — every downstream rebuild from a new `:edge` pays it.

### Why the store can be kept without losing the security fix

All twelve `github-pat` / `github-app-token` findings (#271–#282) were in the **metadata** cache, `/root/.cache/pnpm/v11/metadata-full/*.jsonl` — the per-package JSONL files that embed README text. Not one was in the store. The store shipped in this image for its entire history and never tripped a Trivy rule; it holds content-addressed blobs, not README prose.

So deleting the metadata cache is the whole security win, and deleting the store was pure cost.

### Acceptance criteria

- [ ] `/root/.local/share/pnpm` is retained in the runtime image, with a comment recording the downstream coupling so it is not deleted again as "unused build residue".
- [ ] `/root/.cache/pnpm` continues to be removed inside the single install layer.
- [ ] A Trivy image scan of the rebuilt image still reports **zero** `github-pat` / `github-app-token` findings, and alerts #271–#282 stay closed.
- [ ] The build's `tsx --version` assertion still passes.
- [ ] A downstream `Dockerfile.curia` build reports a non-zero `reused` count from `pnpm add` again.

josephfung:
Closes #1756. Fixes a deploy regression I introduced in #1750/#1751.

#1750 removed pnpm's metadata cache **and** its content-addressable store from the runtime image. Removing the store was a mistake.

`curia-deploy`'s thin downstream image (`deploy/compose/Dockerfile.curia`) builds `FROM ghcr.io/josephfung/curia:edge`, switches to root, and runs `pnpm add` with an inline `HOME=/root` for the express purpose of resolving the store to `/root/.local/share/pnpm/store/v11` — the path this stage creates. Its own comment says so:

> the core image built its node_modules as root with the default `HOME=/root`, so the store actually lives at `/root/.local/share/pnpm/store/v11`

Deleting it left that build with an empty store and a cold metadata cache. Since `pnpm add` forces a strict full-workspace re-resolution (the same behaviour that makes `-w --prod --save-prod` necessary here), a deploy turned into a full re-fetch of the dependency graph:

```
Progress: resolved 378, reused 0, downloaded 0, added 0
[WARN] Request took 1530710ms: https://registry.npmjs.org/@opencode-ai%2Fsdk
[WARN] GET https://registry.npmjs.org/drizzle-orm error (23). Will retry in 10 seconds.
[WARN] GET https://registry.npmjs.org/drizzle-orm error (UND_ERR_SOCKET). Will retry in 10 seconds.
[WARN] GET https://registry.npmjs.org/@opencode-ai%2Fsdk error (ECONNRESET). Will retry in 10 seconds.
```

- **`reused 0`** — the empty store. With a populated one pnpm reports a non-zero reuse count.
- **`@opencode-ai/sdk`** is a `promptfoo` transitive, i.e. a **devDependency** — the downstream was pulling packuments for the entire workspace graph.

To be fair about causation: the `ECONNRESET` / `UND_ERR_SOCKET` / `error (23)` failures are ordinary registry flakiness and were not introduced here. What changed is *exposure* — a build that touched the registry lightly now pulls the whole graph, so it is far likelier to land on a bad connection. And the cost is permanent, not a cold start: every downstream rebuild from a new `:edge` pays it.

## The security fix is unaffected

All twelve `github-pat` / `github-app-token` findings (#271–#282) were in the **metadata** cache, `/root/.cache/pnpm/v11/metadata-full/*.jsonl` — the per-package JSONL files that embed README text, including `@octokit/auth-token`'s `createTokenAuth("ghp_…")` snippet.

**None were in the store.** It shipped in this image for its entire history without ever tripping a Trivy rule; it holds content-addressed blobs, not README prose. Deleting the metadata cache was the whole win. Deleting the store was pure cost.

So this keeps the store and continues to delete the metadata cache inside the same single layer. The comment now records the downstream coupling explicitly, so it doesn't get deleted again as "unused build residue" — which is exactly how I got here.

## Verification

- [ ] Dispatched Trivy image scan on this branch still reports **zero** `github-pat` / `github-app-token` findings.
- [ ] The build's `tsx --version` assertion still passes.

## Follow-up, not in this PR

The downstream image has no `pnpm-retry` wrapper — this stage deletes it as build-only tooling (#1699), so `Dockerfile.curia` runs a bare `pnpm add` and a single unhandled socket abort kills the deploy outright. Worth fixing on the `curia-deploy` side.

Longer term, the deeper issue is that adding one package downstream forces a re-resolution of the entire workspace. That is the thing actually making deploys network-heavy; the warm store only hides it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Security
 
+- **Docker image** — keep pnpm's store; removing it made the deploy build re-download every package. (#1756)
 - **Semgrep SARIF** — `# nosemgrep`-suppressed findings no longer upload as open code-scanning alerts. (#1754)
 - **Dependency pins** — `qs`, `hono`, and `fflate` raised to patched releases, clearing six advisories. (#1752)
 - **Docker image** — drop pnpm's metadata cache and store from the runtime image; they aren't used at runtime. (#1750)
--- a/Dockerfile
+++ b/Dockerfile
@@ -162,9 +162,6 @@ COPY docker/pnpm-retry.sh /usr/local/bin/pnpm-retry
 #     `createTokenAuth("ghp_…")` documentation snippet alone produced twelve CRITICAL
 #     github-pat / github-app-token findings (code-scanning #271-#282). False positives,
 #     but they were the only CRITICALs in the Security tab and buried everything else.
-#   - /root/.local/share/pnpm — the content-addressable store and pnpm's global bin dir.
-#     node_modules entries are hardlinks INTO the store, so unlinking the store paths
-#     leaves those inodes referenced by node_modules; the installed tree is untouched.
 #   - /root/.cache/node/corepack and /root/.cache/corepack — corepack's cached pnpm
 #     tarballs. Node 24's bundled corepack pre-caches pnpm@11.0.8 during `corepack
 #     enable` even though this project pins pnpm@11.7.0, and 11.0.8 bundles tar@7.5.13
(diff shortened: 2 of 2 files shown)


josephfung:
Follow-up filed: josephfung/curia-deploy#220 — `Dockerfile.curia` has no `pnpm-retry`, because this stage deletes `/usr/local/bin/pnpm-retry` before publishing, so the downstream image has nothing to inherit.

Worth noting for whoever picks it up: the recommended option there is to **stop deleting the wrapper here**, which would be a change to this file. The removal was hygiene ("consistent with stripping unused npm/npx and the corepack cache"), not security, and there is now a real downstream consumer — the same shape of mistake as the store removal this PR is fixing.

PR #1761, opened 2026-09-09
feat(health): probe the Signal voice audio socket in /api/health

### LINKED ISSUE #1760: Dashboard reports "All systems nominal" while Signal voice audio is dead — no health check covers the PulseAudio path
During curia-deploy#221 the shared PulseAudio socket was dead for hours — Signal voice calls could not work at all — and the console dashboard read **"All systems nominal"** with a green **Voice** chip and *"Nothing needs your attention"*.

The dashboard was not wrong. It faithfully renders `/api/health`, and the backend genuinely did not know. This is a missing probe, not a rendering bug.

## What the dashboard actually shows

`flattenHealthChecks()` in `apps/console/src/pages/dashboard/dashboard-utils.ts` iterates `Object.entries(checks)` and expands nested groups (`mcp: { google_workspace }` → `Mcp.Google_workspace`). It is fully dynamic: **any check added to `/api/health` renders as a chip with no console change.** The fix is one-sided — backend only.

## Finding 1 — nothing probes the Signal audio path

`/api/health` has two related checks, and neither covers it:

| Check | What it actually probes |
|---|---|
| `signal` | `checkSignal(signalRpcClient)` — the JSON-RPC socket, i.e. **messaging** |
| `voice` | `checkVoice(voiceLiveKit)` → `livekit.listRooms()` — **LiveKit**, i.e. console/WebRTC voice |

`grep -rn "pulse\|Pulse\|PULSE" src/` returns hits only in `config.ts`, `index.ts` (boot gating), `signal-call-bridge.ts`, and `signal-audio-transport.ts`. **Nothing under `src/health/` references PulseAudio or the shared socket.**

So the green `Voice` chip was telling the truth about LiveKit. There are two independent voice paths and only one is monitored.

## Finding 2 — the boot gate checks configuration, not liveness

`src/index.ts:1982`:

```ts
if (signalRpcClient && voiceRuntimeRef && config.signalPulseSocketPath && channelShouldStart.has('signal')) {
```

and the failure log records `hasPulseSocket: !!config.signalPulseSocketPath`.

That is the *env var being set*, not the socket existing. `SignalCallBridge` is therefore constructed and reported as enabled against a dead audio stack. The failure only surfaces per-call, deep inside `SignalAudioTransport` when `parec`/`pacat` cannot connect — with nothing tying it back to a subsystem that never started.

## Finding 3 — "Needs attention" never reads health at all

`fetchAttention()` aggregates exactly four sources:

```ts
'/api/jobs?status=failed'
'/api/jobs?status=suspended'
'/api/registry/channels'
'/api/registry/email-accounts'
```

`/api/health` is not among them. So **even a hard `fail` today** — `signal: 'fail'`, `db: 'fail'` — leaves the card saying "Nothing needs your attention" beside a red chip and a "Down" banner. Adding a Pulse probe alone will not surface it there. This is a separate coherence gap in the same dashboard and should be decided deliberately, not inherited.

## Finding 4 — a new check will not move the status dot on its own

`aggregateStatus()` enumerates checks **explicitly**:

```ts
const nonCritical: CheckResult[] = [
  checks.signal, checks.email, checks.browser, ...Object.values(checks.mcp),
  checks.nylas_calendar, checks.slack, checks.sms, checks.voice, checks.scheduler,
];
```

A new key renders a chip (Finding 1) but the big dot stays green unless it is added here too. Easy to miss.

## Design notes

**Probe by connecting, not by `stat`.** `test -S` is not sufficient. In curia-deploy#221 the container held a stale socket *inode* at `/tmp/pulse-runtime/pulse/native` from a week-dead daemon — an existence check passes against a corpse. (The shared path happened to be absent in that incident, so a stat would have caught this one, but not the general case.) Connect to the socket the way `checkSignal` exercises the RPC client.

**Shape — two options, and this is a public API surface.** `/api/health` is consumed by the console and by external monitoring, so the response shape is not free to change:

1. *Additive (recommended):* a sibling key, e.g. `signal_voice`. No breaking change; renders as a `Signal_voice` chip automatically. Leaves the `Voice` label ambiguous.
2. *Nested:* `voice: { livekit, signal }` → `Voice.Livekit` / `Voice.Signal` chips, consistent with the existing `mcp` group and self-documenting. **Breaking** — `voice` changes from `CheckResult` to an object, and `aggregateStatus` plus `isCheckResult` consumers must follow.

**`skipped` when not applicable.** Match `checkVoice`'s existing convention: return `skipped` when Signal voice calls are disabled, so instances without the feature do not show a permanently grey or red chip.

**Severity.** A dead Pulse socket should map to `degraded`, not `down` — messaging is unaffected, as curia-deploy#221 demonstrated at length.

## Acceptance criteria

- [ ] `/api/health` reports the liveness of the Signal voice audio path, distinctly from `signal` (messaging) and `voice` (LiveKit).
- [ ] The probe verifies the daemon is reachable, not merely that a socket inode exists — a stale socket from a dead daemon must report `fail`.
- [ ] The check returns `skipped` when Signal voice calls are not enabled.
- [ ] The new check is added to `aggregateStatus`'s `nonCritical` list so a failure moves the overall status to `degraded` (not `down`).
- [ ] The console renders the new chip with no console-side change (confirming `flattenHealthChecks` stays dynamic), and the chip label distinguishes the two voice paths.
- [ ] If the response shape is nested rather than extended, the breaking change to `voice` is called out explicitly in CHANGELOG per the public-API-surface rule.
- [ ] A decision is recorded on whether `/api/health` failures should feed "Needs attention" — today none do, so a red chip sits beside "Nothing needs your attention".
- [ ] The boot-time gate in `src/index.ts` either probes the socket or its log line stops implying liveness (`hasPulseSocket` currently reports only that an env var is set).

josephfung:
Closes #1760.

During curia-deploy#221 the shared PulseAudio socket was dead for hours — Signal voice calls could not carry audio at all — and the dashboard read **"All systems nominal"** with a green **Voice** chip.

Nothing was lying. `signal` probes the JSON-RPC socket (messaging); `voice` probes LiveKit `listRooms()` (console/WebRTC voice). A grep for `pulse` across `src/health/` returned nothing. The console faithfully renders `/api/health`, so this was a missing probe, not a rendering bug.

## The probe connects, it does not stat

`test -S` cannot tell a live daemon from a corpse. The #221 container held a socket inode written by a daemon dead for a week — `srwxrwxrwx native`, owner long gone.

The regression test builds exactly that shape rather than mocking it: listen in a **child process**, `SIGKILL` it so no cleanup runs, then assert the inode survived before probing.

```ts
// The inode must have outlived the process, or this test proves nothing.
expect(statSync(path).isSocket()).toBe(true);
```

An existence check passes that fixture. This probe reports `fail`.

An in-process `server.close()` would unlink the path, and dropping a regular file there fails with `ENOTSOCK` — a different error than production hit. Hence the child process.

## Wiring details that are easy to get wrong

Each has a test, because none is caught by the probe's own unit tests:

- **`aggregateStatus` enumerates checks by hand.** A new key renders a console chip while the overall status dot stays green — precisely the failure mode #1760 is about. I verified the guard bites by removing the wiring and watching the test fail, then restoring it.
- **The dep is gated on the call *bridge* existing**, not on `config.signalPulseSocketPath`. Keying off config would report `fail` on any instance that merely has the env var set — the same conflation of "configured" with "alive" that let #221 sit green.
- **`skipped` when the bridge is absent**, matching `checkVoice`, so instances without Signal voice show no permanently red chip.

## Shape: additive, not nested

`voice: { livekit, signal }` reads better, but `checks` is a public API surface consumed by the console and external monitoring, and changing `voice` from `CheckResult` to an object breaks every consumer. So `signal_voice` is a sibling key.

The console needs **no change** either way — its `checks` type is `Record<string, CheckResult | Record<string, CheckResult>>` and it derives chips by iterating `Object.entries`. A console test pins that contract, since narrowing it later would silently hide backend failures from the operator.

## Also

`hasPulseSocket` in the boot-gate log is renamed `hasPulseSocketPath`. It only ever meant "the env var is set"; reading it as liveness is part of why #221 was invisible.

## Known tradeoff

Connecting without a handshake may make PulseAudio log a short-lived client per probe, and the Docker healthcheck hits `/api/health` every 30s. Accepted deliberately and documented in the code: a handshake-free connect cannot desync from a protocol change, and the alternative is the existence check that started this. If the log noise ever obscures debugging, raise the probe interval rather than weaken it back.

## Verification

- `pnpm typecheck` — clean across root, skills, tests, scripts, and all three workspace packages
- `pnpm lint` — clean (one pre-existing warning in `src/pii/scrubber.ts`, untouched here)
- `pnpm test` — **6375 passed, 0 failed** (388 skipped)
- New: 6 probe tests, 2 service-wiring tests, 1 console contract test

## Related audit

While here I checked the other probes for the same stat-vs-connect shape, per the follow-up request. Findings filed separately as #1762 — `checkBrowser` is the notable one, and it comes with a contradiction in the existing comments worth resolving.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ bus event types) are noted explicitly even in the `0.x` range.
 
 ### Added
 
+- **`signal_voice` health check** — `/api/health` now probes the Signal call audio socket by connecting, not stat'ing. (#1760)
 - **Auto-generated mail detection** — classifies inbound email from message signals and suppresses replies at dispatch. (#1734)
 - **`apps/*` lint coverage** — ESLint now checks console and antfarm; `no-console` stays backend-only. (#1727)
 - **React hooks linting** — `rules-of-hooks` and `exhaustive-deps` now gate the browser apps. (#1727)
--- a/apps/console/src/pages/dashboard/dashboard-utils.test.ts
+++ b/apps/console/src/pages/dashboard/dashboard-utils.test.ts
@@ -60,6 +60,24 @@ describe('formatUptime', () => {
 });
 
 describe('flattenHealthChecks', () => {
+  // #1760 relies on this being fully dynamic: a check added to /api/health must render
+  // as a chip with NO console change. During curia-deploy#221 the dashboard showed
+  // "All systems nominal" while Signal voice audio was dead, and the backend fix was a
+  // new `signal_voice` check. If this function (or HealthResponse['checks']) is ever
+  // narrowed to a fixed key set, that contract breaks silently — the backend reports a
+  // failure the operator never sees. This test is the tripwire.
+  it('renders a chip for a check key it has never seen before (#1760)', () => {
+    const pills = flattenHealthChecks({
+      db: 'ok',
+      signal: 'ok',
+      voice: 'ok',
+      signal_voice: 'fail',
+    });
+    expect(pills).toContainEqual({ name: 'signal_voice', result: 'fail' });
+    // And the failing one must be visually distinct, not merely present.
+    expect(healthCheckPillClass('fail')).toBe('blocked');
+  });
+
   it('flattens top-level checks and expands nested mcp.*', () => {
     const pills = flattenHealthChecks({
       db: 'ok',
--- a/src/channels/http/routes/health.ts
+++ b/src/channels/http/routes/health.ts
@@ -45,6 +45,7 @@ export async function healthRoutes(
           slack: 'skipped' as const,
           sms: 'skipped' as const,
           voice: 'skipped' as const,
+          signal_voice: 'skipped' as const,
           scheduler: 'fail' as const,
         },
       });
--- a/src/health/health-checks.ts
+++ b/src/health/health-checks.ts
@@ -2,10 +2,12 @@
 //
 // Each function is independent, has a hard timeout for async probes, and returns
 // CheckResult. 'skipped' means the service is not configured — never affects the
-// overall health status. Probe-based checks (db, bus, signal, browser, mcp) run
+// overall health status. Probe-based checks (db, bus, signal, signal_voice, browser,
+// mcp) run
 // on every request. Time-based checks (email, scheduler) use startedAt as a
 // grace-period anchor so they don't fail immediately at boot.
 
+import { connect as netConnect } from 'node:net';
 import type { Pool } from 'pg';
 import type { EventBus } from '../bus/bus.js';
 import type { CheckResult } from './types.js';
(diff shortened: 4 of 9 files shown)


josephfung:
CI's first run failed in `tests/integration/contact-dedup-exclusions.test.ts` with `ContactNotFoundError` — unrelated to this change, which touches only `src/health/`, the health route, and the boot-gate log field.

That is the known shared-Postgres integration flake: the integration files run in parallel against one database and a different file dies each run. Re-ran the failed job and it passed clean (`ci=success`, `postgres-image=success`), so the failure did not reproduce. Locally the full suite was green on the first attempt: 6375 passed, 0 failed.

Flagging rather than burying it — that flake is a real cost on every PR and worth its own issue if it hasn't got one.

josephfung:
@Coderabbitai review

PR #1769, opened 2026-09-10
fix(calendar): stop advertising RSVP writes Nylas cannot apply

### LINKED ISSUE #1735: calendar-update-event silently discards attendee response status and reports success
## Summary

`calendar-update-event` advertises that it can set an event's attendees, but an attendee's response status is silently discarded before the request ever leaves the process. The tool then reports success. It also replaces the entire participant array rather than patching one entry, and never controls whether the write re-issues invitations.

Nothing depends on this today — see *History* — so it is latent rather than blocking. It is worth fixing because the tool currently lies: a caller that asks it to record an RSVP gets `success: true` and no change.

## The defects

**1. Attendee status is dropped client-side.** `CreateEventInput.attendees` is typed `Array<{ email: string; name?: string }>` (`src/channels/calendar/nylas-calendar-client.ts:195`) — there is no status field — and `updateEvent` maps attendees to `{ email, name }` when building the request body (`:359`). Any status a caller supplies is discarded before the API call.

The Nylas SDK supports this: `UpdateEventRequest.participants` is typed `Participant[]`, and `Participant` carries a required `status: ParticipantStatus`. We simply never populate it.

**2. Silent success.** `skills/calendar/tools/calendar-update-event/handler.ts` returns the normalized event with `success: true` without comparing what was requested against what the provider reports. A status write that had no effect is indistinguishable from one that worked.

**3. Whole-array overwrite.** `requestBody.participants` is rebuilt from the caller's input, so updating one attendee rewrites the entire list — and since (1) drops status for every entry, it resets everyone. The manifest's own note that *"Attendee changes trigger invitation updates"* is the hazard in (4).

**4. No notification control.** `notifyParticipants` is never set anywhere in the repo, so create and update both inherit the Nylas default. A status-reconcile write has no way to avoid re-issuing invitations to the whole attendee list.

**5. The manifest oversells it.** `tool.json` describes `attendees` as *"list of attendees to set, each with email and optional name"*, which reads as a general attendee-management capability. A caller cannot infer from this that status is unsupported.

Note that `sendRsvp` is not a substitute: it responds as the authenticated grant (first-person, by #1217's design), so it can record the principal's own RSVP but never another attendee's.

## History

This issue was filed as an orphaned-RSVP bug and rewritten twice as evidence came in. Both earlier framings were wrong, and the record is kept in the issue comments rather than rewritten away.

- **Original framing:** RSVP replies land in Curia's mailbox, which cannot apply them, so attendee status goes stale — and Google refuses organizer-set response status. Wrong on both counts: the calendar grant authenticates *as the principal* (#1217), so the principal is the organizer and RSVPs are applied server-side; and the `noreply` result that looked like a provider refusal was defect (1) above, discarding the status before any API call.
- **Second framing:** RSVPs apply for some attendees and not others. Also wrong. The raw ICS shows `ATTENDEE ... PARTSTAT=DECLINED` naming the *work* address — the one that did apply. All 14 declines received are from that address; there is no inbound of any kind from the second address since the events were created. One person was invited under two addresses and replied from one. The calendar was accurate the whole time.

Worth recording for whoever reads a similar report later: Google's `From: calendar-notification@google.com` message is an **informational notification to the organizer**, sent *after* Google has already applied the RSVP server-side. It is not the mechanism by which the RSVP is applied. The RFC 5546 iTIP reply (`From: <attendee>`, `method=REPLY`) is the one that carries `PARTSTAT`. Conflating the two is what produced both wrong framings.

A separate product observation fell out and is **not** tracked here: inviting one contact under two known addresses guarantees a permanently half-answered guest list, which is what made the calendar look broken and sent the calendar agent down the hand-reconciliation path in the first place. That is a contacts/identity concern.

## Acceptance criteria

- [ ] `calendar-update-event` can set a named attendee's response status, and that status reaches the provider request body.
- [ ] It returns a failure when the requested attendee status is not reflected in the provider's response — no silent success.
- [ ] Updating one attendee's status leaves every other attendee's status unchanged.
- [ ] A status-reconcile update does not re-issue invitations to the attendee list.
- [ ] `tool.json` describes what the tool can actually do, including whether a status change notifies attendees.
- [ ] Tests cover: status round-trip through the client, a multi-attendee event where only one status changes, a provider response that does not reflect the requested status, and notification suppression.

## Evidence

Prod `audit_log`, 2026-09-04: `tool.invoke calendar-update-event` at 07:48:55 and 07:53:52; the approved re-execution of `726e16b3` returned both participants at `status: "noreply"` while reporting success. Approval `f2cb3472` expired unused. That `noreply` result is defect (1), not a provider restriction.

josephfung:
## Summary

`calendar-update-event` advertised attendee management, then silently dropped RSVP status before the request left the process and reported success. The issue's original acceptance criteria asked us to send that status through to Nylas. That is not actually possible.

**Nylas rejects organizer-set participant status on PUT** (`"Updating the status for participants is not allowed, only participants can rsvp"`). The SDK's `Participant.status` field is a shared read/write type, not a write capability. Google Calendar's native API can write `attendees[].responseStatus` as organizer; Microsoft Graph cannot set another attendee's response; Nylas fails closed for both. First-person RSVP remains `calendar-respond-to-invite` (`sendRsvp`), by #1217's design.

This PR makes the tool describe and enforce what the providers actually support:

- Reject attendee `status` / `responseStatus` / `participationStatus` instead of dropping them
- Treat `attendees` as a full guest-list replace (`email` + optional `name` only)
- Fail if the provider response omits a requested attendee email
- Add `notifyAttendees` → Nylas `notify_participants`, with an explicit warning that Microsoft and iCloud ignore `false`

Closes #1735

## Changes

- `calendar-update-event` handler validates attendees, refuses RSVP-status writes, verifies the returned guest list, and forwards `notifyAttendees`
- `NylasCalendarClient.updateEvent` never puts `status` on `participants`; `notify_participants` is a query param
- Manifest drops unimplemented `colorId` / `reminders` and documents guest-list replace + Google-only notification suppression
- Calendar agent prompt: do not use update-event to record another guest's RSVP

## Related Issues

Closes #1735

## Checklist

- [ ] Spec docs updated if behavior changes
- [ ] CI passes

### DIFF
--- a/agents/calendar.yaml
+++ b/agents/calendar.yaml
@@ -1,5 +1,5 @@
 name: calendar
-version: "0.7.0"
+version: "0.7.1"
 role: specialist
 description: >
   Calendar domain specialist — scheduling, free/busy, conflict resolution,
@@ -164,8 +164,13 @@ system_prompt: |
   - For accepts, pass `holdMatchCriteria` and a broad hold search window when
     the consult provides them so all holds associated with the scheduling
     conversation are released, not just the hold overlapping the final invite.
-  - Never update a participants array to RSVP. The RSVP skill uses Nylas
-    sendRsvp so only the principal attendee's own status changes.
+  - Never update a participants array to RSVP — not for the principal, and not
+    to record another guest's reply. Nylas rejects organizer-set participant
+    status on event update; Microsoft Graph cannot set another attendee's
+    response. The RSVP skill uses sendRsvp so only the principal attendee's
+    own status changes. `calendar-update-event` attendees replaces the whole
+    guest list (email + optional name) and re-issues invitations; it cannot
+    patch one person's RSVP.
 
   ## Unregistered Calendar Handling
   When `calendar-list-calendars` surfaces an unregistered calendar during any
--- a/agents/calendar.yaml
+++ b/agents/calendar.yaml
@@ -1,5 +1,5 @@
 name: calendar
-version: "0.7.1"
+version: "0.7.2"
 role: specialist
 description: >
   Calendar domain specialist — scheduling, free/busy, conflict resolution,
@@ -164,13 +164,11 @@ system_prompt: |
   - For accepts, pass `holdMatchCriteria` and a broad hold search window when
     the consult provides them so all holds associated with the scheduling
     conversation are released, not just the hold overlapping the final invite.
-  - Never update a participants array to RSVP — not for the principal, and not
-    to record another guest's reply. Nylas rejects organizer-set participant
-    status on event update; Microsoft Graph cannot set another attendee's
-    response. The RSVP skill uses sendRsvp so only the principal attendee's
-    own status changes. `calendar-update-event` attendees replaces the whole
-    guest list (email + optional name) and re-issues invitations; it cannot
-    patch one person's RSVP.
+  - Never update a participants array to RSVP, and never use
+    `calendar-update-event` to record another guest's RSVP. Use
+    `calendar-respond-to-invite` for the principal's own RSVP.
+    `calendar-update-event` attendees replaces the whole guest list
+    (email + optional name).
 
   ## Unregistered Calendar Handling
   When `calendar-list-calendars` surfaces an unregistered calendar during any
--- a/skills/calendar/tools/calendar-update-event/handler.ts
+++ b/skills/calendar/tools/calendar-update-event/handler.ts
@@ -4,63 +4,17 @@
 // Checks the read-only flag before attempting the update.
 //
 // Guest-list writes replace the entire attendee array (email + optional name).
-// Organizer-set RSVP status is not supported: Nylas rejects it on PUT, and
-// Microsoft Graph cannot set another attendee's response. First-person RSVP
-// is calendar-respond-to-invite (sendRsvp).
+// This Nylas tool does not set RSVP status; first-person RSVP is
+// calendar-respond-to-invite (sendRsvp).
 
 import type { ToolHandler, ToolContext, ToolResult } from '../../../../src/skills/types.js';
-import type { CalendarAttendeeInput, CreateEventInput } from '../../../../src/channels/calendar/nylas-calendar-client.js';
+import type { CreateEventInput } from '../../../../src/channels/calendar/nylas-calendar-client.js';
+import {
+  parseWritableAttendees,
+  collectGuestListMismatches,
+  guestListMismatchWarnings,
+} from '../../../../src/channels/calendar/attendee-input.js';
 import { toLocalIso, formatDisplayTimezone } from '../../../../src/time/timestamp.js';
-
-const RSVP_STATUS_KEYS = ['status', 'responseStatus', 'participationStatus'] as const;
-
-const ATTENDEE_STATUS_UNSUPPORTED =
-  'Attendee response status cannot be set through calendar-update-event. ' +
-  'Nylas rejects organizer-set participant status on event update; Microsoft Graph cannot set another attendee\'s response; ' +
-  'only the authenticated principal can RSVP, via calendar-respond-to-invite. ' +
-  'Pass attendees as email and optional name to replace the guest list.';
-
-function isRecord(value: unknown): value is Record<string, unknown> {
-  return typeof value === 'object' && value !== null && !Array.isArray(value);
-}
-
-function parseAttendees(raw: unknown): { ok: true; attendees: CalendarAttendeeInput[] } | { ok: false; error: string } {
-  if (!Array.isArray(raw)) {
-    return { ok: false, error: 'Invalid input: attendees must be an array' };
-  }
-  const attendees: CalendarAttendeeInput[] = [];
-  for (const item of raw) {
-    if (!isRecord(item)) {
-      return { ok: false, error: 'Invalid input: each attendee must be an object with email and optional name' };
-    }
-    if (RSVP_STATUS_KEYS.some((key) => key in item)) {
-      return { ok: false, error: ATTENDEE_STATUS_UNSUPPORTED };
-    }
-    if (typeof item.email !== 'string' || item.email.trim() === '') {
-      return { ok: false, error: 'Invalid input: each attendee must have a non-empty email' };
-    }
-    if (item.name !== undefined && typeof item.name !== 'string') {
-      return { ok: false, error: 'Invalid input: attendee name must be a string when provided' };
-    }
-    const attendee: CalendarAttendeeInput = { email: item.email.trim() };
-    if (typeof item.name === 'string') attendee.name = item.name;
-    attendees.push(attendee);
-  }
-  return { ok: true, attendees };
-}
-
-function missingRequestedEmails(
-  requested: CalendarAttendeeInput[],
-  returned: Array<{ email: string }>,
-): string[] {
-  const returnedSet = new Set(returned.map((p) => p.email.trim().toLowerCase()));
-  const missing: string[] = [];
-  for (const attendee of requested) {
-    const email = attendee.email.trim().toLowerCase();
-    if (!returnedSet.has(email)) missing.push(email);
-  }
-  return missing;
-}
 
 export class CalendarUpdateEventHandler implements ToolHandler {
   async execute(ctx: ToolContext): Promise<ToolResult> {
@@ -113,7 +67,7 @@ export class CalendarUpdateEventHandler implements ToolHandler {
       if (description !== undefined) changes.description = description;
       if (location !== undefined) changes.location = location;
       if (attendees !== undefined) {
-        const parsed = parseAttendees(attendees);
+        const parsed = parseWritableAttendees(attendees);
         if (!parsed.ok) return { success: false, error: parsed.error };
         changes.attendees = parsed.attendees;
       }
@@ -131,32 +85,27 @@ export class CalendarUpdateEventHandler implements ToolHandler {
         notifyAttendees,
       );
 
+      const warnings: string[] = [];
       if (changes.attendees) {
-        const missing = missingRequestedEmails(changes.attendees, event.participants);
-        if (missing.length > 0) {
-          ctx.log.error(
-            { calendarId, eventId, missingCount: missing.length },
-            'calendar-update-event: provider response omitted requested attendees',
+        const { missing, extra } = collectGuestListMismatches(changes.attendees, event.participants);
+        warnings.push(...guestListMismatchWarnings(missing, extra));
+        if (missing.length > 0 || extra.length > 0) {
+          ctx.log.warn(
+            { calendarId, eventId, missingCount: missing.length, extraCount: extra.length },
+            'calendar-update-event: provider guest list did not match the request',
           );
-          return {
-            success: false,
-            error:
-              `Failed to update event: provider response did not include ${missing.length} requested attendee(s). ` +
-              'The guest list replace may not have applied. Re-read the event before retrying.',
-          };
         }
+      }
+      if (notifyAttendees === false) {
+        warnings.push(
+          'notifyAttendees=false is honoured by Google Calendar; Microsoft and iCloud ignore it and still email attendees.',
+        );
       }
 
       ctx.log.info({ calendarId, eventId }, 'Updated calendar event');
       // Format timestamps in the user's local timezone so the confirmation matches
       // what calendar-list-events returns. toLocalIso handles null/invalid values internally.
       const tz = ctx.timezone;
-      const warnings: string[] = [];
-      if (notifyAttendees === false) {
-        warnings.push(
-          'notifyAttendees=false is honoured by Google Calendar; Microsoft and iCloud ignore it and still email attendees.',
-        );
-      }
       return {
         success: true,
         data: {
--- a/skills/calendar/tools/calendar-update-event/handler.ts
+++ b/skills/calendar/tools/calendar-update-event/handler.ts
@@ -2,9 +2,65 @@
 //
 // Updates an existing calendar event with partial field changes.
 // Checks the read-only flag before attempting the update.
+//
+// Guest-list writes replace the entire attendee array (email + optional name).
+// Organizer-set RSVP status is not supported: Nylas rejects it on PUT, and
+// Microsoft Graph cannot set another attendee's response. First-person RSVP
+// is calendar-respond-to-invite (sendRsvp).
 
 import type { ToolHandler, ToolContext, ToolResult } from '../../../../src/skills/types.js';
+import type { CalendarAttendeeInput, CreateEventInput } from '../../../../src/channels/calendar/nylas-calendar-client.js';
 import { toLocalIso, formatDisplayTimezone } from '../../../../src/time/timestamp.js';
+
+const RSVP_STATUS_KEYS = ['status', 'responseStatus', 'participationStatus'] as const;
+
+const ATTENDEE_STATUS_UNSUPPORTED =
+  'Attendee response status cannot be set through calendar-update-event. ' +
+  'Nylas rejects organizer-set participant status on event update; Microsoft Graph cannot set another attendee\'s response; ' +
+  'only the authenticated principal can RSVP, via calendar-respond-to-invite. ' +
+  'Pass attendees as email and optional name to replace the guest list.';
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value);
+}
+
+function parseAttendees(raw: unknown): { ok: true; attendees: CalendarAttendeeInput[] } | { ok: false; error: string } {
+  if (!Array.isArray(raw)) {
+    return { ok: false, error: 'Invalid input: attendees must be an array' };
+  }
+  const attendees: CalendarAttendeeInput[] = [];
+  for (const item of raw) {
+    if (!isRecord(item)) {
+      return { ok: false, error: 'Invalid input: each attendee must be an object with email and optional name' };
+    }
+    if (RSVP_STATUS_KEYS.some((key) => key in item)) {
+      return { ok: false, error: ATTENDEE_STATUS_UNSUPPORTED };
+    }
+    if (typeof item.email !== 'string' || item.email.trim() === '') {
+      return { ok: false, error: 'Invalid input: each attendee must have a non-empty email' };
+    }
+    if (item.name !== undefined && typeof item.name !== 'string') {
+      return { ok: false, error: 'Invalid input: attendee name must be a string when provided' };
+    }
+    const attendee: CalendarAttendeeInput = { email: item.email.trim() };
+    if (typeof item.name === 'string') attendee.name = item.name;
+    attendees.push(attendee);
+  }
+  return { ok: true, attendees };
+}
+
+function missingRequestedEmails(
+  requested: CalendarAttendeeInput[],
+  returned: Array<{ email: string }>,
+): string[] {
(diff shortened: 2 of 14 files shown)


josephfung:
@coderabbitai review

src/channels/calendar/nylas-calendar-client.ts
```diff
@@ -475,6 +490,30 @@ export class NylasCalendarClient {
 
   // -- Helpers --
 
+  /**
+   * Map Curia attendees to the Nylas write shape (email + name only).
+   *
+   * Nylas `Participant.status` is required on the *read* type and reused on
```
josephfung: **This claim is doing all the work in the PR, and I can't find any evidence for it.**

The quoted error — `"Updating the status for participants is not allowed, only participants can rsvp or change the status of an event"` — is asserted here, in the commit message, in the PR body, in `tool.json`, and in the calendar agent's prompt. I searched the repo, the bundled Nylas SDK, and the PR discussion; there is no test that exercises it, no linked doc, no captured response, no spike. It is a bare assertion repeated five times, which reads as corroboration but isn't.

I want to be scrupulous about my own position here, because I'm the one who claimed the opposite in #1735. My reasoning was that `UpdateEventRequest.participants` is `Participant[]` and `Participant.status` is required, therefore the write path accepts it. Your rebuttal — shared read/write type, not a write capability — is correct, and my inference was unsound. But that argument only demolishes my evidence. It doesn't supply yours. We have both reasoned from type shapes rather than from the API.

The difference is what happens when the claim is wrong. Mine sat in an issue where it got caught. This one becomes a hard `throw` in the client, a validation rejection in the handler, a permanent line in the tool manifest, and an instruction to the calendar agent — and it closes #1735 as impossible. If Nylas does accept participant status on PUT (or starts to), nothing here will ever discover that; the tool will keep refusing and the prompt will keep telling the agent it can't be done.

One live PUT against a throwaway event on the principal grant settles it, and the response body becomes the citation. A link to the Nylas docs or changelog would do just as well. Either way I'd want the source recorded next to the claim rather than the claim alone, so the next person can re-check it when Nylas ships a change.

Secondary point on the same comment: it says "Google Calendar can write `attendees[].responseStatus` as organizer; Microsoft Graph cannot; Nylas fails closed for both." If that's accurate, then the capability exists in a system Curia already talks to — google-workspace MCP — and "impossible" is really "impossible through Nylas". #1330 already tracks Curia using workspace-mcp for calendar. Closing #1735 outright forecloses that without naming it. I'd rather the PR said "not available on the Nylas path; the Google-native path is #1330's territory" and left a pointer, than have the door shut on a capability that demonstrably exists one integration over.

skills/calendar/tools/calendar-update-event/tool.json
```diff
@@ -12,13 +12,13 @@
     "end": "timestamp? (new event end time)",
     "description": "string?",
     "location": "string?",
-    "attendees": "object[]? (list of attendees to set, each with email and optional name)",
+    "attendees": "object[]? (full guest-list replacement: each { email, name? }. Cannot set RSVP status. Replaces every attendee, it does not patch one.)",
```
josephfung: **Does a full guest-list replace preserve the existing attendees' RSVPs?**

This is now the tool's headline documented behaviour, and I don't think anyone has established what it actually does to RSVP state. The two possibilities are very far apart:

- Nylas/Google match the incoming `participants` by email and retain each existing attendee's `status`. Then adding one guest is safe and this is a fine primitive.
- The replace is literal and every retained attendee drops back to `noreply`. Then "add one guest to a five-person meeting" silently wipes four RSVPs and re-invites everyone, and the tool is more dangerous after this PR than the docs suggest — because the docs now actively encourage using `attendees` as the way to manage a guest list.

The irony is that the second case is the exact failure #1735 was originally filed about: attendee status not reflecting reality. It would be an unfortunate way to close it.

The manifest says "Replaces every attendee, it does not patch one", which describes *membership* but is silent on *status* — and status is the thing this whole issue has been about. If the answer is "statuses are preserved", say so, because that's the reassurance a reader needs. If it's "statuses reset", that belongs in the description in strong terms, and `action_risk: high` starts looking like an understatement.

The verification added in the handler is well-placed to answer this if you want it enforced rather than documented: it already reads `event.participants` back. Comparing returned statuses for *retained* attendees against a read-before-write would catch a destructive replace at the point it happens. That's more work than this PR needs to take on, but the question should at least be answered before the docs tell agents this is the normal way to edit a guest list.

skills/calendar/tools/calendar-update-event/handler.ts
```diff
@@ -2,26 +2,83 @@
...
+    attendees.push(attendee);
+  }
+  return { ok: true, attendees };
+}
+
+function missingRequestedEmails(
+  requested: CalendarAttendeeInput[],
+  returned: Array<{ email: string }>,
+): string[] {
+  const returnedSet = new Set(returned.map((p) => p.email.trim().toLowerCase()));
```
josephfung: **`p.email.trim()` will throw if the provider returns a participant without an email.**

The internal response type at `nylas-calendar-client.ts:112` declares `participants[].email` as required `string`, but that's Curia's assertion about the payload, not a guarantee. The Nylas SDK types it `email?: string`, and Nylas's own docs note participants may be rooms or resources — the class of entry most likely to arrive without a mailbox.

If one shows up, this throws a `TypeError` inside the handler's `try`, and the caller gets `Failed to update event: Cannot read properties of undefined (reading 'trim')`. The update itself will have already succeeded, so the agent is told the write failed when it didn't — and the suggested recovery in the sibling error path is "re-read the event before retrying", which would lead it to retry a write that already applied. Given the tool is `action_risk: high` and re-issues invitations, a spurious retry is not free.

`(p.email ?? '')` or filtering falsy emails out of the set both fix it. Worth a test with a participant missing `email`, since the current tests all construct well-formed responses.

Separately, the check is one-directional: it catches requested attendees the provider dropped, but not participants the provider returned that weren't requested. On a full-replace operation an unexpected extra is equally good evidence the replace didn't apply as asked. Cheap to add while you're in here, and it would also surface the case where the provider silently retains someone you meant to remove — which for a guest list is the more consequential direction.

agents/calendar.yaml
```diff
@@ -164,8 +164,13 @@ system_prompt: |
   - For accepts, pass `holdMatchCriteria` and a broad hold search window when
     the consult provides them so all holds associated with the scheduling
     conversation are released, not just the hold overlapping the final invite.
-  - Never update a participants array to RSVP. The RSVP skill uses Nylas
-    sendRsvp so only the principal attendee's own status changes.
+  - Never update a participants array to RSVP — not for the principal, and not
+    to record another guest's reply. Nylas rejects organizer-set participant
```
josephfung: **The prompt states the vendor claim as settled fact, which is where an unverified claim is hardest to dislodge.**

Code that's wrong gets found when someone reads it. A prompt line that's wrong shapes the agent's behaviour silently and indefinitely — it will decline to attempt something and won't surface why, so there's no failure to investigate. If the Nylas claim turns out to be wrong or becomes wrong after a vendor change, this line keeps the agent away from a working capability with no signal that anything is off.

The behavioural instruction is right regardless of the vendor detail: the agent should not be using `calendar-update-event` to record someone else's RSVP, and it should reach for `calendar-respond-to-invite` for first-person RSVP. That part I'd keep exactly as-is. It's the causal explanation — "Nylas rejects organizer-set participant status on event update; Microsoft Graph cannot set another attendee's response" — that's carrying an unverified assertion into the agent's world model, where it can generalise in ways nobody intended.

The handler already returns a clear, specific error if the agent tries. That's the durable mechanism: enforcement in code, with the reason attached to the failure at the moment it matters. The prompt can say "this tool cannot record another guest's RSVP; use `calendar-respond-to-invite` for the principal's own" and stop there, letting the error message carry the why. Same behaviour, and nothing to go stale.

The version bump to `0.7.1` and the tripwire test update are handled correctly.

skills/calendar/tools/calendar-update-event/tool.json
```diff
@@ -12,13 +12,13 @@
...
     "conferencing": "object?",
-    "colorId": "string?",
-    "reminders": "object?"
+    "notifyAttendees": "boolean? (default true. false suppresses invitation emails on Google only; Microsoft and iCloud always notify.)"
```
josephfung: **The same dishonest manifest is still sitting in `calendar-create-event`.**

Removing `colorId` and `reminders` here is a genuine fix — I checked, and neither handler destructures them, so both tools have been advertising inputs that do nothing. But `skills/calendar/tools/calendar-create-event/tool.json:16-17` still lists both, and its `attendees` is documented as bare `"object[]?"` with no shape at all, which is less honest than the version this PR just replaced.

The asymmetry gets worse now that the client throws. `toWritableParticipants` is shared by create and update, so passing `status` to `calendar-create-event` no longer drops it silently — it raises `Cannot set attendee response status on create/update; Nylas allows RSVP only via sendRsvp` from the client, which surfaces as a generic `Failed to create event: …`. Update gets a validated, actionable error at the boundary; create gets the same rejection wrapped in a less helpful shape, from a manifest that never warned it was possible. An agent that learns the constraint from update's error has no reason to expect it applies to create.

This is a small extension — the same `parseAttendees` helper and the same manifest language applied one directory over — and it would leave the calendar tools consistent rather than trading one inconsistency for another. If you'd rather keep this PR tight, a follow-up issue is fine, but I wouldn't leave it undocumented: the create path's behaviour changed in this PR and its manifest doesn't reflect that.

One small thing on this file: the version went `1.0.0` → `1.1.0`, but removing two advertised inputs is a contract removal. Nothing consumes these manifests with semver resolution today so it's cosmetic, and the repo's convention (minor for new user-facing capability) arguably covers `notifyAttendees` — just noting it in case manifest versions ever start being enforced.

josephfung (commented):
## Review

The reframe is right and it's a better answer than the acceptance criteria I wrote on #1735. "Make the tool do X" was the wrong instruction if X isn't supported; "make the tool refuse X loudly and say why" is the honest version, and rejecting at the boundary instead of dropping silently fixes the actual complaint. Removing the phantom `colorId` / `reminders` inputs is a real catch — I confirmed the handler never read them.

I also want to concede a point directly. My issue asserted the capability existed because `UpdateEventRequest.participants` is typed `Participant[]` and `Participant` carries a required `status`. This PR's rebuttal — that it's a shared read/write type, not a write capability — is correct reasoning, and my evidence was never sufficient. That matters for what follows: it establishes that *my* claim was unfounded, not that the opposite one is proven. Both of us reasoned from indirect signals.

Which is the substance of my main concern.

### Main concern

**The load-bearing claim has no citation anywhere, and it's now enforced in code.** The assertion that Nylas returns `"Updating the status for participants is not allowed, only participants can rsvp or change the status of an event"` on PUT appears in five places — commit message, PR body, client doc comment, tool manifest, agent prompt — and I can't find its source in the repo, the SDK, the PR discussion, or any linked evidence. The error string is specific enough to look real, which is exactly what makes it worth pinning down rather than assuming.

It's cheap to settle: one PUT against a throwaway event on the principal grant, or a link to the Nylas docs/changelog stating it. Detail inline on `nylas-calendar-client.ts:496`.

### Also worth addressing

1. **Does a guest-list replace preserve existing RSVPs?** (`tool.json:15`) Nobody has established this, and it's now the tool's primary documented behaviour.
2. **`missingRequestedEmails` can throw on a participant with no email** (`handler.ts:56`), surfacing as a confusing generic failure.
3. **`calendar-create-event` still has the identical dishonest manifest** (`tool.json:17`) — same phantom inputs, and it now throws a worse error on the same rejected input.
4. **The agent prompt states the vendor claim as settled fact** (`calendar.yaml:168`), which is where an unverified claim does the most damage.

### Checked and fine

- `colorId` / `reminders` really were unimplemented — neither handler destructures them.
- The `notifyAttendees` → `notify_participants` plumbing is correct, and only setting the query param when the caller passes it keeps existing calls byte-identical.
- Surfacing the Microsoft/iCloud caveat as a returned `warnings` array rather than burying it in a doc comment is the right call — the agent sees it at the point of use.
- `action_risk` correctly stays `high`.
- CI is green.

agents/calendar.yaml
```diff
@@ -164,8 +164,13 @@ system_prompt: |
...
+    status on event update; Microsoft Graph cannot set another attendee's
+    response. The RSVP skill uses sendRsvp so only the principal attendee's
+    own status changes. `calendar-update-event` attendees replaces the whole
+    guest list (email + optional name) and re-issues invitations; it cannot
```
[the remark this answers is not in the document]
josephfung: @coderabbitai This comment is on c494d078. The current prompt (3d9d28c7) no longer says guest-list replace re-issues invitations — it only says attendees replaces the guest list.

The Google notifyAttendees=false exception and the Microsoft/iCloud limitation already live in tool.json and in the handler warnings array, which the agent sees at the point of use. Putting those vendor details back into the prompt would undo the review request to keep unverified/stale provider claims out of the agent’s world model.

src/channels/calendar/nylas-calendar-client.ts
josephfung: This is the right treatment, and thank you for saying plainly where the string came from — that it was a support article rather than a live PUT is exactly the fact that was missing, and naming it is worth more than the citation itself.

The comment now does what an unverifiable vendor claim should do: it cites the source, dates it, states the endpoint shape it describes, and marks its own limitation ("v2-shaped … has not been re-verified against v3 on a live grant here"). Someone hitting this in a year can re-check it instead of inheriting an assertion. The two supporting v3 links — send-rsvp as the documented RSVP write, and the PUT example whose participants carry no status — are corroborating rather than load-bearing, which is the honest weight to give them.

One thing I could not do: I tried to fetch the support article to verify it says what you quote, and the Nylas help centre returns 403 to unauthenticated requests. So I am taking the quotation on trust. That is a real residual, and it is why I would still like the live v3 PUT done at some point — the response body would replace the v2 article and settle it permanently. Not something this PR can do without a grant.

On the Google-native point: you are right and I was wrong to reach for #1330. Its scope is Curia's own calendar via workspace-mcp, not organizer RSVP writes on the principal's calendar — different problem, and retargeting #1735 there would have muddled both. "Impossible on this Nylas client" is the accurate framing and the comment now says so.

Resolving. I will note the outstanding live-verification in my summary rather than hold the thread open for it.

skills/calendar/tools/calendar-update-event/tool.json
josephfung: Good — the manifest is now honest about not knowing, which is the right answer when nobody has checked.

Both the description and the `attendees` input carry it ("RSVP preservation on retained guests is unverified — re-read after writing"), so an agent reading either surface gets the caveat. And your reasoning for why it is a worry rather than a finding is sound: Nylas PUT replacing nested objects plus a v3 example that omits `status` is grounds for suspicion, not evidence of a reset. Documenting it at that exact strength is better than either reassuring or alarming.

Agreed on not adding read-before-write status comparison here — I said as much when I raised it, and bolting a read onto a high-risk write to answer an open question is the wrong order. Answer the question first, then decide whether enforcement is needed.

Resolving.

skills/calendar/tools/calendar-update-event/handler.ts
josephfung: Fixed, and you took it further than I asked in the direction that mattered most.

`normalizedEmail` returns null for non-strings and empties, and both sides filter before comparing, so a room or resource participant with no mailbox can no longer throw. The bidirectional check is there too, with `extra` correctly framed as "often the organizer" rather than treated as a fault.

The part I want to call out is the one I only mentioned in passing: dropping `success: false` when the write already applied. I raised the retry hazard as a consequence of the crash, and you recognised it as the more serious bug underneath — a `high`-risk write that re-issues invitations must never report failure after it has landed, because the honest recovery advice and the safe recovery advice point in opposite directions. `guestListMismatchWarnings` says "The write already applied; do not retry. Re-read the event" on both branches, which resolves that cleanly.

Also a better outcome than my suggestion of failing on extras — you are right that Google putting the organizer in the attendee list would have made that a routine false negative.

Resolving.

agents/calendar.yaml
josephfung: Verified — the prompt is down to behaviour with no vendor causation:

> Never update a participants array to RSVP, and never use `calendar-update-event` to record another guest's RSVP. Use `calendar-respond-to-invite` for the principal's own RSVP. `calendar-update-event` attendees replaces the whole guest list (email + optional name).

That is the durable half. The agent gets the same constraint, and if the Nylas behaviour ever changes, nothing here quietly steers it away from a working capability — the handler error carries the why, at the moment it matters, where it can be corrected in one place.

Version bumped to `0.7.2` with the tripwire test updated, and the prompt test now asserts the behavioural instruction rather than the vendor string, so it will not re-break when the explanation moves again.

Resolving.

skills/calendar/tools/calendar-update-event/tool.json
josephfung: Verified, and doing it in-PR was the right call — the create path's behaviour changed here, so shipping it with a stale manifest would have left the inconsistency worse than before.

`parseWritableAttendees` now lives in `src/channels/calendar/attendee-input.ts` and both handlers use it, so status is rejected at the boundary with the same message on both paths instead of create throwing from the client. The manifest documents `{ email, name? }`, names the constraint, points at `calendar-respond-to-invite`, and `colorId` / `reminders` are gone from both tools.

Extracting the helper rather than duplicating the validation is also what makes the next tool that writes a guest list correct by default.

Resolving. The manifest-semver nit was cosmetic and I am not going to hold anything for it.

josephfung:
## Re-review of `3d9d28c7`

All five of my threads resolved. CI green.

Two of the responses improved on what I asked for. Dropping `success: false` when the write already applied fixed a hazard I had only raised as a side effect of the crash — a `high`-risk write that re-issues invitations must never report failure after it has landed, because at that point the honest advice and the safe advice diverge. And declining to fail on *extra* returned participants is correct where my suggestion was not: Google putting the organizer in the attendee list would have made that a routine false negative.

The citation response is the one I most wanted to see. Saying outright that the error string came from a support article rather than a live PUT is worth more than the link, and the comment now dates the source, names the endpoint shape it describes, and marks its own limitation instead of presenting an inference as a finding.

### One residual, not blocking

I could not verify the support article: the Nylas help centre returns 403 to unauthenticated fetches, so the quoted error text is taken on trust. Combined with the guest-list caveat, there are now **two provider behaviours documented as unverified** in this code:

1. Whether Nylas v3 PUT rejects organizer-set participant status (cited from a v2-era article).
2. Whether a guest-list replace preserves retained attendees' RSVPs.

Both are answerable in a few minutes against a live grant, and both currently shape behaviour — the first is enforced as a hard rejection, the second tells agents to re-read after every attendee write. Worth one small follow-up that checks both and replaces the caveats with the response bodies. If (2) turns out to reset RSVPs, that is a materially more dangerous tool than the docs currently suggest and `action_risk: high` would be doing a lot of quiet work.

Not a reason to hold this PR. The code is honest about what it does not know, which is the property that was missing before.

### Note on the CodeRabbit thread

I agree with the reply there and did not resolve it, since it is not mine to close. The prompt no longer claims guest-list replace re-issues invitations, and the Google/Microsoft notification detail lives in `tool.json` and the returned `warnings` array — at the point of use rather than baked into the agent's world model, which is what the `calendar.yaml` thread was about. Putting it back in the prompt would undo that.

josephfung (commented):
Thanks, changes look good. And references confirmed

josephfung:
@coderabbitai approve

